PR #188, opened 2025-01-02
Support the BEST operator

### LINKED ISSUE #143: Add support for the BEST functionality in PyDough
The specification is as follows:
- `x.BEST(y, by=...)` is equivalent to `x.y` except that it filters `y` to only include a single subcollection record per record of `x` according to the `by` criteria. It is equivalent to a min row number filter in SQL, where `ROW_NUMBER()` is computed for each record of `y` where the partition keys are what is used to join it with `x` and the ordering keys are the `by` clause, and the only rows that are kept are the ones where the row number is 1.
- The `by` semantics obey the same rules as `ORDER_BY` and `TOP_K`.
- `BEST(y, by=...)` can also be used standalone to create a child collection access. For example, `regions(name, best_nation_name=BEST(nation, by=COUNT(customers.DESC())).name` finds, for each region, the name of the nation with the most customers.
- The subcollection obtained by `BEST` is treated as singular.
- Optional keyword argument: `allow_ties` (default=False). If true, uses `RANK` instead of `ROW_NUMBER`, but the result is no longer treated as singular.
- Optional keyword argument: `n` (default=1). If anything other than 1, keeps the top `n` values of `ROW_NUMBER`, but the result is no longer treated as singular. `n` must always be a positive integer. If `allow_ties=True`, then `n` **must** be equal to 1.
- Support for this operation is needed in the unqualified nodes, qualification process, QDAG nodes, hybrid conversion, hybrid tree operators, relational conversion, relational nodes, and SQLGlot translation.

More examples:
```py
# For each customer, finds the name of the part they have ordered the most (by quantity) in a single order:
customers(name, best_part_name=BEST(orders.lines, by=quantity).part.name)

# For each part, find the name of the suppliers that provides it the most cheaply, listing all of them if there are ties:
parts.BEST(supply_records, by=supply_cost.ASC(), allow_ties=True).supplier(part_name=BACK(2).name, supplier_name=name)

# Filter the parts to only include the most expensive part of each part type
part_types = PARTITION(parts, name="cluster", by=part_type)
part_types.BEST(cluster, by=retail_price.DESC())

# Find the part that each customer has ordered the most, by quantity
customers_parts_ordered = PARTITION(
 customers.orders.lines(cust_name=BACK(3).name, part_name=part.name),
 name="combo",
 by=(cust_name, part_name),
)(total_amount_ordered=SUM(combo.quantity))
customers_data = PARTITION(
 customers_parts_ordered,
 name="parts_ordered",
 by=cust_name,
)
customers_data.BEST(parts_ordered, by=total_amount_ordered.DESC())

# For each part, compute the percentage of the total available quantity of that part that comes from the 3 suppliers with the highest available quantity.
top_3_suppliers = BEST(supply_records, by=avail_qty.DESC(), n=3)
parts(
 name,
 top_3_percent=100 * SUM(top_3_suppliers.avail_qty) / SUM(supply_records.avail_qty)
)
```

knassre-bodo:
Resolves #143. Adds the `BEST` operator by making it syntactic sugar that is expanded during the unqualified node step into `.WHERE(RANKING(...) == 1).SINGULAR()` (or variants thereof, depending on the keyword arguments). Also, makes a change to how window functions are handled in hybrid conversion that avoids unecessary correlated references by replacing the uniqueness keys of a parent hybrid tree with values in the current hybrid tree if they are equal according to the join keys. For example, if deriving the average time-difference between orders per-customer with a window function, which normally uses `levels=1` and would create a correlated reference out of `orders` back to `customer` to use the customer's key as the partition key, can instead just use the customer key field of the orders as the partition key (since we know that the two will be equal after joining).


pydough/conversion/hybrid_tree.py
```diff
@@ -2093,13 +2093,25 @@ def add_unique_terms(
...
+                    partition_args.append(equivalent_key)
+                else:
+                    if not isinstance(arg, HybridCorrelExpr):
+                        if child_idx is not None:
+                            prev_hybrid.correlated_children.add(child_idx)
+                        else:
+                            prev_hybrid.correlated_children.add(
+                                len(prev_hybrid.children)
+                            )
+                    partition_args.append(HybridCorrelExpr(prev_hybrid, arg))
```
knassre-bodo, line 2114: This is the implementation of the optimization alluded to in the PR description. It drastically compresses the plan of one of the new tests (supplier_best_part) but also makes positive changes to a bunch of other tests by avoiding unnecessary correlation->de-correlation paths, which has a nice final-plan benefit since it causes the window functions to be computed before the join instead of after.

tests/test_plan_refsols/avg_order_diff_per_customer.txt
```diff
@@ -1,9 +1,9 @@
...
-   PROJECT(columns={'avg_diff': agg_0, 'name': agg_6})
-    AGGREGATE(keys={'key': key}, aggregations={'agg_0': AVG(day_diff), 'agg_6': ANYTHING(name)})
-     PROJECT(columns={'day_diff': DATEDIFF('days':string, PREV(args=[order_date], partition=[key], order=[(order_date):asc_last]), order_date), 'key': key, 'name': name})
-      JOIN(conditions=[t0.key == t1.customer_key], types=['inner'], columns={'key': t0.key, 'name': t0.name, 'order_date': t1.order_date})
-       SCAN(table=tpch.CUSTOMER, columns={'key': c_custkey, 'name': c_name})
+   PROJECT(columns={'avg_diff': agg_0, 'name': name})
+    JOIN(conditions=[t0.key == t1.customer_key], types=['inner'], columns={'agg_0': t1.agg_0, 'name': t0.name})
+     SCAN(table=tpch.CUSTOMER, columns={'key': c_custkey, 'name': c_name})
+     AGGREGATE(keys={'customer_key': customer_key}, aggregations={'agg_0': AVG(day_diff)})
+      PROJECT(columns={'customer_key': customer_key, 'day_diff': DATEDIFF('days':string, PREV(args=[order_date], partition=[customer_key], order=[(order_date):asc_last]), order_date)})
```
knassre-bodo, line 8: This is an example of the optimization caused by avoiding the unnecessary correlation/de-correlation. Now, the `PREV` uses `customer_key` (within `ORDERS`) as its partition key, instead of joining `ORDERS` to `CUSTOMER` just so it could use the `key` field from `CUSTOMER`.

pydough/unqualified/unqualified_node.py
```diff
@@ -386,6 +386,57 @@ def SINGULAR(self) -> "UnqualifiedSingular":
...
+        allow_ties: bool = False,
+        n_best: int = 1,
+    ) -> "UnqualifiedNode":
+        """
+        Method used to create the BEST logic, where x.BEST(y, by=z) is later
+        expanded into `x.y.WHERE(RANKING(by=z) == 1, levels=...).SINGULAR()`,
+        where the levels is the height of y. There are also the following
+        variations:
+        - If `allow_ties` is True: `x.y.WHERE(RANKING(by=z, allow_ties=True) == 1, levels=...)`
+        - If `n_best` > 1 : `x.y.WHERE(RANKING(by=z) <= n_best, levels=...)`
```
knassre-bodo, line 402: This expansion will happen when the UnqualifiedBest node gets qualified in `qualification.py`.

pydough/conversion/hybrid_tree.py
knassre-bodo: Everything in this section of code is the same as the old version except for lines 2100-2107. If we know that the argument that we **would** wrap in a correlated reference (because it belongs to `prev_hybrid`) is also an LHS join key, we can just replace it with the corresponding RHS key, which belongs to `hybrid` and therefore doesn't need to be wrapped in a `HybridCorrelExpr`. The avoidance of creating an unnecessary correlated reference allows us to avoid the inferior relational plan structure that is created when it gets de-correlated. 

The trigger for this occurring is a very common pattern with window functions (which occurs 100% of the time whenever you use `BEST`):
```
# Here, this creates a correlated reference from within `nations.customers`
# back to `regions`. since `levels=2` means the `RANKING` must partition
# on the uniqueness keys of regions (aka `regions.key`). However, `regions.key`
# is also the join key to connect `regions` to `nations.customers`, where
# `nations.region_key` is the corresponding join key on the other side. So we can
# substitute `nations.region_key` as the partition key, which means we do not
# need a correlated reference since `nations.region_key` is on the same side of
# the join that the window function is happening in.
regions.CALCULATE(richest_customer_name=nations.customers.WHERE(RANKING(by=acctbal.DESC(), levels=2) == 1).SINGULAR().name)
```

pydough/unqualified/qualification.py
```diff
@@ -745,6 +747,74 @@ def qualify_singular(
...
+        # breakpoint()
+
+        # Build the final expanded window-based filter, with `SINGULAR` added
+        # if-applicable (not allowing ties, not keeping multiple bests, not in
+        # the form `x.BEST(y)` unless as a child & the x is already singular).
+        if n_best == 1 and not allow_ties and is_child:
+            base: PyDoughCollectionQDAG = qualified_parent.starting_predecessor
+            relative_ancestor: PyDoughCollectionQDAG = context.starting_predecessor
+            if base is relative_ancestor or base.is_singular(relative_ancestor):
+                qualified_child = self.builder.build_singular(qualified_child)
```
knassre-bodo, line 814: We don't want to insert `SINGULAR` in a case like the `region_orders_from_nations_richest` test. In that case, the child inside the calculate is `nations.BEST(customers, by=(acctbal.DESC(), name.ASC())).orders`. In this case, it is inaccurate to rewrite this as `nations.customers.WHERE(...).SINGULAR().orders` since even though the `customers` are now singular with regards to `nations` post-filter, the `nations` are still plural with regards to the parent context (meaning `nations.customers` is as well), and `SINGULAR` would suggest that the whole thing up to that point is singular with regards to the parent context (which is inaccurate).

documentation/dsl.md
```diff
@@ -1242,12 +1242,9 @@ Addresses.CALCULATE(
 <!-- TOC --><a name="best"></a>
 ### BEST
 
-> [!IMPORTANT]
-> This feature has not yet been implemented in PyDough
-
 PyDough supports identifying a specific record from a sub-collection that is optimal with regards to some metric, per-record of the current collection. This is done by using `BEST` instead of directly accessing the sub-collection. The first argument to `BEST` is the sub-collection to be accessed, and the second is a `by` argument used to find the optimal record of the sub-collection. The rules for the `by` argument are the same as `PREV`, `NEXT`, `TOP_K`, etc.: it must be either a single collation term, or an iterable of 1+ collation terms.
```
vineetg3, line 1245: I think there might be instances where a user might think a `sub-collection` is just the immediate one as opposed to "any" grand-child `sub-collection`. Maybe we can clarify that here?

knassre-bodo: This is demonstrated by some of the examples, but sure thing I can also clarify it here.

pydough/unqualified/qualification.py
```diff
@@ -745,6 +747,74 @@ def qualify_singular(
...
+            (rank == n_best) if n_best == 1 else (rank <= n_best)
+        )
+        children: list[PyDoughCollectionQDAG] = []
+        qualified_cond: PyDoughExpressionQDAG = self.qualify_expression(
+            unqualified_cond, qualified_child, children
+        )
+        qualified_child = self.builder.build_where(
+            qualified_child, children
+        ).with_condition(qualified_cond)
+        # breakpoint()
```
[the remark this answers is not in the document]
knassre-bodo: Good catch!

pydough/unqualified/qualification.py
```diff
@@ -920,6 +924,82 @@ def qualify_singular(
...
+        ).with_condition(qualified_cond)
+
+        # Extract the `levels` argument from the condition
+        assert isinstance(qualified_cond, ExpressionFunctionCall)
+        rank_call = qualified_cond.args[0]
+        assert isinstance(rank_call, WindowCall)
+        levels: int | None = rank_call.levels
+
+        # Add `SINGULAR` if applicable (there is a levels argument and that
+        # ancestor either is the context or is singular withr egards to it).
```
hadia206, line 987:
```suggestion
        # ancestor either is the context or is singular withr regards to it).
```

pydough/unqualified/qualification.py
```diff
@@ -920,6 +924,82 @@ def qualify_singular(
...
+        # if-applicable (not allowing ties, not keeping multiple bests, not in
```
hadia206, line 988: For my understanding, why?

pydough/unqualified/qualification.py
```diff
@@ -920,6 +924,82 @@ def qualify_singular(
...
+        # the form `x.BEST(y)` unless as a child & the x is already singular).
```
hadia206, line 989: > the form `x.BEST(y)` unless as a child & the x is already singular

which expression "as a child" refers to ? The whole `x.BEST(y)`

pydough/unqualified/unqualified_node.py
```diff
@@ -385,6 +385,51 @@ def SINGULAR(self) -> "UnqualifiedSingular":
...
+        self,
+        by: Union[Iterable["UnqualifiedNode"], "UnqualifiedNode"],
+        per: str | None = None,
+        allow_ties: bool = False,
+        n_best: int = 1,
+    ) -> "UnqualifiedNode":
+        """
+        Method used to create the BEST logic, where x.y.BEST(by=z, per="x") is later
+        expanded into `x.y.WHERE(RANKING(by=z) == 1, per="x").SINGULAR()`,
+        where the levels is the height of y. There are also the following
```
hadia206, line 398: This line about `levels` is better to be after the variations. It's confusing here since the expansion doesn't include it

pydough/unqualified/unqualified_node.py
```diff
@@ -385,6 +385,51 @@ def SINGULAR(self) -> "UnqualifiedSingular":
...
+            `by`: the collation terms to order by in the `RANKING` call.
+            `per`: the ancestor that the `BEST` computation is happening with
+            regards to, or None if it is a global optimum being sought.
+            `allow_ties`: whether to allow ties in the ranking. If True, it
+            means that if there are multiple entries with the same rank, they
+            will all be considered as "best" candidates. This cannot be used
+            with `n_best > 1`.
+            `n_best`: the number of best entries to consider. If greater than
+            1, it will find all entries with a rank less than or equal to
+            `n_best`. This cannot be used when `allow_ties` is True.
```
hadia206, line 415: why it can't be used?

hadia206 (approved):
Overall looks good to me.

pydough/unqualified/unqualified_node.py
knassre-bodo: The three legal combinations represent different variations of an MRNF filter:
- `n_best=1` & `allow_ties=False` => `ROW_NUMBER = 1`
- `n_best=1` & `allow_ties=True` => `RANK = 1`
- `n_best=n` & `allow_ties=False` => `ROW_NUMBER <= n`

But if we say we want the `n` best values allowing ties, there is no coherent logic for that (and frankly it doesn't even make sense in English). 

For example, `RANK <= n` makes no sense because if I have values like this `[a, a, a, b, b, b, c, c, c]` and I use a value of `n` between 1 and 5 I will get `[a, a, a]` then if I use a value between 6 and 8 I'll get `[b, b, b]` and if I use a value >= 9 I'll get the entire list. Doing `DENSE_RANK <= n` would give yet another answer which is potentially valid under one obscure interpretation of what n-best with ties means, but its still not a very intuitive/coherent idea.

Since `BEST` is meant to be a simplification, I'd rather just have a user write out whatever convoluted ranking logic they want to use if it is anything that is not strongly rooted in an intuitive analytical phrasing.

pydough/unqualified/unqualified_node.py
knassre-bodo: Whoops, should delete the `levels` bit (I wrote this originally before the `per` part got implemented).

pydough/unqualified/qualification.py
knassre-bodo: Oops this should get deleted, its a legacy of an earlier variant of how `BEST` was going to work.

pydough/unqualified/qualification.py
knassre-bodo: Lets say I do `regions.CALCULATE(n=nations.customers.BEST(by=acctbal.DESC(), per="nations").name)`. This should be illegal because even though the `BEST` picks the best customer per-nation, there are still multiple nations per region ergo there would be multiple values of `n` for each region (which is invalid). Therefore, we should **NOT** insert `SINGULAR` when converting this, because the following _would_ be considered "valid" by PyDough then produce nonsense runtime results:

```py
regions.CALCULATE(n=nations.customers.WHERE(RANKING(by=acctbal.DESC(), per="nations") == 1).SINGULAR().name)`
```

**HOWEVER**, if I did the exact same thing but with `per="regions"` then it would be fine because there would only be one selected value of `nations.customers` per-region.

Issue #141, opened 2024-12-17
Support correlated backreferences in QDAG to Relational conversion


knassre-bodo:
Currently, `BACK` references are not supported in QDAG to Relational conversion if they step out of a child subtree of the hybrid tree back into the parent hybrid tree. This is caused by operations such as `nations.WHERE(HAS(customer.WHERE(name == BACK(1).name)))`, since the `BACK(1).name` steps out of `HAS(...)` into `nations`. This corresponds to a correlated reference between the left side of the join (`nations`) and the right side (the `HAS(...)`). Implementing this functionality will require the following:
- A way to represent correlated references in the hybrid tree
- A way to represent correlated references in the relational nodes
- Support for converting these back references into the hybrid tree correlated references (even if they step out of the child more than once)
- Support for converting these correlated hybrid nodes in the relational conversion step

knassre-bodo:
# Update

As of #254, correlation & de-correlation support has been added, but the special optimized cases for singular (only match) and aggregation (only match) have not been implemented; instead they do the same thing as regular singular & aggregation. Adding these optimizations will be a followup issue to add at a later date as part of a larger pattern of removing duplicate/unnecessary subtrees.

PR #232, opened 2025-01-23
Support uses of BACK that cause correlated references: Hybrid & Relational support for Correlate nodes

knassre-bodo:
Child of #269 that is part of addressing #141. This PR deals with the hybrid & relational conversion, including the creation of new types of hybrid/relational nodes to express correlation, and also defines all of the correlation unit tests. This means that all correlated queries should produce a Relational plan, but will not be possible to convert to SQLGlot until the next PR. Subsequent PRs will deal with cases where a the relational plan needs to be de-correlated.


tests/test_pipeline.py
```diff
@@ -14,6 +14,16 @@
 )
 from simple_pydough_functions import (
     agg_partition,
```
knassre-bodo, line 16: TPCH query 21 now runs e2e w/o issue. Queries 5/22 still fail, as do half of the `correl_x` tests, because they fall into the patterns shown in the comment I added to this issue #141 (mostly the AGGREGATION case).

tests/simple_pydough_functions.py
```diff
@@ -270,3 +270,129 @@ def triple_partition():
     )(supp_region, avg_percentage=AVG(cust_regions.percentage)).ORDER_BY(
         supp_region.ASC()
     )
+
+
+def correl_1():
```
knassre-bodo, line 275: Some of these correl examples are about different patterns of how the HybridCorrelRefExpr is derived (different edge cases), but most are about the different connection types.

tests/simple_pydough_functions.py
```diff
@@ -270,3 +270,216 @@ def triple_partition():
...
+    # Which part brands have at least 1 part that more than 40% above the
+    # average retail price for all parts from that brand.
+    # (This is a correlated SEMI-join)
+    brands = PARTITION(Parts, name="p", by=brand)(avg_price=AVG(p.retail_price))
+    outlier_parts = p.WHERE(retail_price > 1.4 * BACK(1).avg_price)
+    selected_brands = brands.WHERE(HAS(outlier_parts))
+    return selected_brands(brand).ORDER_BY(brand.ASC())
+
+
+def correl_12():
```
knassre-bodo, line 412: This is to deal with edge cases in how the `PARTITION` handling works in the hybrid tree

tests/simple_pydough_functions.py
```diff
@@ -270,3 +270,216 @@ def triple_partition():
...
+    selected_parts = p.WHERE(
+        (retail_price > BACK(1).avg_price)
+        & (retail_price < BACK(2).avg_price)
+        & (size < 3)
+    )
+    selected_brands = brands.WHERE(HAS(selected_parts))
+    return selected_brands(brand).ORDER_BY(brand.ASC())
+
+
+def correl_13():
```
knassre-bodo, line 431: These last 3 deal with edge cases involving multiple correlations / multiple levels of joins, and with order of operations.

tests/simple_pydough_functions.py
```diff
@@ -270,3 +270,216 @@ def triple_partition():
...
+    )
+    selected_supply_records = supply_records.WHERE(HAS(selected_part))
+    supplier_info = Suppliers.WHERE(nation_key == 19)(
+        avg_price=AVG(supply_records.part.retail_price)
+    )
+    selected_suppliers = supplier_info.WHERE(HAS(selected_supply_records))
+    return TPCH(n=COUNT(selected_suppliers))
+
+
+def correl_15():
```
knassre-bodo, line 466: This one involves multiple types of correlation behavior overlapping: semi + aggregate.

tests/test_plan_refsols/tpch_q5.txt
```diff
@@ -0,0 +1,21 @@
+ROOT(columns=[('N_NAME', N_NAME), ('REVENUE', REVENUE)], orderings=[(ordering_1):desc_last])
```
knassre-bodo, line 1: Correlation of this query: finds lines where the supplier & customer are from the same nation.
- The join in question is the one with the name `corr10`.
- The left side has all of the nations whose region is asia
- The right side aggregates all of the lineitems from customers in that region.
- Important filter is `name_9 == corr10.name` (meaning that the name of the supplier from the right side is the same as the name of the nation from the left side)

tests/test_plan_refsols/tpch_q21.txt
```diff
@@ -0,0 +1,21 @@
+ROOT(columns=[('S_NAME', S_NAME), ('NUMWAIT', NUMWAIT)], orderings=[(ordering_1):desc_last, (ordering_2):asc_first])
```
knassre-bodo, line 1: Correlation of this query: goes from lineitem (L1) -> order -> lineitem (L2) to find all instances where a lineitem is in an order with multiple suppliers but that specific supplier's lineitem entries in the order are the only ones that are late, so it needs to know if  the supplier key from L1 and L2 are different to determine if two lineitems in the same order have different suppliers.
- First important join in question is the one with the name `corr5`.
  - The left side has L1 & order, filtered to only include late lines
  - The right side has L2.
  - Important filter is `supplier_key != corr5.supplier_key`, used to semi-join L1 and L2 when there is a matching L2 entry that does _not_ have the same supplier key as L1.
- Second important join in question is the one with the name `corr6`.
  - The left side has the result of the first correlate.
  - The right side has _another_ L2.
  - Important filter is `supplier_key != corr6.supplier_key`,  used to anti-join L1 and L2 when there is a matching L2 entry that does _not_ have the same supplier key as L1 (as well as some extra properties).

tests/test_plan_refsols/tpch_q22.txt
```diff
@@ -0,0 +1,18 @@
+ROOT(columns=[('CNTRY_CODE', CNTRY_CODE), ('NUM_CUSTS', NUM_CUSTS), ('TOTACCTBAL', TOTACCTBAL)], orderings=[])
```
knassre-bodo, line 1: Correlation of this query: first derives a global average of the selected customers' account balances, then partitions the customers but only the ones who are _above_ the global average.

- Important join in question is the one with the name `corr1`.
- The left side has the global aggregation
- The right side has the selected customers who have been aggregated.
- Inside the RHS, before the aggregation happens, the selected customers get further filtered to only include the ones where `acctbal > corr1.avg_balance`

pydough/sqlglot/sqlglot_relational_expression_visitor.py
```diff
@@ -133,6 +134,11 @@ def visit_literal_expression(self, literal_expression: LiteralExpression) -> Non
         )
         self._stack.append(literal)
 
+    def visit_correlated_reference(
+        self, correlated_reference: CorrelatedReference
+    ) -> None:
+        raise NotImplementedError("TODO")
```
knassre-bodo, line 140: The next PR in the stack deals with this part: https://github.com/bodo-ai/PyDough/pull/234

pydough/conversion/hybrid_tree.py
```diff
@@ -229,6 +229,27 @@ def shift_back(self, levels: int) -> HybridExpr | None:
         return HybridBackRefExpr(self.name, self.back_idx + levels, self.typ)
 
 
+class HybridCorrelExpr(HybridExpr):
+    """
+    Class for HybridExpr terms that are expressions from a parent hybrid tree
+    rather than an ancestor, which requires a correlated reference.
```
knassre-bodo, line 235: If it needs to go out _several_ levels, then it will be `HybridCorrelExpr(HybridCorrelExpr(...))`

pydough/conversion/hybrid_tree.py
```diff
@@ -1269,108 +1295,7 @@ def populate_children(
             for con_typ in reference_types:
                 connection_type = connection_type.reconcile_connection_types(con_typ)
             child_idx_mapping[child_idx] = hybrid.add_child(subtree, connection_type)
-
-    def make_hybrid_agg_expr(
```
knassre-bodo, line 1273: This API is no longer needed/useful; can merge with `make_hybrid_expr` & have an argument for if we are inside of an aggregation.

pydough/conversion/hybrid_tree.py
```diff
@@ -1609,80 +1694,51 @@ def make_hybrid_expr(
...
-                # For non-aggregate function calls, translate their arguments
-                # normally and build the function call. Does not support any
+            case ExpressionFunctionCall():
+                if expr.operator.is_aggregation and inside_agg:
+                    raise NotImplementedError(
+                        "PyDough does not yet support calling aggregations inside of aggregations"
+                    )
+                # Do special casing for operators that an have collection
+                # arguments.
+                # TODO: (gh #148) handle collection-level NDISTINCT
```
knassre-bodo, line 1704: This TODO already existed, just got moved around.

pydough/conversion/hybrid_tree.py
```diff
@@ -1533,11 +1458,165 @@ def handle_has_hasnot(
...
+                    "PyDough does yet support aggregations whose arguments mix between subcollection data of the current context and fields of the context itself"
+                )
+            case HybridWindowExpr():
+                raise NotImplementedError(
+                    "PyDough does yet support aggregations whose arguments mix between subcollection data of the current context and window functions"
+                )
+            case _:
+                raise NotImplementedError(
+                    f"TODO: support converting {expr.__class__.__name__} in aggregations"
+                )
```
knassre-bodo, line 1509: This TODO already existed, just got moved around.

pydough/conversion/hybrid_tree.py
```diff
@@ -1862,13 +1922,16 @@ def make_hybrid_tree(
                         successor_hybrid = self.make_hybrid_tree(
                             node.child_access.child_access, parent
                         )
-                        partition_by = node.child_access.ancestor_context
+                        partition_by = (
+                            node.child_access.ancestor_context.starting_predecessor
+                        )
```
knassre-bodo, line 1927: This is necessary for cases like this:
```
a = PARTITION(data, name="cluster", by=(x,y))
b = a.WHERE(COUNT(cluster) > 10)
c = b(n=AVG(cluster.z)) 
d = c.cluster.WHERE(z > BACK(1).n)
```
In this case, whenever dealing with `c` or `d`,  we need `partition_by` to refer to `a`, even though `node.child_access.ancestor_context` can refer to `b` or `c`. That's why `.starting_prdecessor` is used (keeps calling `.predecessor` until we get back to `a`).

pydough/conversion/relational_converter.py
```diff
@@ -199,6 +228,28 @@ def translate_expression(
...
+                            expr.typ,
+                        )
+                    case CorrelatedReference():
+                        return ancestor_expr
+                    case _:
+                        raise ValueError(
+                            f"Unsupported expression to reference in a correlated reference: {ancestor_expr}"
+                        )
             case _:
                 raise NotImplementedError(expr.__class__.__name__)
```
amullerbodo, line 254: @knassre-bodo not sure if we could add additional info inside the NotImplementedError. The team has struggled to interpret some of the errors so maybe if they knew what is the expression and some context of the expression so they can modify more easily.

pydough/relational/relational_expressions/correlated_reference_finder.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def reset(self) -> None:
+        self._correlated_references = set()
+
+    def get_correlated_references(self) -> set[CorrelatedReference]:
+        return self._correlated_references
+
+    def visit_call_expression(self, call_expression: CallExpression) -> None:
+        for arg in call_expression.inputs:
+            arg.accept(self)
```
amullerbodo, line 31: @knassre-bodo this is for my own benefit you are traversing expressions here to find CorrelatedExpressions... 
The accept method in CallExpression would do something like this: 
def accept(self, visitor: RelationalExpressionVisitor) -> None:
        visitor.visit_call_expression(self)
so you are calling accept on the finder so that the CallExpression calls visit_call_expression of self again? why is that? why not call directly visit_correlated_reference(arg)?

pydough/relational/relational_expressions/correlated_reference_finder.py
```diff
@@ -0,0 +1,48 @@
...
+            order_arg.expr.accept(self)
+
+    def visit_literal_expression(self, literal_expression: LiteralExpression) -> None:
+        pass
+
+    def visit_column_reference(self, column_reference: ColumnReference) -> None:
+        pass
+
+    def visit_correlated_reference(self, correlated_reference) -> None:
+        self._correlated_references.add(correlated_reference)
```
amullerbodo, line 48: The argument here will be added to a set. @knassre-bodo can you share if the set operation is working on the actual memory position of the object or you have overloaded __eq__ or __hash__ somewhere above RelationalExpression?

tests/correlated_pydough_functions.py
amullerbodo, line 1: looks good

pydough/conversion/relational_converter.py
knassre-bodo: Will do, though in this specific case the error should be unreachable (all other cases are accounted for) so it's mostly there for internal debugging purposes.

pydough/relational/relational_expressions/correlated_reference_finder.py
knassre-bodo: Every RelationalExpression class has its own `.accept` that dispatches to the visitor's corresponding `visit_xxx` method. This is needed because we don't know what the arguments are:
- If the function call arguments are more function calls, it recursively calls `visit_call_expression`
- If the function call arguments are literals, it calls `visit_literal`
- If the function call arguments are correlatex expressions, it calls `visit_correlated_reference`
etc.

pydough/relational/relational_expressions/correlated_reference_finder.py
knassre-bodo: `__eq__` is defined in `RelationalExpression` (points to a `.equals()` method which each of the implementing classes extends), but good catch regarding `__hash__` being incomplete & only present in some (will add to `RelationalExpression`).

pydough/conversion/hybrid_tree.py
```diff
@@ -1533,11 +1458,195 @@ def handle_has_hasnot(
...
+            raise ValueError(
+                f"Expected aggregation call to contain references to exactly one child collection, but found {len(child_indices)} in {expr}"
+            )
+        hybrid_call: HybridFunctionExpr = HybridFunctionExpr(
+            expr.operator, converted_args, expr.pydough_type
+        )
+        # Identify the child connection that the aggregation call is pushed
+        # into.
+        child_idx: int = child_indices.pop()
+        child_connection = hybrid.children[child_idx]
```
vineetg3, line 1547: Type is missing here.

pydough/conversion/hybrid_tree.py
```diff
@@ -1533,11 +1458,195 @@ def handle_has_hasnot(
...
+            `steps_taken_so_far`: the number of steps already taken to step
+            up from the BACK node. This is needed so we know how many steps
+            still need to be taken upward once we have stepped out of hte child
+            subtree back into the parent subtree.
+        """
+        if len(self.stack) == 0:
+            raise ValueError("Back reference steps too far back")
+        # Identify the parent subtree that the BACK reference is stepping back
+        # into, out of the child.
+        parent_tree = self.stack.pop()
```
vineetg3, line 1581: Type hint of parent_tree is missing here.

pydough/conversion/hybrid_tree.py
```diff
@@ -1533,11 +1458,195 @@ def handle_has_hasnot(
...
+        Converts a BACK reference into a correlated reference when the number
+        of BACK levels exceeds the height of the current subtree.
+
+        Args:
+            `back_expr`: the original BACK reference to be converted.
+            `collection`: the collection at the top of the current subtree,
+            before we have run out of BACK levels to step up out of.
+            `steps_taken_so_far`: the number of steps already taken to step
+            up from the BACK node. This is needed so we know how many steps
+            still need to be taken upward once we have stepped out of hte child
```
vineetg3, line 1574: Typo: ..out of `the` child..

vineetg3 (approved):
I just have minor nits. Otherwise, it LGTM to the best of my knowledge.

knassre-bodo:
Related to #141.

pydough/conversion/hybrid_tree.py
knassre-bodo: Thank you! I have a typing propensity for accidentally pressing the `h` just before the `t`.

PR #233, opened 2025-01-28
Added a logger and implemented in execute_df (#163)

### LINKED ISSUE #163: Add proper Python logging to PyDough
Should be used for places such as displaying the generated SQL in `execute_df`, instead of printing.

vineetg3:
Resolves #163. 
* Exposed the display_sql kwarg to_df and subsequently to execute_df.
* Pretty SQL is logged now in execute_df.
* Created a configured logger for reuse across PyDough.


vineetg3:
This Resolves #163

pydough/logger/__init__.py
```diff
@@ -0,0 +1,5 @@
+__all__ = [
+    "get_logger"
+]
+
+from .logger import get_logger
```
knassre-bodo, line 5: Add a newline at the bottom of the file

pydough/logger/__init__.py
```diff
@@ -0,0 +1,5 @@
+__all__ = [
```
knassre-bodo, line 1: The module needs a module-level docstring & module-level README (see other modules for examples). Alternatively, you could forgo a new folder and place `logger.py` in the `configs` module, which would make sense since configuring the logger could eventually be tied to the PyDough session configs.

pydough/logger/logger.py
```diff
@@ -0,0 +1,43 @@
+import logging
```
knassre-bodo, line 1: This needs a file-level docstring (see other files like `pydough_configs.py` or `builtin_databases.py` for examples)

pydough/logger/logger.py
```diff
@@ -0,0 +1,43 @@
...
+import os
+import sys
+
+
+def get_logger(
```
knassre-bodo, line 6: I'd make sure this gets imported to the main `__init__.py` for PyDough, then include documentation for this in:
- The README.md of the module it is in
- The `usage.md` file

pydough/logger/logger.py
```diff
@@ -0,0 +1,43 @@
...
+    name=__name__, 
+    default_level=logging.INFO, 
+    fmt="%(asctime)s [%(levelname)s] %(name)s: %(message)s",
+    handlers=None,
+    level_env_var="PYDOUGH_LOG_LEVEL"
```
knassre-bodo, line 11: I wouldn't allow this to be a variable. Instead, I would note the `PYDOUGH_LOG_LEVEL` environment variable in the `usage.md` file when this function is documented.

pydough/logger/logger.py
```diff
@@ -0,0 +1,43 @@
+import logging
+import os
+import sys
+
+
+def get_logger(
+    name=__name__, 
+    default_level=logging.INFO,
```
knassre-bodo, line 8: The arguments & outputs to this function should be type annotated.

pydough/logger/logger.py
```diff
@@ -0,0 +1,43 @@
...
+    name=__name__, 
+    default_level=logging.INFO, 
+    fmt="%(asctime)s [%(levelname)s] %(name)s: %(message)s",
+    handlers=None,
+    level_env_var="PYDOUGH_LOG_LEVEL"
+):
+    """
+    Returns a logger with specified handlers, allowing the logging level to be set externally.
+
+    :param name: Logger name, usually __name__ from the calling module
```
knassre-bodo, line 16: The format should be:
```
Description description description description.

Args:
   `arg0_name`: Argument description.
   `arg1_name`: Argument description.

Returns:
   Description of the return value
```

tests/test_logging.py
```diff
@@ -0,0 +1,122 @@
...
+
+def get_function_name():
+    return inspect.currentframe().f_back.f_code.co_name
+
+
+def test_get_logger_default_level(capsys):
+    """
+    Test logger with the default logging level and output.
+    """
+    logger = get_logger(get_function_name()+"_test_logger")
```
knassre-bodo, line 28: Instead of `get_function_name`, I recommend manually providing a name for each of these tests, and making sure the appropriate name is used..

tests/test_logging.py
```diff
@@ -0,0 +1,122 @@
...
+    assert "INFO" in captured.out
+    assert "This is an INFO message" in captured.out
+
+
+def test_get_logger_env_override(monkeypatch, capsys):
+    """
+    Test overriding the log level using an environment variable.
+    """
+    # Set the environment variable to override the log level
+    monkeypatch.setenv("PYDOUGH_LOG_LEVEL", "DEBUG")
```
knassre-bodo, line 40: That is a neat trick I hadn't heard of before!

tests/test_logging.py
```diff
@@ -0,0 +1,122 @@
...
+    "root, output",
+    [
+        pytest.param(
+            tpch_query_1_plan(),
+            tpch_q1_output(),
+            id="tpch_q1",
+        ),
+    ],
+)
+def test_execute_df_logging(
```
knassre-bodo, line 100: Couple of things here:
- This shouldn't be parameterized since you only ever have 1 hardcoded output. 
- `tpch_q1_output` shouldn't be here since you never use `output`
- If you want to make it parameterized, `required_op` should be part of the parameterization (but it'd be fine to just make it non-parameterized & do everything in the function button).

tests/test_logging.py
```diff
@@ -0,0 +1,122 @@
...
+    captured_output = output_capture.getvalue()
+
+    required_op = '''[INFO] pydough.sqlglot.execute_relational: SQL query:\n, SELECT L_RETURNFLAG, L_LINESTATUS, SUM_QTY, SUM_BASE_PRICE, SUM_DISC_PRICE, SUM_CHARGE, CAST(SUM_QTY AS REAL) / COUNT_ORDER AS AVG_QTY, CAST(SUM_BASE_PRICE AS REAL) / COUNT_ORDER AS AVG_PRICE, CAST(SUM_DISCOUNT AS REAL) / COUNT_ORDER AS AVG_DISC, COUNT_ORDER FROM (SELECT L_RETURNFLAG, L_LINESTATUS, SUM(L_QUANTITY) AS SUM_QTY, SUM(L_EXTENDEDPRICE) AS SUM_BASE_PRICE, SUM(L_DISCOUNT) AS SUM_DISCOUNT, SUM(TEMP_COL0) AS SUM_DISC_PRICE, SUM(TEMP_COL1) AS SUM_CHARGE, COUNT() AS COUNT_ORDER FROM (SELECT L_QUANTITY, L_EXTENDEDPRICE, L_DISCOUNT, L_RETURNFLAG, L_LINESTATUS, TEMP_COL0, TEMP_COL0 * (1 + L_TAX) AS TEMP_COL1 FROM (SELECT L_QUANTITY, L_EXTENDEDPRICE, L_DISCOUNT, L_TAX, L_RETURNFLAG, L_LINESTATUS, L_EXTENDEDPRICE * (1 - L_DISCOUNT) AS TEMP_COL0 FROM (SELECT L_QUANTITY, L_EXTENDEDPRICE, L_DISCOUNT, L_TAX, L_RETURNFLAG, L_LINESTATUS FROM (SELECT L_QUANTITY, L_EXTENDEDPRICE, L_DISCOUNT, L_TAX, L_RETURNFLAG, L_LINESTATUS, L_SHIPDATE FROM LINEITEM) WHERE L_SHIPDATE <= '1998-12-01'))) GROUP BY L_RETURNFLAG, L_LINESTATUS) ORDER BY L_RETURNFLAG, L_LINESTATUS
```
knassre-bodo, line 120: If this is a multiline string, make it so instead of using `\n`

pydough/sqlglot/execute_relational.py
```diff
@@ -87,8 +88,7 @@ def execute_df(
     """
     sqlglot_dialect: SQLGlotDialect = convert_dialect_to_sqlglot(ctx.dialect)
     sql: str = convert_relation_to_sql(relational, sqlglot_dialect, bindings)
-    # TODO: (gh #163) handle with a proper Python logger instead of
-    # just printing
     if display_sql:
-        print("SQL query:\n", sql)
+        pyd_logger = get_logger(__name__)
+        pyd_logger.info(f"SQL query:\n, {sql}")
```
knassre-bodo, line 93: I just learned that sqlglot has a `pretty` argument that can be passed when converting to SQL (done inside `convert_relation_to_sql`). I realized we should probably use it in general, but for now can you make it so `pretty=True` is passed in when `display_sql` is True (that way it only affects the one test you wrote that invokes it)? Down the line we can make it do so for all instances where the user will see the SQL.

knassre-bodo (commented):
Great job @vineetg3! Had a few revisions for you to do, but an excellent first PR!

pydough/logger/logger.py
vineetg3: Is there a specific reason to import it to the main `__init__.py`?

knassre-bodo: That way, you can do `from pydough import get_logger`, which makes sense because this seems like the type of utility you'd want to be directly available to users. For internal usage you can still use `from pydough.logger import get_logger` to avoid circularity issues.

vineetg3: Got it thank you. I'll make the change.

knassre-bodo (approved):
Nearly LGTM! Fantastic job @vineetg3. My remaining comments are mostly nitpicks / documentation.

documentation/usage.md
```diff
@@ -730,3 +731,22 @@ This term is singular with regards to the collection, meaning it can be placed i
...
 ```
+## Logging
+
+Logging is enabled and set to INFO level by default. We can change the log level by setting the environment variable `PYDOUGH_LOG_LEVEL` to the standard levels: DEBUG, INFO, WARNING, ERROR, CRITICAL.
+
+A new `logger` object can be created using `get_logger`.
+This function configures and returns a logger instance. It takes the following arguments:
+
+- `name` : The logger's name, typically the module name (`__name__`).
+- `default_level` : The default logging level if not set externally via environment variable `PYDOUGH_LOG_LEVEL`.
```
hadia206, line 742: Let's mention again that `INFO` is the default value

documentation/usage.md
```diff
@@ -730,3 +731,22 @@ This term is singular with regards to the collection, meaning it can be placed i
...
+- `fmt` : An optional log message format compatible with Python's logging.
+- `handlers` : An optional list of logging handlers to attach to the logger.
+
+It returns a configured `logging.Logger` instance.
+
+```py
+from pydough import get_logger
+pyd_logger = get_logger(__name__)
+#add examples of using the logger
```
hadia206, line 751: shouldn't we add an example of using it here?

pydough/logger/README.md
```diff
@@ -0,0 +1,7 @@
+# Logger
+
+This directory of PyDough deals with logger utilities.
+
+## Available APIs
+
+- `get_logger`: The main function to get a logger as defined in `./logger.py`.
```
hadia206, line 7: I should show full API with its arguments.

pydough/logger/__init__.py
hadia206: Reminder to address this

pydough/logger/logger.py
```diff
@@ -0,0 +1,49 @@
...
+        default_level (int): Default logging level if not set externally.
+        fmt (str): The format of the string compatible with python's logging library.
+        handlers (Optional[List[logging.Handler]]): A list of `logging.Handler` instances to add to the logger.
+    Returns:
+        logging.Logger: Configured logger instance.
+    """
+    logger = logging.getLogger(name)
+    level_env_var = "PYDOUGH_LOG_LEVEL"
+    level = os.getenv(level_env_var, default_level)
+    if isinstance(level, str):
```
hadia206, line 31: What happens if it's not a string or a string that is not one of the expected values?

hadia206 (approved):
I have minor comments. Otherwise looks good to me.
Great job on your 1st PR Vineet :)

pydough/logger/logger.py
vineetg3: It falls back to the default value. Should I throw a Value and Type error or just log as a warning?

hadia206: log as warning is fine.

knassre-bodo: I think we might as well do:
```py
assert isinstance(level, str), f"expected environment variable 'PYDOUGH_LOG_LEVEL' to be a string, found {level.__class__.__name__}"
```

PS: I forgot this in my earlier review, but wherever possible can you make variables defined in functions have type annotations? This serves two benefits:
- Good for review / code archeology from the GitHub UI since you won't have the same tooltip highlighting that you get form an IDE
- mypy enforces these annotations strongly, so adding annotations increases the chances that any bugs/errors in the code can be caught by mypy isntead of you having to track them down yourself.

pydough/logger/README.md
knassre-bodo: Since we describe this elsewhere, we can also just provide a relative link to that section of `usage.md` for details on how this API works.

PR #234, opened 2025-01-28
Support uses of BACK that cause correlated references: SQLGlot conversion

knassre-bodo:
Child PR of #269 that is part of addressing #141. Adds the SQLGlot conversion step. With these changes, all SQLGlot correlation queries are functional except for:
- TPC-H queries: 5 & 22
- Correl queries: 1, 2, 3, 6, 8, 9, 15, 17, 18, 19, 20

These remainders need to be handled via the decorrelation cases in the comments of #141. Specifically:
- Singular: Correl 8
- Aggregation:  TPC-H Q5, TPC-H Q22, Correl 1, Correl 2, Correl 3, Correl 18, Correl 19
- Semi-Singular: Correl 9, Correl 17, Correl 20
- Semi-Aggregation: Correl 6


pydough/sqlglot/sqlglot_relational_visitor.py
```diff
@@ -154,11 +159,22 @@ def _try_merge_columns(
...
+                elif isinstance(new_column, SQLGlotColumn):
+                    expr = set_glot_alias(
+                        old_column_map[new_column.this.this], new_name
+                    )
+                    modified_old_columns.append(expr)
+                    if isinstance(expr, Identifier):
+                        seen_cols.add(expr)
+                else:
+                    raise ValueError(
+                        f"Unsupported expression type for column merging: {new_column.__class__.__name__}"
```
amullerbodo, line 176: are you able to provide column name here?

amullerbodo (commented):
no major things to share here.

pydough/sqlglot/sqlglot_relational_visitor.py
knassre-bodo: A column name doesn't necessarily exist in this case (depends on what type `new_column` is and this is the catch-all for _all_ unsupported types). This case should never occur by construction of the `_is_mergeable_column` API which prevents anything except these cases.

vineetg3 (approved):
LGTM.

PR #244, opened 2025-02-03
Update column name generation in relational conversion

knassre-bodo:
Ancillary task related to #141. Changes how column names are handled in CALC so existing names are not overriden, but instead a new name is used. This becomes crucial for when correlated references are implemented, but until then causes harmless naming changes to many relational plans. These changes are benign because they are not observable in the final answer, and only change the internal names.


tests/test_plan_refsols/join_asia_region_nations.txt
```diff
@@ -1,5 +1,5 @@
-ROOT(columns=[('key', key), ('name', name), ('region_key', region_key), ('comment', comment)], orderings=[])
- PROJECT(columns={'comment': comment_1, 'key': key_2, 'name': name_3, 'region_key': region_key})
+ROOT(columns=[('key', key_6), ('name', name_7), ('region_key', region_key), ('comment', comment_5)], orderings=[])
+ PROJECT(columns={'comment_5': comment_1, 'key_6': key_2, 'name_7': name_3, 'region_key': region_key})
```
knassre-bodo, line 2: Notice that the changes here are just to the internal names used. These names got scrambled because some of them overlapped with other column names (but those other columns later got pruned). E.g. in this example, there were multiple columns named `comment`, `key` and `name` from multiple tables (region vs nation), but only 1 of each got kept in the final answer (in this case, the ones from the nation). However, the names were adjusted before the duplicates got pruned.

pydough/conversion/relational_converter.py
```diff
@@ -681,6 +712,10 @@ def translate_calc(
             rel_expr: RelationalExpression = self.translate_expression(
                 hybrid_expr, context
             )
+            # Ensure the name of the new column is not already being used. If
+            # it is, choose a new name.
```
vineetg3, line 716: NIT: Clarify that the new name is basically mimicking the previous name with an index. This helps in debugging purposes.

pydough/conversion/relational_converter.py
```diff
@@ -78,7 +78,17 @@ class TranslationOutput:
     """
 
     relation: Relational
```
vineetg3, line 80: As a follow-up can we rename the class `Relational` to `RelationalNode` so that there is a clarity that this is a node?

vineetg3: Moreover, in the same lines, we can rename the field from `relation` to `relational_node`.

vineetg3 (approved):
My comments are mostly Nits for clarity purposes. Otherwise, good.

hadia206 (approved):

Issue #247, opened 2025-02-06
Ensure that pretty SQL is always used when invoking SQLGlot


knassre-bodo:
When #163 was implemented, the `pretty=True` argument was added to the `.sql()` call to SQLGlot when displaying SQL in the logger. Goal of this task: ensure `pretty=True` is **always** used so the generated SQL is always pretty, regardless of whether the SQL is being logged or not. This is useful when debugging, or when a runtime error happens because the generated SQL was bad (because in its current form, its largely unreadable). Any tests that case on the outputted SQL will also need to be changed.

knassre-bodo:
Resolved by #248

PR #248, opened 2025-02-06
Changing SQL conversion to always use pretty=True

knassre-bodo:
This drastically increases the readability of the generated SQL text when debugging, especially when errors are caused during runtime execution of the query without a clear cause unless you look at the generated SQL text. Since this changed a lot of tests, also moved all of the SQL generation tests to SQL files in a folder that either get read or written to depending on the same environment variable as planner tests.


tests/test_pipeline.py
```diff
@@ -754,15 +754,26 @@ def test_pipeline_e2e_errors(
                 lambda: pd.DataFrame(
                     {
                         "transaction_id": [
-                            "TX001", "TX005", "TX011", "TX015", "TX021", "TX025",
```
knassre-bodo, line 757: These changes are just the formatter.

knassre-bodo:
Resolves #247

tests/test_pipeline.py
hadia206: For future, you can add format checker on CI. I believe we have that in the engine.

tests/test_session.py
```diff
@@ -124,6 +124,11 @@ def test_active_session_to_sql(sample_graph_path: str) -> None:
     Verify that the active session can generate SQL by default
     without any configuration.
     """
+    output_query: str = """
+SELECT
```
hadia206, line 128: nit: Formatting seems off. Shouldn't this have an indentation?

hadia206 (approved):
Looks good to me

tests/test_pydough_to_sql.py
```diff
@@ -25,42 +25,41 @@
...
         pytest.param(
             rank_a,
-            "SELECT ROW_NUMBER() OVER (ORDER BY acctbal DESC NULLS FIRST) AS rank FROM (SELECT c_acctbal AS acctbal FROM tpch.CUSTOMER)",
+            "rank_a",
             id="rank_a",
         ),
         pytest.param(
             rank_b,
-            " SELECT RANK() OVER (ORDER BY order_priority NULLS LAST) AS rank FROM (SELECT o_orderpriority AS order_priority FROM tpch.ORDERS)",
+            "rank_b)",
```
vineetg3, line 47: Is the parenthesis deliberate? I see that `rank_b).sql` file has been generated.

vineetg3 (changes requested):

tests/test_session.py
knassre-bodo: No, because is like this:
```sql
SELECT
  o_orderkey AS key
FROM tpch.ORDERS
```

Its all the way to the left because its a multiline string, so if I indented it to match the Python code it would be adding _extra_ indents like this, which is incorrect:
```sql
    SELECT
      o_orderkey AS key
    FROM tpch.ORDERS
```

tests/test_pipeline.py
knassre-bodo: I'll add that

knassre-bodo: https://github.com/bodo-ai/PyDough/pull/252

PR #251, opened 2025-02-06
Support uses of BACK that cause correlated references: setup decorrelation handling

knassre-bodo:
Child PR of #269 that is part of addressing #141. Adds the handling of de-correlation as a post-processing step of the hybrid tree before relational conversion starts. Handles the singular & aggregate cases as described in the issue, and handles the semi-singular and semi-aggregate cases in the same manner as their non-semi equivalents (for now), saving the optimized variants for another day.

These remaining still need to be addressed in the next follow up:
- TPC-H Q22: error during relational conversion that likely means a bug happened during hybrid conversion or hybrid decorrelation
- Correl 15: error during decorrelation causing over-excessive column pruning
- Correl 18: tbd bug in decorrelation
- Correl 19: bug in name resolution handling that causes the wrong term to be used when certain property names collide


pydough/pydough_operators/base_operator.py
```diff
@@ -80,3 +80,15 @@ def to_string(self, arg_strings: list[str]) -> str:
...
+    def equals(self, other: object) -> bool:
+        """
+        Returns whether this operator is equal to another operator.
+        """
+
+    def __eq__(self, other: object) -> bool:
+        return self.equals(other)
+
+    def __hash__(self) -> int:
+        return hash(repr(self))
```
knassre-bodo, line 94: This was missing before and was causing all manner of havoc until I added these in. I'm a bit flabbergasted that this never raised a problem until now, and only on one test that was using `LOWER` but couldn't match it (these operators get used as dictionary keys). I'll chalk it up to one of those "sometimes bad things work in Python until they don't" moments 🤷

knassre-bodo:
As of this stage, all queries are passing except these 3 (which seem like high priority bugs to fix):
- TPC-H Q22: error during relational conversion that likely means a bug happened during hybrid conversion or hybrid decorrelation
- Correl 3: error during decorrelation causing over-excessive column pruning
- Correl 15: not being decorrelated

tests/test_plan_refsols/tpch_q5.txt
```diff
@@ -1,21 +1,26 @@
...
-       JOIN(conditions=[t0.supplier_key == t1.key], types=['left'], columns={'discount': t0.discount, 'extended_price': t0.extended_price, 'name_9': t1.name_9, 'nation_key': t0.nation_key})
-        JOIN(conditions=[t0.key_5 == t1.order_key], types=['inner'], columns={'discount': t1.discount, 'extended_price': t1.extended_price, 'nation_key': t0.nation_key, 'supplier_key': t1.supplier_key})
-         FILTER(condition=order_date >= datetime.date(1994, 1, 1):date & order_date < datetime.date(1995, 1, 1):date, columns={'key_5': key_5, 'nation_key': nation_key})
-          JOIN(conditions=[t0.key == t1.customer_key], types=['inner'], columns={'key_5': t1.key, 'nation_key': t0.nation_key, 'order_date': t1.order_date})
-           SCAN(table=tpch.CUSTOMER, columns={'key': c_custkey, 'nation_key': c_nationkey})
+    AGGREGATE(keys={'key': key}, aggregations={'agg_0': SUM(value)})
+     PROJECT(columns={'key': key, 'value': extended_price * 1:int64 - discount})
+      FILTER(condition=name_15 == name, columns={'discount': discount, 'extended_price': extended_price, 'key': key})
```
knassre-bodo, line 11: Notice here, the filter `name_9 == corr10.name` has been replaced with `name_15 == name`, where `name_15` refers to the supplier nation name & `name` refers to the customer nation name. Previously, `corr10.name` was required since the customer natior name was on the left side of the join, and the supplier nation name was on the right side. But now they are both on the right side, so no correlation is required.

tests/test_plan_refsols/tpch_q5.txt
```diff
@@ -1,21 +1,26 @@
...
-     PROJECT(columns={'nation_key': nation_key, 'value': extended_price * 1:int64 - discount})
-      FILTER(condition=name_9 == corr10.name, columns={'discount': discount, 'extended_price': extended_price, 'nation_key': nation_key})
-       JOIN(conditions=[t0.supplier_key == t1.key], types=['left'], columns={'discount': t0.discount, 'extended_price': t0.extended_price, 'name_9': t1.name_9, 'nation_key': t0.nation_key})
-        JOIN(conditions=[t0.key_5 == t1.order_key], types=['inner'], columns={'discount': t1.discount, 'extended_price': t1.extended_price, 'nation_key': t0.nation_key, 'supplier_key': t1.supplier_key})
-         FILTER(condition=order_date >= datetime.date(1994, 1, 1):date & order_date < datetime.date(1995, 1, 1):date, columns={'key_5': key_5, 'nation_key': nation_key})
-          JOIN(conditions=[t0.key == t1.customer_key], types=['inner'], columns={'key_5': t1.key, 'nation_key': t0.nation_key, 'order_date': t1.order_date})
-           SCAN(table=tpch.CUSTOMER, columns={'key': c_custkey, 'nation_key': c_nationkey})
+    AGGREGATE(keys={'key': key}, aggregations={'agg_0': SUM(value)})
```
knassre-bodo, line 9: The aggregation keys now changed. Previously we could aggregate on the customer's nation_key value, but now we aggregate on the nation's key value. In this case those are the same thing, but the distinction matters because now we need to guarantee that the join between the left side & right side is a 1:1 zippering without any warping of cardinality.

pydough/conversion/hybrid_tree.py
```diff
@@ -184,6 +184,8 @@ def apply_renamings(self, renamings: dict[str, str]) -> "HybridExpr":
         return self
 
     def shift_back(self, levels: int) -> HybridExpr | None:
+        if levels == 0:
+            return self
```
knassre-bodo, line 188: This is just a useful trick to have rather than having to case on this behavior elsewhere, so `x.shift_back(0)` returns `x`.

pydough/conversion/hybrid_tree.py
```diff
@@ -895,6 +897,7 @@ def __init__(
         self._is_connection_root: bool = is_connection_root
         self._agg_keys: list[HybridExpr] | None = None
         self._join_keys: list[tuple[HybridExpr, HybridExpr]] | None = None
+        self._correlated_children: set[int] = set()
```
knassre-bodo, line 900: This set of indices indicates **which** of the elements in `_children` contain a correlated reference specifically to `self`.

pydough/conversion/hybrid_tree.py
```diff
@@ -1584,36 +1595,19 @@ def make_hybrid_correl_expr(
...
+            ).expr
             self.stack.pop()
-            self.stack.append(parent_tree)
-            # Then, postprocess the output to account for the fact that a
-            # BACK level got skipped due to the change in subtree.
-            match parent_result.expr:
-                case HybridRefExpr():
-                    parent_result = HybridBackRefExpr(
-                        parent_result.expr.name, 1, parent_result.typ
-                    )
```
knassre-bodo, line 1606: This weird extra-shifting trick is no longer needed, and was actually incorrect.

pydough/conversion/hybrid_tree.py
```diff
@@ -1634,6 +1628,8 @@ def make_hybrid_correl_expr(
                 collection, back_expr.term_name, remaining_steps_back
             )
             parent_result = self.make_hybrid_expr(parent_tree, new_expr, {}, False)
+        if not isinstance(parent_result, HybridCorrelExpr):
+            parent_tree.correlated_children.add(len(parent_tree.children))
```
knassre-bodo, line 1632: This is how we know which children of the hybrid tree are correlated. If `parent_result` is a correl expr, it means that the correlated reference is actually to the tree **containing** `parent_tree`.

tests/test_plan_refsols/correl_1.txt
```diff
@@ -1,8 +1,10 @@
 ROOT(columns=[('name', name), ('n_prefix_nations', n_prefix_nations)], orderings=[(ordering_1):asc_first])
  PROJECT(columns={'n_prefix_nations': n_prefix_nations, 'name': name, 'ordering_1': name})
   PROJECT(columns={'n_prefix_nations': DEFAULT_TO(agg_0, 0:int64), 'name': name})
-   JOIN(conditions=[t0.key == t1.region_key], types=['left'], columns={'agg_0': t1.agg_0, 'name': t0.name}, correl_name='corr1')
+   JOIN(conditions=[t0.key == t1.key], types=['left'], columns={'agg_0': t1.agg_0, 'name': t0.name})
```
knassre-bodo, line 4: This is a "simple" example of how the de-correaltion works.
- The left subtree (region) became a part of the right subtree (nation) before the correlated reference (the filter) occurred. This results in selecting region twice, but that's an efficiency problem for another day.
- The join is now on the uniquness keys of the left subtree: the region's `.key` property.

pydough/conversion/hybrid_decorrelater.py
```diff
@@ -0,0 +1,330 @@
...
+                case (
+                    ConnectionType.SEMI
+                    | ConnectionType.ANTI
+                    | ConnectionType.NO_MATCH_SINGULAR
+                    | ConnectionType.NO_MATCH_AGGREGATION
+                    | ConnectionType.NO_MATCH_NDISTINCT
+                ):
+                    # These patterns do not require decorrelation since they
+                    # are supported via correlated SEMI/ANTI joins.
+                    continue
```
amullerbodo, line 311: do we need a default "case _:" here?

pydough/pydough_operators/base_operator.py
amullerbodo, line 94: any other way we could generate this hash? If someone changes __repr__ inadvertently somewhere else it could break things. Not sure it is something we should be worried about at this stage.

amullerbodo (commented):
general thoughts about hashes and equality.

pydough/conversion/hybrid_decorrelater.py
knassre-bodo: No because these are exhaustive with regards to the enum. If any new case is generated, mypy will throw an exception because these branches are no longer exhaustive.

pydough/pydough_operators/base_operator.py
knassre-bodo: I can't think of a better way since the best way to hash would be in a manner that depends on the name of the function/operator, which is always inside of its repr string (but each class handles that naming differently).

pydough/conversion/hybrid_decorrelater.py
```diff
@@ -0,0 +1,330 @@
...
+        is_aggregate: bool,
+    ) -> None:
+        """
+        Runs the logic to de-correlate a child of a hybrid tree that contains
+        a correlated reference. This involves linking the child to a new parent
+        as its ancestor, the parent being a snapshot of the original hybrid
+        tree that contained the correlated child as a child. The transformed
+        child can now replace correlated references with BACK references that
+        point to terms in its newly expanded ancestry, and the original hybrid
+        tree cna now join onto this child using its uniqueness keys.
```
vineetg3, line 232: Typo : cna

vineetg3 (approved):
Few Typos, but LGTM!

PR #252, opened 2025-02-06
Add pre-commit CI to PyDough

### LINKED ISSUE #250: Add pre-commit CI to PyDough
Ensure that the pre-commit hooks are checked during PR CI using the pre-commit CI flow: https://github.com/marketplace/pre-commit-ci. This prevents new developers from accidentally pushing PRs without formatting because they have not installed pre-commit.

Also, should enable the autoupdate with a monthly cadence feature.

knassre-bodo:
Resolves #250. See issue for more details.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,6 @@
+ci:
+  autoupdate_schedule: monthly
+
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.6.7
--- a/pydough/__init__.py
+++ b/pydough/__init__.py
@@ -8,11 +8,11 @@ __all__ = [
     "explain",
     "explain_structure",
     "explain_term",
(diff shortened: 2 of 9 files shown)


.pre-commit-config.yaml
```diff
@@ -1,3 +1,6 @@
+ci:
+  autoupdate_schedule: monthly
```
hadia206, line 2: monthly seems too long.
What about weekly?

hadia206 (approved):
Nice 👍

.pre-commit-config.yaml
knassre-bodo: This is what we use for Bodo, and its just for checking if any of the pre-commit dependencies got changed.

hadia206: Oh I see. Monthly is fine then

vineetg3 (approved):

PR #253, opened 2025-02-07
Adding a Jupyter Notebook guide to the PyDough documentation

aramirez-bodo:
Information added:

Introduced a new Jupyter Notebook guide for using PyDough effectively. Walkthrough of setting up the PyDough
Explained key PyDough features with code snippets and example use cases. Added step-by-step instructions for integrating PyDough with SQL-based workflows.

Feedback Incorporation:
Enhanced query clarity and consistency across sections, including adding context to later queries (aligned with earlier examples).
Simplified query syntax, removing unnecessary comparisons like ==1 with HASNOT, and streamlined SUM(IFF(X, 1, 0)) to SUM(X) for better efficiency.
Removed redundant use of PARTITION by aggregating at the collection level where possible (e.g., per-nation and per-region).


demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "%%pydough\n",
+    "\n",
+    "filter_c= nations(key, name, comment)\n",
+    "\n",
+    "pydough.to_df(filter_c)\n"
```
knassre-bodo, line 276: Two things here:
- What is `filter_c`? Can this name be changed to something that makes sense in the context of the question?
- Technically the calc is unnecessary; can just do `nations` as the equivalent of `SELECT *` (the code you wrote is technically not a `SELECT *` since it doesn't select the `region_key`)

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+    "SELECT * FROM nation; \n",
+    "```\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In PyDough, we can perform the same operation using the nation collection, which retrieves all documents from a collection. \n",
+    "A collection in PyDough is an abstraction for any \"document\", but in most cases represents a table. If we want to access the nations table, we will use our corresponding PyDough collection."
```
knassre-bodo, line 48: May want to note that the column names of the table may have been tweaked (e.g. `n_name` and `n_key` became `name` and `key`)

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+   "metadata": {},
+   "source": [
+    "### 2. Find Nations Whose Name Starts with \"A\"\n",
+    "Next, we’ll use WHERE and LIKE to filter nations whose names start with the letter \"A\".\n",
+    "\n",
+    "```SQL\n",
+    "SELECT C_NAME, C_ACCTBAL\n",
+    "FROM n\n",
+    "WHERE C_NAME LIKE 'A%';\n",
+    "```"
```
knassre-bodo, line 297: This query does not match with the PyDough code. Should it be the following?
```sql
SELECT n_name, n_comment
FROM nation
WHERE n_name LIKE 'A%'
```

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+    "The next situation involves a client seeking to identify all customers from a specific country, Peru.\n",
+    "\n",
+    "SQL example: \n",
+    "\n",
+    "```SQL\n",
+    "SELECT C.C_NAME\n",
+    "FROM customers C\n",
+    "JOIN nation N\n",
+    "ON C.C_NATIONKEY = N.N_NATIONKEY\n",
+    "WHERE N.N_NAME = 'Peru';\n"
```
knassre-bodo, line 340: - The table name is `customer`, not `customers`
- Should be `= 'PERU'` since the nation names in the SQL table are capitalized.

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "%%pydough\n",
+    "\n",
+    "peru_nations= customers.WHERE(nation.name == \"PERU\")\n",
+    "\n",
+    "customers_from_peru= peru_nations(n_name=name)\n",
+    "\n",
```
knassre-bodo, line 459: NIT: a lot of these code blocks dont' have a space before the `=`. Can they all be reformatted like this?
```py
%%pydough

peru_nations = customers.WHERE(nation.name == "PERU")

customers_from_peru = peru_nations(n_name=name)

pydough.to_df(customers_from_peru)
```

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+    "FROM n\n",
+    "WHERE C_NAME LIKE 'A%';\n",
+    "```"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In PyDough, the WHERE statement is similar to the WHERE operation in SQL. It can be used to filter unwanted entries in a context. You can use the STARTSWITH() method to match patterns or LIKE() using the same syntax as SQL. Here's how you would do it:"
```
knassre-bodo, line 304: NIT: `STARTSWITH()` is a function, not a method.

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+       "[5975 rows x 1 columns]"
+      ]
+     },
+     "execution_count": 43,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "%%pydough\n",
```
knassre-bodo, line 454: Since you pose alternatives for some other blocks, I thought I'd mention another way to write this one:
```
customers_from_peru = nations.WHERE(name == "PERU").customers(name)
```

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+    "\n",
+    "peru_nations= customers.WHERE(nation.name == \"PERU\")\n",
+    "\n",
+    "customers_from_peru= peru_nations(n_name=name)\n",
```
knassre-bodo, line 458: Also why does that say `n_name`?

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+    "pydough.to_df(customers_from_peru)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 4. Get a specific customer\n",
+    "\n",
+    "The next situation involves analyzing a specific set of customers based on several conditions. The goal is to identify customers who are in debt, meaning their account balance is negative, and have placed more than 5 orders. Additionally, the focus is on customers from America, excluding those from Brazil."
```
knassre-bodo, line 469: You said "more than 5 orders", but the query you wrote says `>= 5` which means it will allow _at least_ 5 orders.

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This query filters customers based on specific conditions related to their account balance, order count, and geographical region. It checks if a customer’s account balance is negative, if they have made at least 5 orders, if their region is \"AMERICA,\" and if they are not from Brazil. The WHERE clause applies all these conditions using & (AND) to ensure that all must be true for a customer to be included in the results.\n",
+    "\n",
+    "PyDough does not yet support the AND, OR, NOT, IN expressions, as well as trying in-between comparisons like (1 < x < 5)"
```
knassre-bodo, line 478: Is there no SQL query for this one?

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+   "source": [
+    "### 5. Find the total number of orders per customers placed in 1998\n",
+    "\n",
+    "The next situation consists of analyzing customer order activity from the year 1998. The query retrieves customer details, specifically their customer keys and names, while also counting how many orders each customer placed in 1998. The goal is to identify which customers were the most active that year, ordering them by the total number of orders in descending order\n",
+    "\n",
+    "```SQL\n",
+    "SELECT c.c_custkey, c.c_name, COUNT(o.o_orderkey) AS total_orders\n",
+    "FROM customer c\n",
+    "JOIN orders o ON c.c_custkey = o.o_custkey\n",
+    "WHERE strftime('%Y', o.o_orderdate) = '1998'  \n",
```
knassre-bodo, line 608: Two things:
- There should be a `GROUP BY` clause here (on custkey & name)
- Do we want to include customers with _zero_ orders from 1998? The PyDough code you wrote does, but this query you wrote won't. If so, the join should be written as follows:
```sql
FROM customer c
LEFT JOIN (SELECT o_custkey FROM orders WHERE strftime('%Y', o.o_orderdate) = '1998') o
ON c_custkey = o_custkey
```

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+    "    ),\n",
+    ").ORDER_BY(num_orders.DESC())\n",
+    "pydough.to_df(customer_order_counts )\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This demonstrates an alternative approach to solving the same query by leveraging date and time filtering in the orders table. In this case, we define a specific date range to filter orders, selecting only those with an order_date between January 1, 1998 (inclusive) and January 1, 1999 (exclusive). The WHERE clause applies the condition directly to the orders table, and the key parameter specifies how the results should be grouped or indexed after filtering."
```
knassre-bodo, line 764: > and the key parameter specifies how the results should be grouped or indexed after filtering

Thats not quite accurate. The idea here is that we're selecting, for every customer, their name, key, and # of orders made in 1998, just like in the first version. The grouping is all done under-the-hood, not based on what was written here (what it actually does is do the grouping/aggregation **before** the join, using `o_custkey` as the grouping key). If you want to learn more about this, see the "hybrid tree" and "relational converter" stages of the pipeline in the "conversion" module (though I'd advise against it for now, since those are arguably the _nastiest_ parts of the codebase). 

I would reccomend either deleting this alternative, or just focusing on using `>=` and `<` as opposed to `YEAR`.

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+    "JOIN orders o ON c.c_custkey = o.o_custkey\n",
+    "GROUP BY n.n_name\n",
+    "ORDER BY order_count DESC;\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In PyDough, to achieve this, we would use the metadata relationship and the PARTITION method, which is equivalent to GROUP BY in SQL. Keys can be specified using the by argument, which is the element to be grouped, and data columns to be aggregated can be referenced using the name argument."
```
knassre-bodo, line 932: This example is not one where `PARTITION` is needed. Can just aggregate like this:

```py
nations(name, order_count=COUNT(customers.orders)).ORDER_BY(order_count.DESC())
```
Here, the grouping is implied to be per-nation, but the grouping keys are a detail that the user needn't worry about. Under the hood, what is happening is that customers/orders get joined, then grouped on `c_nationkey` to compute `COUNT(*)`, then that result is joined with nations.

knassre-bodo: In general, you only use `PARTITION` when you are grouping per-XXX, where XXX is not an existing collection but rather a specific combination of fields in 1 or more collections. The next query you wrote below this is a **perfect** example of where `PARTITION` needs to be used.

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+    ")\n",
+    "\n",
+    "pydough.to_df(high_value_customers_in_asia)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 9. Average Order Value by Region \n",
```
knassre-bodo, line 1632: This is another example where `PARTITION` is not needed. Can rewrite as:
```py
regions(region_name=name, avg_order_value=SUM(nations.customers.total_price))
```

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+    ")\n",
+    "\n",
+    "pydough.to_df(region_revenue_summary)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 10. Identifying the 5 Most Profitable Regions\n",
```
knassre-bodo, line 1753: This query & the PyDough code don't match up.
- Did you mean to find the 5 most profitable **nations**? Because there are only 5 total regions
- You don't have a TOP_K in the PyDough code even though you mention it in the text

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+    "JOIN orders o ON o.o_custkey = c.c_custkey\n",
+    "GROUP BY r.r_name, n.n_name\n",
+    "ORDER BY order_value_difference DESC;\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In PyDough, to achieve this we first select customers based on their region and nation using selected_customers. Then, we filter their associated orders and calculate the total price for each order. Using PARTITION, we group the data by region and nation and calculate the maximum and minimum order values using the MAX and MIN functions, their difference, and the total number of orders. Finally, we sort the results by the order value difference in descending order using ORDER_BY."
```
knassre-bodo, line 2009: This description doesn't quite match the code

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+    "%%pydough\n",
+    "# Retrieve customers' order details for each nation, calculating max, min, and order value differences\n",
+    "selected_customers_by_nation = nations(\n",
+    "    region_name=region.name,  \n",
+    "    nation_name=name,  \n",
+    "    max_order_value=MAX(customers.orders.total_price), \n",
+    "    min_order_value=MIN(customers.orders.total_price),  \n",
+    "    order_value_difference=MAX(customers.orders.total_price) - MIN(customers.orders.total_price), \n",
+    "    total_orders=COUNT(customers.orders.total_price)  \n",
+    ")\n",
```
knassre-bodo, line 2347: Can use variables to help out here (and don't need to do `.total_price` when doing `COUNT` if you want an equivalent of `COUNT(*)`):
```py
order_prices = customers.orders.total_price
smallest_order_value = MIN(order_prices)
largest_order_value = MAX(order_prices)
selected_customers_by_nation = nations(
    region_name=region.name,  
    nation_name=name,  
    max_order_value=largest_order_value,
    min_order_value=smallest_order_value,
    order_value_difference=largest_order_value - smallest_order_value,
    total_orders=COUNT(customers.orders)  
)
```

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+    "sorted_customers_by_order_value_difference = selected_customers_by_nation.ORDER_BY(order_value_difference.DESC())\n",
+    "\n",
+    "pydough.to_df(sorted_customers_by_order_value_difference)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 12. Nations with the Most Customers in a Specific Market Segment\n",
```
knassre-bodo, line 2359: This one doesn't need `PARTITION` either. Can rewrite as:
```py
customer_in_target_mktsegment = customers.WHERE(ISIN(mktsegment, ('MACHINERY', 'AUTOMOBILE')))
nation_customer_count = nations(
    nation_name=name, 
    customer_count=COUNT(customer_in_target_mktsegment)  
).ORDER_BY(customer_count.DESC())  
```

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+    "\n",
+    "\n",
+    "pydough.to_df(nation_customer_count)\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 14. Region with the Highest Percentage of High-Priority Orders\n",
```
knassre-bodo, line 2602: This one also doesn't need `PARTITION` since you are aggregating per-region:
```py
high_priority_orders = ISIN(nations.orders.order_priority, ('1-URGENT', '2-HIGH'))
region_priority_summary = regions(
    region_name=name, 
    high_priority_percentage=ROUND(
        (SUM(high_priority_orders) * 100) / COUNT(nations.orders), 2
    ) 
).ORDER_BY(high_priority_percentage.DESC())
```

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+    "FROM customer c\n",
+    "LEFT JOIN orders o ON c.c_custkey = o.o_custkey\n",
+    "WHERE o.o_orderkey IS NULL;\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In Pydough we use the function HAS/HASNOT to resolve the is null statement. HAS returns True if at least one record of the sub-collection exists and HASNOT returns True if at least one record of the sub-collection does'nt exists. So the steps to follow are first filtering the customers who have not placed any orders using the WHERE clause combined with the HASNOT function. This identifies the customers who have no associated orders. Then, we select these customers by retrieving their unique customer_key and customer_name. "
```
knassre-bodo, line 2750: You can also explain this in terms of EXISTS & NOT EXISTS (which the planner will usually turn into left-joins, under the hood):
```sql
SELECT c.c_custkey, c.c_name
FROM customer c
WHERE NOT EXISTS(
  SELECT *
  FROM ORDERS o
  WHERE c.c_custkey = o.o_custkey
)
```

Under the hood, PyDough turns `HAS` and `HASNOT` calls into SEMI and ANTI joins respectively, which then get turned into `EXISTS` and `NOT EXISTS` by SQLGlot, depending on the dialect of SQL being used.

demos/notebooks/6_pydough_TCPH_guide.ipynb
```diff
@@ -0,0 +1,3391 @@
...
+     },
+     "execution_count": 6,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "%%pydough\n",
+    "\n",
+    "# Retrieve customers by nation, classifying them into active and inactive based on order history\n",
```
knassre-bodo, line 3161: The `PARTITION` and `NDISTINCT` are both unnecessary here. Can rewrite as:
```py
cust_info = customers(is_active=HAS(orders))
nation_customer_summary = nations(
  nation_name = name,
  total_customers = COUNT(cust_info),
  active_customers = SUM(customers.is_active),
  inactive_customers = COUNT(cust_info) - SUM(customers.is_active),
).ORDER_BY(total_customers.DESC())
```

If you want an example where `NDISTINCT` is required, consider something like this:
```py
# Which 10 customers have purchased the most unique parts?
customers(name, n_parts=NDISTINCT(orders.lines.part.key)).TOP_K(10, by=n_parts.DESC())
```
Here, we need `NDISTINCT` because if we used `COUNT`, we could count a single part more than once, since each customer can have multiple orders/lines, but different lines can share the same part.

If you want another example where `PARTITION` is required, consider something like this:
```py
# What is the largest amount spent by a single customer in a single year?
# Include the amount, the customer's name, and the year.
cust_year_groups = PARTITION(orders(year=YEAR(order_date)), name="o", by=(name, year))
result = cust_year_groups(name, year, total=SUM(o.total_price)).TOP_K(1, by=total)
```

knassre-bodo (commented):
Fantastic job Aramis, this notebook is filled with very useful examples! My comments are largely about clarity / textual corrections, though there are also a few cases where the PyDough code could be rearranged.

demos/notebooks/6_pydough_TCPH_guide.ipynb
vineetg3, line 478: For now, it is better to say "PyDough does not support the AND, OR, NOT, IN expressions", since we don't have current plans to support them.

knassre-bodo: Agreed; it would require a _much_ more aggressive form of Python "hacking" in order to support these syntaxes, since we could not reply on magic methods to do custom overloading of the operators.

PR #254, opened 2025-02-10
Support uses of BACK that cause correlated references: fix remaining decorrelation edge cases

knassre-bodo:
Child PR of #269 that is part of addressing #141. Handles certain edge cases, such as:
- When the de-correlation causes an aggregate to prune all of its outputs, creating an empty SELECT clause.
- Fixing a bug in how decorrelation of a PARTITION child results in a cartesian join instead of joining on the partition keys.
- Fixing a bug in renaming of hybrid children causing the wrong name overload to be chosen.
- Fixing nondeterministic join key ordering during decorrelation.
- Adjusting how BACK shifts for join/agg keys are handled so the levels takes into account levels that were skipped when snapshotting the parent.
- Adjusting the SQLGlot conversion for aggregations so if there is no aggfunc, it instead does `SELECT DISTINCT`
- Added 3 more correlation tests for very particular edge cases involving `PARTITION`, and also made some adjustments to the qualification/hybrid conversion of `PARTITION` to account for cases where a `PARTITION` node is the root of a child operator child access.


pydough/sqlglot/sqlglot_relational_visitor.py
```diff
@@ -511,7 +511,7 @@ def visit_limit(self, limit: Limit) -> None:
         self._stack.append(query)
 
     def visit_empty_singleton(self, singleton: EmptySingleton) -> None:
-        self._stack.append(Select().from_(values([()])))
+        self._stack.append(Select().select(SQLGlotStar()).from_(values([()])))
```
knassre-bodo, line 514: Without this change, you can end up with `SELECT FROM ...` clauses (without anything between `SELECT` and `FROM`), which is invalid.

tests/tpch_test_functions.py
```diff
@@ -494,16 +494,20 @@ def impl_tpch_q22():
     PyDough implementation of TPCH Q22.
     """
     selected_customers = Customers(cntry_code=phone[:2]).WHERE(
-        ISIN(cntry_code, ("13", "31", "23", "29", "30", "18", "17")) & HASNOT(orders)
+        ISIN(cntry_code, ("13", "31", "23", "29", "30", "18", "17"))
```
knassre-bodo, line 497: This fixes the query to be in line with how the TPC-H version is written (the `HASNOT(orders)` condition is **not** considered when finding the average)

pydough/conversion/relational_converter.py
```diff
@@ -883,11 +883,26 @@ def rel_translation(
...
                     if context is not None:
+                        # If the collection access is the child of something
+                        # else, join it onto that something else. Use the
+                        # uniqueness keys of the ancestor, which should also be
+                        # present in the collection (e.g. joining a partition
+                        # onto the original data using the partition keys).
+                        assert preceding_hybrid is not None
+                        join_keys: list[tuple[HybridExpr, HybridExpr]] = []
+                        for unique_column in sorted(
+                            preceding_hybrid[0].pipeline[0].unique_exprs, key=str
```
knassre-bodo, line 894: This case arises from `correl_18`.

pydough/conversion/hybrid_tree.py
```diff
@@ -520,7 +529,10 @@ class HybridFilter(HybridOperation):
 
     def __init__(self, predecessor: HybridOperation, condition: HybridExpr):
         super().__init__(
-            predecessor.terms, {}, predecessor.orderings, predecessor.unique_exprs
```
knassre-bodo, line 523: The use of `{}` here caused a bug in name handling, caught by `correl_19`.

pydough/conversion/hybrid_decorrelater.py
```diff
@@ -245,23 +260,26 @@ def decorrelate_child(
...
-            for unique_key in current_level.pipeline[0].unique_exprs:
+            skip_join: bool = (
+                isinstance(current_level.pipeline[0], HybridPartition)
+                and child is current_level.children[0]
+            )
+            for unique_key in sorted(current_level.pipeline[0].unique_exprs, key=str):
                 lhs_key: HybridExpr | None = unique_key.shift_back(additional_levels)
                 rhs_key: HybridExpr | None = unique_key.shift_back(
-                    additional_levels + child_height
+                    additional_levels + child_height - skipped_levels
```
knassre-bodo, line 272: This "skipped_levels" comes into play in TPCH query 22, since the PARTITION should not be considered in the height since it is a level that gets skipped by the parent snapshotting function.

pydough/conversion/hybrid_decorrelater.py
```diff
@@ -62,9 +64,10 @@ def make_decorrelate_parent(
             # case, all of the parent's children & pipeline operators should be
             # included in the snapshot.
             assert hybrid.parent is not None
-            return self.make_decorrelate_parent(
+            result = self.make_decorrelate_parent(
                 hybrid.parent, len(hybrid.parent.children), len(hybrid.pipeline)
             )
+            return result[0], result[1] + 1
```
knassre-bodo, line 70: Changing this to keep track of how many times we had to recursively step upward, because those levels should **not** be counted later.

pydough/conversion/hybrid_tree.py
```diff
@@ -494,17 +494,26 @@ def __init__(
...
+            ):
                 used_name = f"{name}_{idx}"
                 idx += 1
             terms[used_name] = expr
-            renamings[name] = used_name
+            new_renamings[name] = used_name
+        renamings.update(new_renamings)
+        for old_name, new_name in new_renamings.items():
+            expr = new_expressions.pop(old_name)
+            new_expressions[new_name] = expr
```
knassre-bodo, line 516: This change allows us to ensure that the terms in `new_expressions` _also_ get renamed when necessary.

pydough/conversion/hybrid_decorrelater.py
```diff
@@ -62,9 +64,10 @@ def make_decorrelate_parent(
             # case, all of the parent's children & pipeline operators should be
             # included in the snapshot.
             assert hybrid.parent is not None
```
amullerbodo, line 66: any other asserts we should write here regarding hybrid.parent ?

amullerbodo (commented):
checked the changes looks good from my end maybe I would find time to add some more asserts around the code base in general. Great work!

pydough/conversion/hybrid_decorrelater.py
knassre-bodo: Not really since the if statement above should cover it. In general, I only uses assertions when they are needed for mypy to not ocmplain about the typing (e.g. w/o this assertion, the next line would fail because `hybrid.parent.children` isn't legal if the type of `hybrid.parent` _could_ be None). They should never actually fail via data structure invariants (if those break, something catastrophic has ocurred), and anything that _could_ reasonably fail has a proper error message.

pydough/sqlglot/sqlglot_relational_visitor.py
```diff
@@ -472,7 +474,10 @@ def visit_aggregate(self, aggregate: Aggregate) -> None:
                 select_cols, input_expr, find_identifiers_in_list(select_cols)
             )
         if keys:
-            query = query.group_by(*keys)
+            if aggregations:
+                query = query.group_by(*keys)
+            else:
+                query = query.distinct()
```
knassre-bodo, line 480: This is so if we have an aggregate on keys `A, B, C` w/o any aggregation functions, we can just do `SELECT DISTINCT A, B, C FROM (...)`

tests/test_plan_refsols/correl_20.txt
```diff
@@ -2,7 +2,7 @@ ROOT(columns=[('n', n)], orderings=[])
  PROJECT(columns={'n': agg_0})
   AGGREGATE(keys={}, aggregations={'agg_0': COUNT()})
    FILTER(condition=domestic, columns={'account_balance': account_balance})
-    JOIN(conditions=[t0.key_9 == t1.key_21 & t0.order_key == t1.order_key & t0.line_number == t1.line_number & t0.key_5 == t1.key_17 & t0.key_2 == t1.key_14 & t0.key == t1.key], types=['left'], columns={'account_balance': t0.account_balance, 'domestic': t1.domestic})
+    JOIN(conditions=[t0.key_9 == t1.key_21 & t0.line_number == t1.line_number & t0.order_key == t1.order_key & t0.key_5 == t1.key_17 & t0.key_2 == t1.key_14 & t0.key == t1.key], types=['left'], columns={'account_balance': t0.account_balance, 'domestic': t1.domestic})
```
knassre-bodo, line 5: This is just a change in the ordering, caused by sorting to ensure determinism of the plans.

vineetg3 (approved):
LGTM!

tests/test_plan_refsols/correl_18.txt
```diff
@@ -10,10 +10,10 @@ ROOT(columns=[('n', n)], orderings=[])
      AGGREGATE(keys={'customer_key': customer_key, 'order_date': order_date}, aggregations={'agg_2': COUNT()})
       FILTER(condition=total_price_3 >= 0.5:float64 * total_price, columns={'customer_key': customer_key, 'order_date': order_date})
        FILTER(condition=YEAR(order_date_2) == 1993:int64, columns={'customer_key': customer_key, 'order_date': order_date, 'total_price': total_price, 'total_price_3': total_price_3})
-        JOIN(conditions=[True:bool], types=['inner'], columns={'customer_key': t0.customer_key, 'order_date': t0.order_date, 'order_date_2': t1.order_date, 'total_price': t0.total_price, 'total_price_3': t1.total_price})
+        JOIN(conditions=[t0.customer_key == t1.customer_key & t0.order_date == t1.order_date], types=['inner'], columns={'customer_key': t0.customer_key, 'order_date': t0.order_date, 'order_date_2': t1.order_date, 'total_price': t0.total_price, 'total_price_3': t1.total_price})
```
knassre-bodo, line 13: This was a **huge** bug previously and is fixed by the main block of changes in `relational_converter.py`. This was the original code:
```
cust_date_groups = PARTITION(
        Orders.WHERE(YEAR(order_date) == 1993),
        name="o",
        by=(customer_key, order_date),
)
selected_groups = cust_date_groups.WHERE(COUNT(o) > 1)(
        total_price=SUM(o.total_price),
)(n_above_avg=COUNT(o.WHERE(total_price >= 0.5 * BACK(1).total_price))) TPCH(n=SUM(selected_groups.n_above_avg))
```
The problem arises in joining `cust_date_groups` back to `Orders.WHERE(YEAR(order_date) == 1993)` (in order to derive `COUNT(o.WHERE(...))`. In this case, we need to re-derive `Orders.WHERE(YEAR(order_date) == 1993)` but this time it gets joined onto the aggregated data, therefore `Orders.WHERE(YEAR(order_date) == 1993)` has join keys to link it to the ancestor (described by `preceding_hybrid[0].pipeline[0].unique_exprs`).

PR #267, opened 2025-02-14
Support DATETIME function in PyDough

### LINKED ISSUE #208: Support a timestamp construction/alteration API in PyDough similar to the sqlite DATETIME function
This function should be able to do the following:
- Construct a datetime from its first argument which can be a datetime string, the word `'now'` (meaning get the current timestamp), or a datetime literal (`datetime.date`, `datetime.datetime`, `pandas.Timestamp`)
- Use its subsequent arguments as modifiers:
  - Add/subtract year/month/day/hour/minute/second intervals by string: `DATETIME('now', '-1 month', '+3 days')`)
  - Truncates to start of the year, month, day in SQLite (or to the hour, minute, second in ANSI): `DATETIME("now", "start of month")`
- For SQlite, uses the sqlite `datetime` function accordingly
- For ANSI, uses a mix of `CURRENT_DATE`, `CAST(... AS TIMESTAMP)`, `± INTERVAL 'x UNIT'`, and `DATE_TRUNC`

knassre-bodo:
Resolves #208. See issue for more details. Full list of additional changes:
- Ensured the `DATETIME` operator works for both sqlite & ANSI with very permissive spacing/capitalization/alias rules for the various string literals.
  - `now` (& similar aliases): for ANSI generates `CURRENT_TIMESTAMP()`, for sqlite generates `DATETIME('now')`
  - `start of <unit>` (year/month/day): for ANSI generates `DATE_TRUNC(<unit>, dt)`, for sqlite generates `DATETIME(dt, 'start of <unit>')`
  - `start of <unit>` (hour/minute/second): for ANSI generates `DATE_TRUNC(<unit>, dt)`, for sqlite generates `STRFTIME(fmt_str, dt)` where the `fmt_str` truncates to that unit.
  - `±<amt> <unit>`: for ANSI generates `DATEADD(<unit>, ±<amt>dt)`, for sqlite generates `DATETIME(dt, '±<amt> <unit>)`
- Moves a lot of the datetime unit logic in the transform bindings into a `DateTimeUnit` enum to handle common logic, such as converting a unit alias into a cannonical form or identifying the format string used to truncate to that unit.
- Added new fixture to create a DatabaseContext without a connection for each supported dialect
- Changed the tests in `test_pydough_to_sql.py` to be parameterized with the afformentioned new fixture so sqlite & ANSI are both tested
- Changed the refsol `.sql` files to contain the dialect (e.g. `xxx_ansi.sql` vs `xxx_sqlite.sql`)
- Fixed miscellaneous typos found in the repo
- Adjusted the SQLGlot generation of the empty singleton to avoid a malformed `VALUES` clause


tests/test_sql_refsols/simple_values_ansi.sql
```diff
@@ -2,4 +2,4 @@ SELECT
   42 AS A,
   'foo' AS B
 FROM (VALUES
-  ())
+  (NULL))
```
knassre-bodo, line 5: Just having `VALUES ()` is  invalid, so using `VALUES (NULL)` helps

tests/test_sql_refsols/hour_minute_day_sqlite.sql
```diff
@@ -0,0 +1,47 @@
+SELECT
```
knassre-bodo, line 1: These are now automatically generated along with the `XXX_ansi.sql` file for all the tests in `test_pydough_to_sql.py`, so we can verify that each supported dialect works on the expected functions.

tests/conftest.py
```diff
@@ -153,14 +154,14 @@ def impl(file_name: str) -> str:
 
 
 @pytest.fixture(scope="session")
-def get_sql_test_filename() -> Callable[[str], str]:
+def get_sql_test_filename() -> Callable[[str, DatabaseDialect], str]:
```
knassre-bodo, line 157: This change allows the SQL files to be dialect-specific.

tests/simple_pydough_functions.py
```diff
@@ -126,6 +127,209 @@ def function_sampler():
...
+            "-3 minutes",
+            "+1 second",
+        ),
+        d4=DATETIME(pd.Timestamp("2025-07-04 12:58:45"), "start of hour"),
+        d5=DATETIME(pd.Timestamp("2025-07-04 12:58:45"), "start of minute"),
+        d6=DATETIME(pd.Timestamp("2025-07-14 12:58:45"), "+ 1000000 seconds"),
+    )
+
+
+def datetime_sampler():
```
knassre-bodo, line 162: This doesn't cover _every_ edge case for the string parsing of `DATETIME`, but covers a large number of them that were generated.

pydough/sqlglot/transform_bindings.py
```diff
@@ -79,6 +176,182 @@ def impl(
...
+    if dialect == DatabaseDialect.SQLITE:
+        match unit:
+            # For y/m/d, use the `start of` modifier in SQLite.
+            case DateTimeUnit.YEAR | DateTimeUnit.MONTH | DateTimeUnit.DAY:
+                trunc_expr: SQLGlotExpression = sqlglot_expressions.convert(
+                    f"start of {unit.value}"
+                )
+                if isinstance(base, sqlglot_expressions.Datetime):
+                    base.this.append(trunc_expr)
+                    return base
```
knassre-bodo, line 202: This edge case allows us to flatten `DATETIME(DATETIME('now', '-1 day'), 'start of year')` into  `DATETIME('now', '-1 day', 'start of year')`.

pydough/sqlglot/transform_bindings.py
```diff
@@ -79,6 +176,182 @@ def impl(
...
+        `base`.
+    """
+    if dialect == DatabaseDialect.SQLITE:
+        # For sqlite, use the DATETIME operator to add the interval
+        offset_expr: SQLGlotExpression = sqlglot_expressions.convert(
+            f"{amt} {unit.value}"
+        )
+        if isinstance(base, sqlglot_expressions.Datetime):
+            base.this.append(offset_expr)
+            return base
```
knassre-bodo, line 245: This edge case allows us to flatten `DATETIME(DATETIME('now', 'start of year'), '-1 day')` into  `DATETIME('now', 'start of year', '-1 day')`.

knassre-bodo:
**Important:** need to add support for `"start of week"` to `DATETIME` (can also do `±<amt> weeks` by doing days * 7), but don't need to allow `DATEDIFF` week support. This is needed for defog.ai parity.

pydough/sqlglot/transform_bindings.py
```diff
@@ -79,6 +176,182 @@ def impl(
...
+
+    Args:
+        `base`: The base date/time expression to truncate.
+        `unit`: The unit to truncate the date/time expression to.
+        `dialect`: The dialect being used to generate the SQL.
+
+    Returns:
+        The SQLGlot expression to truncate `base`.
+    """
+    if dialect == DatabaseDialect.SQLITE:
```
vineetg3, line 193: For consistency with other SQLite bindings, we should either create a new function here or refactor the other bindings to align with the current conditional approach.

knassre-bodo: The problem is that the core behavior is the same, but the _helper_ function changes.

pydough/sqlglot/transform_bindings.py
vineetg3: Clarity purposes: Does the visitor revisit after appending? Maybe we can keep a small comment for further programmers?

tests/simple_pydough_functions.py
```diff
@@ -126,6 +127,209 @@ def function_sampler():
...
+        ),
+        DATETIME(
+            " Current Date ",
+            " 136 HOURS",
+            " \n +\t104 \nM \n\r ",
+            "-104 months",
+            " \n \t312 \nD \n\r ",
+        ),
+        DATETIME(" Current_Date ", "+ 45 MM", "-135 s"),
+    )
```
vineetg3, line 330: Suggestion: Might be far-fetched but how about adding tests for camel-cases such as `DATETIME("cUrReNt_Timestamp")`.

tests/test_pydough_to_sql.py
```diff
@@ -53,13 +55,19 @@
...
         ),
+        pytest.param(
+            datetime_sampler,
+            "datetime_sampler",
+            id="datetime_sampler",
+        ),
     ],
 )
-def test_pydough_to_sql_tpch(
+def test_pydough_to_ansi_sql_tpch(
```
vineetg3, line 65: Since we are renaming tests to specific dialects, is it suggestable to change rest of the names of tests in `test_pipeline`? Specifically, `test_defog_e2e_with_custom_data`?

vineetg3: Also, since this test uses `empty_context_database` fixture, I think the test name is misleading.

tests/simple_pydough_functions.py
knassre-bodo: 🤷 might as well :)

tests/test_pydough_to_sql.py
knassre-bodo: Good catch

pydough/sqlglot/transform_bindings.py
knassre-bodo: The visitor doesn't visit SQLGlot nodes, it returns them. In this case, `base` is a SQLGlot expression node returned earlier by the SQLGlot relational expression visitor (which created `base` from a relational expression). All we're doing is modifying `base` after the fact, instead of returning a new object containing it.

pydough/sqlglot/transform_bindings.py
knassre-bodo: In other words, I agree with your sentiment, but I want fixing this to be part of a larger overhaul/refactor of the file.

tests/simple_pydough_functions.py
knassre-bodo: Just added a few weird ones.

tests/test_pydough_to_sql.py
knassre-bodo: Reverted the ones in this file to just be `to_sql` instead of `to_ansi_sql` (but kept the ones in `test_relational_to_sql` because those are _actually_ just ANSI).

EDIT nvm those ones in `test_relational_to_sql` should explicitly say `sqlite`.

PR #268, opened 2025-02-15
Implemented remaining Python builtins

### LINKED ISSUE #259: Ensure Python remaining Python builtins with magic methods behave as-expected in PyDough
Goal: a user should be able to use the builtin Python functions like `abs` interchangeably with the PyDough builtin `ABS` function, and for other builtins that aren't supported in PyDough we should raise an exception.

To accomplish support for `abs`, add an implementation of the `__abs__` magic method in the UnqualifiedNode base class which calls the `ABS` PyDough function. See how `__invert__` does this for `~` becoming a call to the `NOT` operator.

To do proper error-handling for `math.floor`, add an implementation of `__floor__` which raises an appropriate exception explaining that `math.floor` is not supported (can later change this if we do add a floor operator).  See how `__bool__` does this in order to prevent things like `x and y`.

A full list of the magic methods to contend with, and the desired behavior:

- Supported:
  - `abs` (`__abs__`): call PyDough `ABS` function
  - `round` (`__round__`): call PyDough `ROUND` function
- Unsupported, until function is added:
  - `math.floor` (`__floor__`): raise exception with explanatory message (open followup to change this once we have a FLOOR function)
  - `math.ceil` (`__ceil__`): raise exception with explanatory message (open followup to change this once we have a CEIL function)
  - `math.trunc` (`__trunc__`): raise exception with explanatory message (open followup to change this once we have a TRUNC function)
  - `reversed` (`__reversed__`): raise exception with explanatory message (open followup to change this once we have a REVERSED function)
- Unsupported, always:
  - `int` (`__int__`): raise exception with explanatory message (open followup issue to create a casting function, then change the error message to refer to it) -> can't ever use `__int__` since the implementation has to return an integer instead of a PyDough object.
  - `float` (`__float__`): raise exception with explanatory message (same followup as `__int__`) -> can't ever use `__float__` since the implementation has to return a float instead of a PyDough object.
  - `complex` (`__complex__`): raise exception with explanatory message (same followup as `__complex__`) -> can't ever use `__complex__` since the implementation has to return a complex instead of a PyDough object.
  - `"hello"[:x]` (`__index__`): raise exception with explanatory message -> can't ever use `__index__` since the implementation has to return an integer instead of a PyDough object; means that someone tried to slice a Python string using a PyDough variable in the slice terms.
  - `"...".format(x)` (`__format__`): raise exception with explanatory message -> can't ever use `__format__` since the implementation has to return a string instead of a PyDough object.
  - `bool` (`__nonzero__`): raise exception with explanatory message -> can't ever use `__nonzero__` since the implementation has to return an integer instead of a PyDough object.
  - `len` (`__len__`): raise exception with explanatory message -> can't ever use `__len__` since the implementation has to return an integer instead of a PyDough object.
  - `x in y` (`__contains__`): raise exception with explanatory message -> can't ever use `__contains__` since the implementation has to return a boolean instead of a PyDough object.
  - `x[0] = 1` (`__setitem__`): raise exception with explanatory message -> operation not allowed
  - `list(x)`, or `for i in x`, or `[elem for elem in x]`, etc. (`__iter__`): raise exception with explanatory message ->  can't ever use `__iter__` since the implementation has to return an iterator instead of a PyDough object.

[**This link**](https://rszalski.github.io/magicmethods/) is quite useful at explaining such magic methods.

Note: all of this behavior should also be noted in the function list documentation. Perhaps there should be a special section on behaviors that are **NOT** allowed (can include some stuff already mentioned in the docs like `and`, `in`, `not,` a <= b <= c`)

vineetg3:
Resolves #259 

Please check issue for more details.

Few differences from the issues details:

`__iter__` : The `__iter__` method is not implemented for `UnqualifiedNode` because doing so would make the node iterable. This behavior could conflict with or disrupt the logic in other components, such as `Top_K`, which rely on specific handling of non-iterable structures. Implementing `__iter__` would introduce unintended behaviors and break the integrity of the system.

The `__format__`: The `__format__` method is not implemented which intentionally does not raise an error because its use interferes with the logic for printing and logging. Additionally, the fallback to `__str__` and `__repr__` provides sufficiently clear and concise representations, making a custom `__format__` unnecessary.


documentation/functions.md
```diff
@@ -580,3 +596,164 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+
+<!-- TOC --><a name="floor"></a>
+### FLOOR
+
+The `math.floor` function calls the `__floor__` magic method, which is not supported in PyDough.
+
+```py
+def bad_floor_1():
+   # Using `math.floor` (calls __floor__)
+   return Customer(age=math.floor(order.total_price))
```
knassre-bodo, line 611: Refactor all of these examples to match the format of other examples in the file (just a single line of code, not a function with a return statement)

documentation/functions.md
```diff
@@ -54,6 +54,18 @@ Below is the list of every function/operator currently supported in PyDough as a
 - [Window Functions](#window-functions)
    * [RANKING](#ranking)
    * [PERCENTILE](#percentile)
+- [Unsupported Magic Methods](#unsupported-magic-methods)
+   * [FLOOR](#floor)
```
knassre-bodo, line 58: Change all of these to the same format as the magic methods: `FLOOR` -> `__floor__`

documentation/functions.md
```diff
@@ -54,6 +54,18 @@ Below is the list of every function/operator currently supported in PyDough as a
...
+   * [CEIL](#ceil)
+   * [TRUNC](#trunc)
+   * [REVERSED](#reversed)
+   * [INT](#int)
+   * [FLOAT](#float)
+   * [COMPLEX](#complex)
+   * [INDEX](#index)
+   * [LEN](#len)
+   * [CONTAINS](#contains)
+   * [SETITEM](#setitem)
```
knassre-bodo, line 68: Include any other unsupported ones (`__call__` is only for function calls, `__bool__` is banned for `x and y`)

documentation/functions.md
```diff
@@ -580,3 +596,164 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
 # For every region, find the top 5% of customers with the highest account balances.
 Regions.nations.customers.WHERE(PERCENTILE(by=acctbal.ASC(), levels=2) > 95)
 ```
+## Unsupported Magic Methods
+
+Below is a list of magic methods that are not supported in PyDough. Calling these methods will result in an Exception.
```
knassre-bodo, line 601: Magic methods is not a terminology that our users will likely understand, since we are not targeting advanced Python developers. We should simplify this to just be about banned Python logic.

documentation/functions.md
```diff
@@ -580,3 +596,164 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+
+<!-- TOC --><a name="floor"></a>
+### FLOOR
+
+The `math.floor` function calls the `__floor__` magic method, which is not supported in PyDough.
```
knassre-bodo, line 606: Not CURRENTLY supported

documentation/functions.md
```diff
@@ -580,3 +596,164 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+```py
+def bad_floor_1():
+   # Using `math.floor` (calls __floor__)
+   return Customer(age=math.floor(order.total_price))
+```
+
+<!-- TOC --><a name="ceil"></a>
+### CEIL
+
+The `math.ceil` function calls the `__ceil__` magic method, which is not supported in PyDough.
```
knassre-bodo, line 617: Not CURRENTLY supported

documentation/functions.md
```diff
@@ -580,3 +596,164 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+```py
+def bad_trunc_1():
+   # Using `math.trunc` (calls __trunc__)
+   return Customer(age=math.trunc(order.total_price))
+```
+
+<!-- TOC --><a name="reversed"></a>
+### REVERSED
+
+The `reversed` function calls the `__reversed__` magic method, which is not supported in PyDough.
```
knassre-bodo, line 639: Not CURRENTLY supported

documentation/functions.md
```diff
@@ -580,3 +596,164 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+```py
+def bad_ceil_1():
+   # Using `math.ceil` (calls __ceil__)
+   return Customer(age=math.ceil(order.total_price))
+```
+
+<!-- TOC --><a name="trunc"></a>
+### TRUNC
+
+The `math.trunc` function calls the `__trunc__` magic method, which is not supported in PyDough.
```
knassre-bodo, line 628: Not CURRENTLY supported

documentation/functions.md
```diff
@@ -580,3 +596,164 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+
+<!-- TOC --><a name="nonzero"></a>
+### NONZERO
+
+Using an object in a boolean context calls the `__nonzero__` magic method, which is not supported in PyDough. This operation is not allowed because the implementation has to return an integer instead of a PyDough object.
+
+```py
+def bad_nonzero_1():
+   # Using in a boolean context (calls __nonzero__)
+   return Orders(discount=not order.total_price)
```
knassre-bodo, line 699: `bool(order.total_price)` would be a better example

documentation/functions.md
```diff
@@ -580,3 +596,164 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+```
+
+<!-- TOC --><a name="len"></a>
+### LEN
+
+The `len` function calls the `__len__` magic method, which is not supported in PyDough. This operation is not allowed because the implementation has to return an integer instead of a PyDough object.
```
knassre-bodo, line 705: Suggest using `LENGTH` instead.

documentation/functions.md
```diff
@@ -580,3 +596,164 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+```py
+def bad_len_1():
+   # Using `len` (calls __len__)
+   return Customers(len(customer.name))
+```
+
+<!-- TOC --><a name="contains"></a>
+### CONTAINS
+
+Using the `in` operator calls the `__contains__` magic method, which is not supported in PyDough. This operation is not allowed because the implementation has to return a boolean instead of a PyDough object.
```
knassre-bodo, line 716: Suggest using `ISIN` instead

pydough/unqualified/unqualified_node.py
```diff
@@ -124,6 +124,10 @@ def __getitem__(self, key):
             args: MutableSequence[UnqualifiedNode] = [self]
             for arg in (key.start, key.stop, key.step):
                 coerced_elem = UnqualifiedNode.coerce_to_unqualified(arg)
+                if not isinstance(coerced_elem, UnqualifiedLiteral):
+                    raise PyDoughUnqualifiedException(
+                        "PyDough objects cannot be used as indices in Python slices."
```
knassre-bodo, line 129: NIT: "are not currently supported"

pydough/unqualified/unqualified_node.py
```diff
@@ -260,6 +264,69 @@ def __call__(self, *args, **kwargs: dict[str, object]):
...
         return UnqualifiedCalc(self, calc_args)
 
+    def __abs__(self):
+        return UnqualifiedOperation("ABS", [self])
+
+    def __round__(self, n):
+        if n is None:
+            raise PyDoughUnqualifiedException(
+                "PyDough requires a specific number of decimal places for rounding."
+            )
```
knassre-bodo, line 274: Can just change this so the default is 0, like the normal Python behavior.

tests/bad_pydough_functions.py
```diff
@@ -73,3 +75,71 @@ def bad_slice_3():
 def bad_slice_4():
     # Unsupported slicing: reversed
     return Customers(name[::-1])
+
+
+def bad_floor_1():
```
knassre-bodo, line 80: NIT: don't use `_1` if it is the only one of its kind.

tests/simple_pydough_functions.py
```diff
@@ -391,6 +391,10 @@ def annotated_assignment():
     return Nations.WHERE(region.name == chosen_region)
 
 
+def abs_round_magic_method():
+    return DailyPrices(abs_low=abs(low), round_low=round(low, 2))
```
knassre-bodo, line 395: Also test `round(low)`

knassre-bodo (commented):
Looks great Vineet! Just left a bunch of clarity/nitpick comments to address, then I'll approve.

Also, can you add a new comment to the issue #259 explaining your deviations described in the top of this PR? These are good details to document in the original issue.

pydough/unqualified/unqualified_node.py
vineetg3: Are we going to support something like this: Customers[:orders] ?

documentation/functions.md
vineetg3: Will we be supporting both .CALCULATE() and conventional syntax of just ()? In that case I'll mention that.

pydough/unqualified/unqualified_node.py
knassre-bodo: At the moment, no. Eventually, 🤷?

documentation/functions.md
```diff
@@ -580,3 +657,178 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+```
+
+<!-- TOC --><a name="__nonzero__"></a>
+
+### \_\_nonzero\_\_
+
+Using an object in a boolean context calls the `__nonzero__` magic method, which is not supported in PyDough. This operation is not allowed because the implementation has to return an integer instead of a PyDough object.
+
+```py
+Orders(discount=bool(order.total_price))
```
knassre-bodo, line 785: NIT: lets change this to something where `bool` makes sense:
```py
Lineitems(is_taxed=bool(tax))
```

documentation/functions.md
```diff
@@ -580,3 +657,178 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+   print("Available")
+   
+# Use & instead of and
+Customers.WHERE((acctbal > 0) & (nation.name == "GERMANY"))
+
+# Use | instead of or 
+Orders.WHERE((discount > 0.05) | (tax > 0.08))
+
+# Use ~ instead of not
+Parts.WHERE(~(retail_price > 1000))
```
knassre-bodo, line 683: Have all of these do the bad thing (`and`, `or`, and `not`, then suggest alternative in the comments)

documentation/functions.md
```diff
@@ -580,3 +657,178 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+Below is a list of banned python logic (magic methods,etc.) that are not supported in PyDough. Calling these methods will result in an Exception.
+
+<!-- TOC --><a name="__bool__"></a>
+
+### \_\_bool\_\_
+
+The `__bool__` magic method is not supported in PyDough. PyDough code cannot be treated as booleans. Instead of using `or`, `and`, and `not`, use `|`, `&`, and `~` for logical operations. Check out the [Logical](#logical) section for more information.
+
+```py
+# Not allowed - will raise PyDoughUnqualifiedException
```
knassre-bodo, line 672: Just say this isn't allowed

documentation/functions.md
```diff
@@ -580,3 +657,178 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+
+### \_\_call\_\_
+
+The `__call__` magic method is not supported in PyDough, but calling an object will not throw an error. Instead, it internally evaluates to the `CALC` property.
+
+```py
+# These are equivalent:
+Customers(name)
+Customers.__call__(name)
+```
```
knassre-bodo, line 696: Let's make the examples more consistent & stick with one pattern:
- Show a bad behavior with regular Python code (not the magi method)
- Include a comment saying it is banned, and why

E.g. for this one it would be:

```py
# Not allowed - calls PyDough code as if it were a function
(1 - discount)(extended_price * 0.5)
```

documentation/functions.md
```diff
@@ -580,3 +657,178 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+
+<!-- TOC --><a name="__reversed__"></a>
+
+### \_\_reversed\_\_
+
+The `reversed` function calls the `__reversed__` magic method, which is currently not supported in PyDough.
+
+```py
+Orders(reversed(order_key))
```
knassre-bodo, line 735: Lets change this to a string so it makes sense, e.g. `Regions(backwards_name=reversed(name))`

documentation/functions.md
```diff
@@ -580,3 +657,178 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+```
+
+<!-- TOC --><a name="__index__"></a>
+
+### \_\_index\_\_
+
+Using an object as an index calls the `__index__` magic method, which is not supported in PyDough. This operation is not allowed because the implementation has to return an integer instead of a PyDough object.
+
+```py
+Customers(sliced = name[:order])
```
knassre-bodo, line 775: A better example would be `Orders(s="ABCDE"[:order_priority])`

tests/bad_pydough_functions.py
```diff
@@ -73,3 +75,71 @@ def bad_slice_3():
...
+    # Using `math.ceil` (calls __ceil__)
+    return Customer(age=math.ceil(order.total_price))
+
+
+def bad_trunc():
+    # Using `math.trunc` (calls __trunc__)
+    return Customer(age=math.trunc(order.total_price))
+
+
+def bad_reversed():
```
knassre-bodo, line 95: Anything I suggested revising in the docs, lets also revise here

documentation/functions.md
```diff
@@ -394,38 +441,56 @@ Lineitems.WHERE(MONOTONIC(10, quantity, 20) & MONOTONIC(5, part.size, 13))
...
+# The below statement is equivalent to above.
+Customers(acct_magnitude = abs(acctbal))
 ```
 
 <!-- TOC --><a name="round"></a>
+
 ### ROUND
 
-The `ROUND` function rounds its first argument to the precision of its second argument. The rounding rules used depend on the database's round function.
+The `ROUND` function rounds its first argument to the precision of its second argument. The rounding rules used depend on the database's round function. The `round()` magic method is also evaluated to the same.
```
knassre-bodo, line 465: Change to "The Python builtin `round()` function can also be used to accomplish the same thing"

documentation/functions.md
```diff
@@ -394,38 +441,56 @@ Lineitems.WHERE(MONOTONIC(10, quantity, 20) & MONOTONIC(5, part.size, 13))
...
 ## Numerical Functions
 
 Below is each numerical function currently supported in PyDough.
 
 <!-- TOC --><a name="abs"></a>
+
 ### ABS
 
-The `ABS` function returns the absolute value of its input.
+The `ABS` function returns the absolute value of its input. The `abs()` magic method is also evaluated to the same.
```
knassre-bodo, line 453: Change to "The Python builtin `abs()` function can also be used to accomplish the same thing."

documentation/functions.md
```diff
@@ -394,38 +441,56 @@ Lineitems.WHERE(MONOTONIC(10, quantity, 20) & MONOTONIC(5, part.size, 13))
...
+Parts(rounded_price = round(retail_price, 1))
+```
+
+Note: The default precision for `round` magic method is 0, to be in alignment with the Python implementation. The pydough `ROUND` function requires the precision to be specified.
+
+```py
+# This is legal.
+Parts(rounded_price = round(retail_price))
+# This is illegal as precision is not specified.
+Parts(rounded_price = ROUND(retail_price))
```
knassre-bodo, line 479: Make sure to open a Github issue to fix this behavior.

documentation/functions.md
```diff
@@ -394,38 +441,56 @@ Lineitems.WHERE(MONOTONIC(10, quantity, 20) & MONOTONIC(5, part.size, 13))
...
-The `ROUND` function rounds its first argument to the precision of its second argument. The rounding rules used depend on the database's round function.
+The `ROUND` function rounds its first argument to the precision of its second argument. The rounding rules used depend on the database's round function. The `round()` magic method is also evaluated to the same. 
 
 ```py
 Parts(rounded_price = ROUND(retail_price, 1))
+# The below statement is equivalent to above.
+Parts(rounded_price = round(retail_price, 1))
+```
+
+Note: The default precision for `round` magic method is 0, to be in alignment with the Python implementation. The pydough `ROUND` function requires the precision to be specified.
```
knassre-bodo, line 473: NIT: the PyDough `ROUND` function..

Also, refer to `round` as a function instead of a magic method.

documentation/functions.md
```diff
@@ -580,3 +657,178 @@ Customers.WHERE(PERCENTILE(by=acctbal.ASC(), n_buckets=1000) == 1000)
...
+
+```py
+Customers(len(customer.name))
+```
+
+<!-- TOC --><a name="__contains__"></a>
+
+### \_\_contains\_\_
+
+Using the `in` operator calls the `__contains__` magic method, which is not supported in PyDough. This operation is not allowed because the implementation has to return a boolean instead of a PyDough object. Instead, usage of [ISIN](#isin) function is recommended.
```
knassre-bodo, line 802: NIT on the last part, change to several recommendations phrased as follows:
"If you need to check if a string is inside another substring, use `CONTAINS`. If you need to check if an expression is a member of a list of literals, use `ISIN`."

knassre-bodo (approved):
Left a few more revision suggestions including some nitpicks, but feel free to merge once you feel they are addressed.

documentation/functions.md
vineetg3: Done, #276 tackles this.

PR #272, opened 2025-02-18
Fix bug with accessing multiple chained PARTITION children

knassre-bodo:
**Goal**: fix the bug causing PyDough such as the following to fail:
```py
data = Tickers(symbol).TOP_K(5, by=symbol.ASC())
grps_a = PARTITION(data, name="child_3", by=(currency, exchange, ticker_type))
grps_b = PARTITION(grps_a, name="child_2", by=(currency, exchange))
grps_c = PARTITION(grps_b, name="child_1", by=exchange)
result = grps_c.child_1.child_2.child_3
```

The error is an `IndexError: list index out of range` in the following section of hybrid conversion:
```py
case PartitionChild():
  hybrid = self.make_hybrid_tree(node.ancestor_context, parent)
  successor_hybrid = HybridTree(
    HybridPartitionChild(hybrid.children[0].subtree)
  )
  ...
```

Reason: if `node.ancestor_context` is **also**  a partition child, the hybrid child subtree handling/invocation is not being done correctly. The correct way is to recursively step into `hybrid.pipeline[0].subtree` until `hybrid.pipeline[0]` is no longer a partition child. Once this happens, then we can access `children[0].subtree`, because then we have found original data being partitioned.


pydough/conversion/hybrid_tree.py
```diff
@@ -1880,9 +1880,14 @@ def make_hybrid_tree(
                 return successor_hybrid
             case PartitionChild():
                 hybrid = self.make_hybrid_tree(node.ancestor_context, parent)
-                successor_hybrid = HybridTree(
-                    HybridPartitionChild(hybrid.children[0].subtree)
-                )
```
knassre-bodo, line 1885: the problem is that if `node.ancestor_context` starts  with another `PartitionChild`, then `hybrid` won't have any children because `hybrid.pipeline[0]` will be  another `HybridPartitionChild` containing the desired subtree in its `.subtree` property. Thus, we would replace `hybrid.children[0].subtree` with `hybrid.pipeline[0].subtree.children[0].subtree`. And if the grandparent is **also** a partition child, it keeps going:  `hybrid.pipeline[0].subtree.pipeline[0].subtree.children[0].subtree`

vineetg3 (approved):
LGTM!

PR #277, opened 2025-02-21
Improved Slicing to support negative indices

### LINKED ISSUE #260: Add support for more slicing cases with SUBSTR
Currently, PyDough only allows string slicing in cases where the step is absent (or 1), and the start/stop are both non-negative (or absent), by translating to `SUBSTR`. However, `SUBSTR` has support for negatives, which we should also support. This means we can expand the conversion logic in `convert_slice` to handle the following additional cases:
1. `s[-n:]` -> `SUBSTR(s, -n)`
2. `s[-n:-m]` -> `SUBSTR(s, -n, n-m))`
3. `s[:-n]` -> `SUBSTR(s, 1, LENGTH(s) -n)`

If the dialect does not allow negative indices (like default ANSI), can change it so `LENGTH(s) - n` is always used for the second argument in the first 2 cases. This is viable because currently we require the slicing indices to be integer literals.

These expansions should all be tested, and should be noted in the function list documentation (which currently specifies the restricted semantics). The only restriction on slicing with `s[...]` syntax should now be that:
- The expression being sliced (`s`) must be a PyDough term (rather than a string literal)
- The step (e.g. `c` in `s[a:b:c]` must be 1 or must be absent
- The start/stop/step must all be litearls

vineetg3:
Resolves #260.
Please read the issue for more details.

The `SUBSTR` conversion logic in `convert_slice` is being expanded to handle negative indices.
Here is the outline of the logic to support Python based slicing semantics:
- Case 1: `(None, None)`
    - Returns the string as is.
- Case 2: `(start, None)`  
    - Positive `start`: Convert to 1-based indexing and slice from `start`.  
    - Negative `start`: Compute `LENGTH(string) + start + 1`; clamp to `1` if less than `1`.
- Case 3: `(None, stop)`  
    - Positive `stop`: Slice from position `1` to `stop`.  
    - Negative `stop`: Compute `LENGTH(string) + stop`; clamp to `0` if less than `0` (empty slice).
- Case 4: `(start, stop)`  
    - 1. Both `start` & `stop` >= 0:  
        - Convert `start` to 1-based.  
        - Set `length = stop - start`.  
    - 2. `start < 0`, `stop >= 0`:  
        - Convert `start` to positive. If < 1, set to 1.  
        - Compute `length = stop - start` (clamp to 0 if negative).  
    - 3. `start >= 0`, `stop < 0`:   
        - Convert `stop` & `start` to positive.  
        - If `stop` < 1, slice is empty (`length = 0`).  
        - Else, `length = stop - start`.  
    - 4. `start < 0`, `stop < 0`:  
        - Convert `start` & `stop` to positive. If `start` < 1, set to 1.  
        - If `stop` < 1, slice is empty (`length = 0`).  
        - Else, `length = stop - start`.


pydough/sqlglot/transform_bindings.py
```diff
@@ -504,12 +504,68 @@ def convert_concat(
...
+        - Negative `stop`: Compute `LENGTH(string) + stop`; clamp to `0` if less than `0` (empty slice).
+    - Case 4: `(start, stop)`
+        - 1. Both `start` & `stop` >= 0:
+            - Convert `start` to 1-based.
+            - Set `length = stop - start`.
+        - 2. `start < 0`, `stop >= 0`:
+            - Convert `start` to positive. If < 1, set to 1.
+            - Compute `length = stop - start` (clamp to 0 if negative).
+        - 3. `start >= 0`, `stop < 0`:
+            - Convert `stop` & `start` to positive.
```
hadia206, line 562: Based on 3. case `start` is already positive.
```suggestion
            - Convert `stop` to positive.
```

pydough/pydough_operators/expression_operators/README.md
```diff
@@ -59,7 +59,7 @@ These are created with a prefix operator syntax instead of called as a function.
 
 These are other PyDough operators that are not necessarily used as functions:
 
-- `SLICE`: operator used for string slicing, with the same semantics as Python string slicing. If `s[a:b:c]` is done, that is translated to `SLICE(s,a,b,c)` in PyDough, and any of `a`/`b`/`c` could be absent. Currently, PyDough does not support negative indices or providing step values other than 1.
+- `SLICE`: operator used for string slicing, with the same semantics as Python string slicing. If `s[a:b:c]` is done, that is translated to `SLICE(s,a,b,c)` in PyDough, and any of `a`/`b`/`c` could be absent. Negative slicing is supported, however currently, PyDough does not support providing step values other than 1.
```
hadia206, line 62: My understanding is this `PyDough does not support providing step values other than 1.` applies for both positive and negative values.
Current edit makes it that only negative doesn't support step values other than 1.
A suggestion.
```suggestion
- `SLICE`: operator used for string slicing, with the same semantics as Python string slicing. If `s[a:b:c]` is done, that is translated to `SLICE(s,a,b,c)` in PyDough, and any of `a`/`b`/`c` could be absent. Negative slicing is supported. Currently PyDough does not support providing step values other than 1.
```

pydough/sqlglot/transform_bindings.py
```diff
@@ -518,60 +574,259 @@ def convert_slice(
...
-    if isinstance(stop, sqlglot_expressions.Literal):
-        if int(stop.this) < 0:
-            raise NotImplementedError(
-                "SLICE function currently only supports non-negative stop indices"
-            )
-        stop_idx = int(stop.this)
     if isinstance(step, sqlglot_expressions.Literal):
-        if int(step.this) != 1:
+        try:
+            step_int = int(step.this)
```
hadia206, line 591: Shouldn't this be like others to handle None case?
`start_idx = int(start.this) if not isinstance(start, sqlglot_expressions.Null) else None`

pydough/sqlglot/transform_bindings.py
```diff
@@ -518,60 +574,259 @@ def convert_slice(
...
+        except ValueError:
+            raise ValueError(
```
hadia206, line 593: It's better to also add which argument was the issue.
e.g. SLICE function step argument must be integer literal or absent

pydough/sqlglot/transform_bindings.py
```diff
@@ -518,60 +574,259 @@ def convert_slice(
...
 
+    start_idx: int | None = None
+    stop_idx: int | None = None
+    try:
+        start_idx = (
+            int(start.this) if not isinstance(start, sqlglot_expressions.Null) else None
+        )
+    except ValueError:
+        raise ValueError(
+            "SLICE function currently only supports the slicing arguments being integer literals or absent"
```
hadia206, line 609: Same. Specify argument

pydough/sqlglot/transform_bindings.py
```diff
@@ -518,60 +574,259 @@ def convert_slice(
...
+        )
+    try:
+        stop_idx = (
+            int(stop.this) if not isinstance(stop, sqlglot_expressions.Null) else None
+        )
+    except ValueError:
+        raise ValueError(
+            "SLICE function currently only supports the slicing arguments being integer literals or absent"
```
hadia206, line 617: Same. Specify argument

pydough/sqlglot/transform_bindings.py
```diff
@@ -518,60 +574,259 @@ def convert_slice(
...
+            )
+        if step_int != 1:
             raise NotImplementedError(
                 "SLICE function currently only supports a step of 1"
             )
 
+    start_idx: int | None = None
+    stop_idx: int | None = None
+    try:
+        start_idx = (
```
hadia206, line 604: Why do we have value in `()`?

documentation/functions.md
```diff
@@ -157,12 +157,15 @@ Below are all other operators currently supported in PyDough that use other synt
...
+A string expression can have a substring extracted with Python string slicing syntax `s[a:b:c]`.
+The implementation is based on Python slicing semantics. PyDough supports negative slicing, but currently, it does not support providing step values other than 1.
 
 ```py
 Customers(
     country_code = phone[:3],
-    name_without_first_char = name[1:]
+    name_without_first_char = name[1:],
+    last_digit = phone[-1:],
+    name_without_start_and_end_char = name[1:-1]
```
knassre-bodo, line 168: Lets also add:
- `phone_without_last_5_chars = phone[:-5]`
- `name_second_to_last_char = name[-2:-1]`

hadia206 (commented):
Looks good to me.
I have minor comments.

pydough/sqlglot/transform_bindings.py
```diff
@@ -504,12 +504,68 @@ def convert_concat(
         return Concat(expressions=inputs)
 
 
+def positive_index(
+    column: SQLGlotExpression, neg_index: int, is_0_based: bool = False
```
knassre-bodo, line 508: NIT: `is_zero_indexed` instead if `is_0_based`

pydough/sqlglot/transform_bindings.py
```diff
@@ -504,12 +504,68 @@ def convert_concat(
...
+) -> SQLGlotExpression:
+    """
+    Gives the SQL Glot expression for converting a
+    negative index to a positive index in 1 or 0 based indexing
+    based on the length of the column.
+
+    Args:
+        `column`: The column expression to reference
```
knassre-bodo, line 516: NIT: this isn't necessarily a column (could be a function call, or a literal). Let's just call this `string_expr` and say it refers to the string that the negative index is referencing a point within.

pydough/sqlglot/transform_bindings.py
```diff
@@ -518,60 +574,259 @@ def convert_slice(
         converted to SQLGlot expressions.
 
     Returns:
-        The SQLGlot expression matching the functionality of `SUBSTRING`.
+        The SQLGlot expression matching the functionality of Python based string slicing
+        with the caveat that it only supports a step of 1.
     """
     assert len(sql_glot_args) == 4
+    _, start, stop, step = sql_glot_args
```
knassre-bodo, line 581: NIT: can unpack `string_expr` as the first arg & use that instead of `sql_glot_args[0]` in the rest of the function.

pydough/sqlglot/transform_bindings.py
```diff
@@ -518,60 +574,259 @@ def convert_slice(
...
 
     for arg in sql_glot_args[1:]:
         if not isinstance(arg, (sqlglot_expressions.Literal, sqlglot_expressions.Null)):
```
knassre-bodo, line 584: NIT: can replace this the loop guard with `for arg in (start, stop, step)`

pydough/sqlglot/transform_bindings.py
knassre-bodo, line 591: NIT: can push this check into the  the earlier for loop by checking if `isinstance(arg.this, int)` when it is a literal.

pydough/sqlglot/transform_bindings.py
knassre-bodo, line 604: Same here: no need to check with try/excepts, just psuh the check into the earlier loop & just do `start_idx = int(start.this)` if `start` isn't null.

pydough/sqlglot/transform_bindings.py
```diff
@@ -518,60 +574,259 @@ def convert_slice(
...
+                                ],
+                            ),
+                        ],
+                    ),
+                )
+                return answer
+            if start_idx < 0 and stop_idx < 0:
+                # Early return if start index is greater than stop index
+                # e.g., "hello"[-2:-4] should return empty string
+                if start_idx > stop_idx:
```
knassre-bodo, line 789: Should be `>=` since `"hello"[-2:-2]` returns an empty string (the upper bound is exclusive.

tests/bad_pydough_functions.py
```diff
@@ -58,23 +59,58 @@ def bad_window_7():
...
-    return Customers(name[1:10:2])
+    # Unsupported slicing: non-integer start
+    return Customers(name["invalid":-1:])
 
 
 def bad_slice_4():
-    # Unsupported slicing: reversed
-    return Customers(name[::-1])
+    # Unsupported slicing: non-integer start
+    return Customers(name[datetime.datetime.now() : -1 :])
```
knassre-bodo, line 78: What about testing with non-literals?

tests/simple_pydough_functions.py
```diff
@@ -686,3 +686,52 @@ def datediff():
         minutes_diff=DATEDIFF("minutes", date_time, y_datetime),
         seconds_diff=DATEDIFF("seconds", date_time, y_datetime),
     ).TOP_K(30, by=years_diff.ASC())
+
+
+def step_slicing():
+    return Customers.WHERE(name == "Jane Smith")(
```
knassre-bodo, line 692: Since the test generates the answers, why not just test on the entire table (its only 20 rows)?

knassre-bodo (commented):
Looks fantastic @vineetg3!!! Just a few comments before merging.

knassre-bodo (approved):
LGTM!

PR #281, opened 2025-02-28
Added config for default and propagation of collations

### LINKED ISSUE #149: Support ordered operations without providing ASC or DESC
The intended functionality:
- It should be legal to have a PyDough ordered expression without an ASC or DESC. If one is not present, one will be automatically provided.
- The behavior for how one is provided, along with the na position behavior, should be controlled by various PyDough configs. The config options in question:
  - Always ASCEND (this is the default value)
  - Always DESCEND
  - Use most recent, and ASCEND for first
  - Use most recent, and DESCEND for first
- Control the default na position behavior with another config:
  - ASC default is null first (DEFAULT) vs ASC default is null last
  - DESC default is null last (DEFAULT) vs ASC default is null first

IMPORTANT: all documentation that refers to this behavior as invalid (e.g. in the [DSL documentation](https://github.com/bodo-ai/PyDough/blob/main/documentation/dsl.md)), as well as any tests that check for this behavior as invalid, need to be updated.

vineetg3:
Resolves #149.

The following configs are added which is used in the behavior of `PERCENTILE`, `RANKING`, `ORDER_BY` and `TOP_K` where a collation(`ASC` or `DESC`) is not explicitly specified:

- `collation_default_asc` (default=True): if True, then the default collation is ascending. If False, then the default collation is descending.
- `propogate_collation` (default=False): if True, then the collation of the current expression, which does not have a collation, uses the most recent available collation in the nodes of the term. If there is no recent available collation, then the default collation is used as specified by `collation_default_asc`. If False, the expression uses the default collation as specified by `collation_default_asc`.


documentation/usage.md
```diff
@@ -167,6 +167,62 @@ old_sum_behavior = configs.sum_default_zero
...
+
+3. `collation_default_asc` (default=True): if True, then the default collation is ascending. If False, then the default collation is descending.
+4. `propogate_collation` (default=False): if True, then the collation of the current expression, which does not have a collation, uses the most recent available collation in the nodes of the term. If there is no recent available collation, then the default collation is used as specified by `collation_default_asc`. If False, the expression uses the default collation as specified by `collation_default_asc`.
+
+For example, consider the following PyDough code:
+
+```py
+%%pydough
+# The collations are not explicitly specified for few of the terms.
+result =Suppliers.ORDER_BY(
```
knassre-bodo, line 180: Make sure to include the space after the `=` for variable definitions (e.g. `result = Suppliers.ORDER_BY(`

documentation/usage.md
```diff
@@ -167,6 +167,62 @@ old_sum_behavior = configs.sum_default_zero
...
+    COUNT(supply_records).ASC(),
+    name, phone,
+    account_balance.DESC(), comment
```
knassre-bodo, line 183: Lets change up these examples slightly for visual clarity, and also for correctness (the two examples further below don't correctly match the ordering behavior for this example).

Lets make it this:
```py
Suppliers.ORDER_BY(
  COUNT(lines),
  nation.name.ASC(),
  COUNT(supply_records),
  account_balance.DESC(),
  key,
)
```

Then, three examples:
```py
# With the default settings
Suppliers.ORDER_BY(
  COUNT(lines).ASC(),
  nation.name.ASC(),
  COUNT(supply_records).ASC(),
  account_balance.DESC(),
  key.ASC(),
)

# With collation_default_asc=False and propogate_collation=False
Suppliers.ORDER_BY(
  COUNT(lines).DESC(),
  nation.name.ASC(),
  COUNT(supply_records).DESC(),
  account_balance.DESC(),
  key.DESC(),
)

# With collation_default_asc=True and propogate_collation=True
Suppliers.ORDER_BY(
  COUNT(lines).ASC(),
  nation.name.ASC(),
  COUNT(supply_records).ASC(),
  account_balance.DESC(),
  key.DESC(),
)
```

pydough/exploration/term.py
```diff
@@ -123,14 +125,14 @@ def explain_term(
     lines: list[str] = []
     root: UnqualifiedRoot | None = find_unqualified_root(node)
     qualified_node: PyDoughQDAG | None = None
-
+    config: PyDoughConfigs = pydough.active_session.config
```
knassre-bodo, line 128: NIT: lets change this API (& the other two) to **optionally** take in a config argument (the default is None, which means use `pydough.active_session.config`). That includes:
- `explain`
- `explain_term`
- `explain_structure`

This will make things easier for when we later add case insensitivity.

tests/test_qualification.py
```diff
@@ -528,10 +530,129 @@ def test_qualify_node_to_ast_string(
...
+            "address.ASC(na_pos='first'), nation_key.ASC(na_pos='first'), "
+            "phone.ASC(na_pos='first'), account_balance.DESC(na_pos='last'), "
+            "comment.DESC(na_pos='last')).TOP_K(5, key.ASC(na_pos='first'), "
+            "COUNT(supply_records).ASC(na_pos='first'), name.DESC(na_pos='last'), "
+            "address.DESC(na_pos='last'), nation_key.DESC(na_pos='last'), "
+            "phone.DESC(na_pos='last'), account_balance.ASC(na_pos='first'), "
+            "comment.ASC(na_pos='first'))",
+            True,
+            True,
+            id="collation1",
```
knassre-bodo, line 566: NIT: name these based on what they each represent:
- `asc-no_propagation`
- `desc-no_propagation`
- `asc-with_propagation`
- `desc-with_propagation`

tests/test_qualification.py
```diff
@@ -528,10 +530,129 @@ def test_qualify_node_to_ast_string(
...
+            "account_balance.ASC(na_pos='first'), comment.DESC(na_pos='last'))",
+            False,
+            False,
+            id="collation4",
+        ),
+    ],
+)
+def test_qualify_node_collation(
+    impl: Callable[[], UnqualifiedNode],
+    answer_tree_str: str,
```
knassre-bodo, line 638: The name `answer_tree_str` is misleading since these aren't tree strings (tree string is what happens if you do `qualified.to_tree_string()` instead of `repr(qualified)`. You may want to switch to tree strings since those are a bit easier to read.

pydough/unqualified/unqualified_node.py
```diff
@@ -451,17 +451,19 @@ def get_by_arg(
             f"The `by` argument to `{func_name}` must be provided"
         )
     by = kwargs.pop("by")
-    if isinstance(by, UnqualifiedCollation):
+    by_allowed_types = (UnqualifiedCollation, UnqualifiedOperation, UnqualifiedAccess)
+    if isinstance(by, by_allowed_types):
```
knassre-bodo, line 455: This should be changed since now we don't care what the `by` values are so long as they are unqualified nodes. The check should just be to make sure that `by` or `arg` are instances of `UnqualifiedNode`.

knassre-bodo (commented):
Didn't finish reviewing the docs but decided to leave my other comments first so you could address those without having to wait for the rest of the review.

documentation/usage.md
```diff
@@ -167,6 +167,52 @@ old_sum_behavior = configs.sum_default_zero
...
+
+# With collation_default_asc=True and propogate_collation=True
+Suppliers.ORDER_BY(
+  COUNT(lines).ASC(),
+  nation.name.ASC(),
+  COUNT(supply_records).ASC(),
+  account_balance.DESC(),
+  key.DESC(),
+)
+```
```
hadia206, line 215: nit, add fourth case as well
`collation_default_asc=False and propogate_collation=True`

documentation/dsl.md
```diff
@@ -784,6 +783,28 @@ Addresses.CALCULATE(state, city).WHERE(
...
+
+**Good Example #8**: Sort every person by their first name in ascending order, last name in descending order, and the number of packages they have ordered in descending order.
+Let's keep the default behavior of `collation_default_asc` and set `propogate_collation` to `True`.
+
+```py
+%%pydough
+People.ORDER_BY(first_name, last_name.DESC(), COUNT(packages))
+```
+
+This is valid because the collation term is by default ascending based on the `collation_default_asc` config. Setting the `propogate_collation` config to `True` will cause the collation to be propogated to the `COUNT(packages)` term. Hence its equivalent to:
```
hadia206, line 801: To avoid confusion be explicit that it will get it from last_name.
e.g. 
`will cause the collation to be propogated to the `COUNT(packages)` term from last used one (`last_name.DESC()`) .

documentation/usage.md
```diff
@@ -167,6 +167,52 @@ old_sum_behavior = configs.sum_default_zero
...
+# With the default settings
+Suppliers.ORDER_BY(
+  COUNT(lines).ASC(),
+  nation.name.ASC(),
+  COUNT(supply_records).ASC(),
+  account_balance.DESC(),
+  key.ASC(),
+)
+
+# With collation_default_asc=False and propogate_collation=False
```
hadia206, line 198: Suggestion: add code for setting the config with each example
```
configs.collation_default_asc = False
configs. propogate_collation = False
```

pydough/configs/pydough_configs.py
```diff
@@ -65,6 +65,19 @@ class PyDoughConfigs:
...
+    collation_default_asc = ConfigProperty[bool](True)
+    """
+    If True, then the collation will default to `ASC`. If False, then the
+    collation will default to `DESC`. The default is True.
+    """
+
+    propogate_collation = ConfigProperty[bool](False)
+    """
+    If True, then the collation(`ASC` or `DESC`), if exists, will be propogated across
+    the nodes for that term. If False, then the collation will not be propogated across
```
hadia206, line 77: Clarify how the value will be identified
e.g. It will use the default collation as specified by `collation_default_asc`

pydough/exploration/README.md
```diff
@@ -4,12 +4,14 @@ This subdirectory of the PyDough directory deals with the exploration of PyDough
 
 ## Available APIs
 
-The exploration module has the following notable APIs available for use:
+The exploration module has the following notable APIs available for use.:
```
hadia206, line 7:
```suggestion
The exploration module has the following notable APIs available for use:
```

pydough/unqualified/unqualified_node.py
```diff
@@ -451,17 +451,19 @@ def get_by_arg(
...
+        and all(isinstance(arg, by_allowed_type) for arg in by)
         and len(by) > 0
     ):
         raise PyDoughUnqualifiedException(
-            f"The `by` argument to `{func_name}` must be a single collation expression or a non-empty iterable of collation expressions"
+            f"The `by` argument to `{func_name}` must be a single expression or a non-empty iterable of expressions."
+            "Please refer to the config documentation for more information."
         )
-    return by
+    return list(by)
```
hadia206, line 466: why `list` here?
Isn't already set to `by = [by]` line 456?
 what cases will this be needed?

hadia206 (commented):
Overall looks good to me.
Left some questions for my understanding.
I'll let Kian give the approval

knassre-bodo (approved):
LGTM! Fantastic job, just a few minor notes.

pydough/unqualified/unqualified_node.py
vineetg3: Yes, that's why I've wrapped it with `list()`.

PR #286, opened 2025-03-10
Implemented default precision of 0 for ROUND

### LINKED ISSUE #276: Add Default Precision 0 to `ROUND` Function
**Description:**

Currently, the `ROUND` function in PyDough requires the user to explicitly provide a precision argument. However, Python's built-in `round()` function defaults to a precision of `0` when no precision is specified. To ensure consistency and make the `ROUND` function more user-friendly, its behavior should be updated to align with Python's `round()` by defaulting the precision to `0`.

**Proposed Solution:**

- Update the `ROUND` function implementation to make the precision argument optional.
- If no precision is provided, default to `0`.

Example:

Current Behavior:
```python
ROUND(expression, 0)  # Users must explicitly provide the precision, e.g., 0 for rounding to the nearest integer.
```

New Behavior:
```python
ROUND(expression)  # Defaults to ROUND(expression, 0)
```

- Ensure that this change is reflected in the documentation.

vineetg3:
Resolves #276.
Please see issue for more details.

Previously, the user had to provide the precision for ROUND function. Now, the user does not have to provide precision, in which case the ROUND defaults to precision of 0.


pydough/sqlglot/transform_bindings.py
```diff
@@ -1650,7 +1696,8 @@ def add_builtin_bindings(self) -> None:
 
         # Numeric functions
         self.bind_simple_function(pydop.ABS, sqlglot_expressions.Abs)
-        self.bind_simple_function(pydop.ROUND, sqlglot_expressions.Round)
+        self.bindings[pydop.ROUND] = convert_round
+        # self.bind_simple_function(pydop.ROUND, sqlglot_expressions.Round)
```
knassre-bodo, line 1700: Delete

tests/simple_pydough_functions.py
```diff
@@ -150,6 +150,13 @@ def regional_suppliers_percentile():
 
 def function_sampler():
     # Examples of using different functions
+    # Functions tested:
+    # JOIN_STRINGS,
+    # ROUND - with and without precision,
+    # KEEP_IF,
+    # PRESENT,
+    # ABSENT,
+    # MONOTONIC
```
knassre-bodo, line 159: NIT: use this format:
```py
# Functions tested:
#   JOIN_STRINGS
#   ROUND (with and without precision)
#   KEEP_IF
...
```

Also, don't forget about some string slicing

knassre-bodo (approved):
Just some minor nitpicks, but LGTM.

tests/simple_pydough_functions.py
vineetg3: Since string slicing is already tested in another test case, do you think it would be useful to include simple slicing here as well?

pydough/sqlglot/transform_bindings.py
```diff
@@ -1492,6 +1492,52 @@ def convert_sqlite_datediff(
...
+    if len(sql_glot_args) == 1:
+        precision_glot = sqlglot_expressions.Literal.number(0)
+    else:
+        # Check if the second argument is a integer literal.
+        if (
+            not isinstance(sql_glot_args[1], sqlglot_expressions.Literal)
+            or sql_glot_args[1].is_string
+        ):
+            raise ValueError(
+                f"Unsupported argument {sql_glot_args[1]} for ROUND."
```
hadia206, line 1524: nit: print received type not just its value.

pydough/sqlglot/transform_bindings.py
```diff
@@ -1492,6 +1492,52 @@ def convert_sqlite_datediff(
...
+                "The precision argument should be an integer literal."
+            )
+        try:
+            int(sql_glot_args[1].this)
+        except ValueError:
+            raise ValueError(
```
hadia206, line 1530: Why do we need try-except here? Isn't if-stmt above enough to guarantee it's an int.

hadia206 (approved):
Minor comments but overall looks good to me.

PR #289, opened 2025-03-11
Optimize PyDough generated SQL to avoid extra joins from PARTITION when possible

knassre-bodo:
Consider this PyDough logic:
```py
priorities = PARTITION(
   Orders.WHERE(YEAR(order_date) == 1992), name="orders", by=order_priority
)
result = (
   priorities.orders.WHERE(RANKING(by=total_price.DESC(), levels=1) == 1)
   .CALCULATE(order_priority, order_key=key, order_total_price=total_price)
   .ORDER_BY(order_priority.ASC())
)
```

The goal is to ensure that the final SQL does **NOT** derive `Orders.WHERE(YEAR(order_date) == 1992)` twice and join them. Here is the relational plan before implementing the change:

```
ROOT(columns=[('order_priority', order_priority_4), ('order_key', order_key), ('order_total_price', order_total_price)], orderings=[(ordering_0):asc_first])
 PROJECT(columns={'order_key': order_key, 'order_priority_4': order_priority_3, 'order_total_price': order_total_price, 'ordering_0': ordering_0})
  PROJECT(columns={'order_key': order_key, 'order_priority_3': order_priority_3, 'order_total_price': order_total_price, 'ordering_0': order_priority_3})
   PROJECT(columns={'order_key': key, 'order_priority_3': order_priority_1, 'order_total_price': total_price})
    FILTER(condition=RANKING(args=[], partition=[order_priority], order=[(total_price):desc_first]) == 1:int64, columns={'key': key, 'order_priority_1': order_priority_1, 'total_price': total_price})
     JOIN(conditions=[t0.order_priority == t1.order_priority], types=['inner'], columns={'key': t1.key, 'order_priority': t0.order_priority, 'order_priority_1': t1.order_priority, 'total_price': t1.total_price})
      AGGREGATE(keys={'order_priority': order_priority}, aggregations={})
       FILTER(condition=YEAR(order_date) == 1992:int64, columns={'order_priority': order_priority})
        SCAN(table=tpch.ORDERS, columns={'order_date': o_orderdate, 'order_priority': o_orderpriority})
      FILTER(condition=YEAR(order_date) == 1992:int64, columns={'key': key, 'order_priority': order_priority, 'total_price': total_price})
       SCAN(table=tpch.ORDERS, columns={'key': o_orderkey, 'order_date': o_orderdate, 'order_priority': o_orderpriority, 'total_price': o_totalprice})
```

Notice how the subtree that filters orders is derived twice: once aggregated on order priority, then joined with the original. This is needed in general with the `PARTITION(...).x` pattern since the partition could have aggregated its original data (so we can compare rows of `x` to the aggregates) or filtered the partitioned data (e.g. finding singleton partitions, or top groups, etc.) so the `.x` access is effectively filtering to only include rows of `x` that meet certain window-esque criteria. However, it is **NOT** needed in this case because the data of the two subtrees is identical (except for the aggregation, which is also used as the join key).

The final plan **should** look like this:

```
ROOT(columns=[('order_priority', order_priority_4), ('order_key', order_key), ('order_total_price', order_total_price)], orderings=[(ordering_0):asc_first])
 PROJECT(columns={'order_key': order_key, 'order_priority_4': order_priority_3, 'order_total_price': order_total_price, 'ordering_0': ordering_0})
  PROJECT(columns={'order_key': order_key, 'order_priority_3': order_priority_3, 'order_total_price': order_total_price, 'ordering_0': order_priority_3})
   PROJECT(columns={'order_key': key, 'order_priority_3': order_priority, 'order_total_price': total_price})
    FILTER(condition=RANKING(args=[], partition=[order_priority], order=[(total_price):desc_first]) == 1:int64, columns={'key': key, 'order_priority': order_priority, 'total_price': total_price})
     FILTER(condition=YEAR(order_date) == 1992:int64, columns={'key': key, 'order_priority': order_priority, 'total_price': total_price})
      SCAN(table=tpch.ORDERS, columns={'key': o_orderkey, 'order_date': o_orderdate, 'order_priority': o_orderpriority, 'total_price': o_totalprice})
```

The implementation of this optimization is done via a special edge-case in the relational conversion of a PartitionChild access. If the specified pattern is detected before joining (meaning context is just an aggregate with no aggregations), then the child output is the answer (with its expressions slightly shifted) and the context can be ignored.


tests/simple_pydough_functions.py
```diff
@@ -31,6 +31,20 @@ def simple_filter_top_five():
...
+    # type in 1992. Specifically, for each order priority, the key & total
+    # price of the order. Order the results by priority.
+    priorities = PARTITION(
+        Orders.WHERE(YEAR(order_date) == 1992), name="orders", by=order_priority
+    )
+    return (
+        priorities.orders.WHERE(RANKING(by=total_price.DESC(), levels=1) == 1)
+        .CALCULATE(order_priority, order_key=key, order_total_price=total_price)
+        .ORDER_BY(order_priority.ASC())
+    )
```
knassre-bodo, line 45: Goal of the issue is to ensure that the final SQL does **NOT** derive `Orders.WHERE(YEAR(order_date) == 1992)` twice and join them.

tests/test_plan_refsols/multi_partition_access_1.txt
```diff
@@ -1,22 +1,4 @@
 ROOT(columns=[('symbol', symbol)], orderings=[(ordering_0):asc_first])
```
knassre-bodo, line 1: The original code for this query:
```py
def multi_partition_access_1():
    # A use of multiple PARTITION and stepping into partition children that is
    # a no-op.
    data = Tickers.CALCULATE(symbol).TOP_K(5, by=symbol.ASC())
    grps_a = PARTITION(data, name="child_3", by=(currency, exchange, ticker_type))
    grps_b = PARTITION(grps_a, name="child_2", by=(currency, exchange))
    grps_c = PARTITION(grps_b, name="child_1", by=exchange)
    return grps_c.child_1.child_2.child_3
```

The partitions here are no-ops and don't do anything, but before this optimization they would result in useless joins. Now, we just cut to the chase and keep the original data without superfluous joins.

tests/test_plan_refsols/order_info_per_priority.txt
```diff
@@ -0,0 +1,6 @@
+ROOT(columns=[('order_priority', order_priority), ('order_key', order_key), ('order_total_price', order_total_price)], orderings=[(ordering_0):asc_first])
```
knassre-bodo, line 1: Notice, this now has the structure described in the earlier comment about what it should look like.

pydough/conversion/relational_converter.py
```diff
@@ -817,6 +817,22 @@ def translate_partition_child(
         child_output: TranslationOutput = self.rel_translation(
             None, node.subtree, len(node.subtree.pipeline) - 1
         )
+
+        # Special case: when the context is the just-partitioned data, just
+        # return the child without bothering to join them.
+        if (
```
knassre-bodo, line 823: The code **below** this if statement is the normal pattern (join `context` to `child_output` using the aggregation keys as join keys). Here, we just toss out `context` and adjust the `expressions` mapping to make `child` the correct answer. The logic inside this if statement (e.g. the for-loop) mimics what `join_outputs` would do when `child_idx` is None.

See the parts of `join_outputs` that case on `child_idx` when building `out_columns` (the equivalent of our `new_expressions` dictionary). In that version, we copy over elements from `lhs_result` (which would be `context`) but shifted back by one (e.g. `name` becomes `BACK(1).name`) because we have stepped down a level, then add new expressions from `rhs_result` (which would be `child_output`). In the new version, we have the expressions from `child_output` (line 827-829), and we mimic the shifted expressions from `lhs_result` by taking each expression in `child_output` and shifting it back by 1 and adding that to the dictionary as well (since we know that `child_output` can be used as a proxy for `context` due to the identical subtrees).
```

**Comment by knassre-bodo** [comment]
Resolves #290

File: pydough/conversion/relational_converter.py [hunk 5]
```diff
@@ -817,6 +817,22 @@ def translate_partition_child(
         child_output: TranslationOutput = self.rel_translation(
```
vineetg3, line 817: I think line 805 has a misplaced docstring, might want to update that.

vineetg3 (approved):
LGTM! Just a minor documentation correction.

pydough/conversion/relational_converter.py
knassre-bodo: Whoops, good catch.

PR #292, opened 2025-03-12
Add week-based calculations, start of week support, DAYOFWEEK, DAYNAME functions, and WEEK-based DATEDIFF functionality

### LINKED ISSUE #271: Add support for WEEK functionality to datetime functions with configurability
**Goal**: Need to add support for the `week` unit in PyDough datetime functions:
- Add `±<amt> weeks` to `DATETIME` by doing days * 7
- Add `"start of week"` to `DATETIME`
- Add a `DAYOFWEEK` function that  returns the integer of the day of week
- Add a `DAYNAME` function that  returns the string representaiton of the day of the week
- Add support for `WEEK` based `DATEDIFF`

Allow the following aliases for the `week` name: `weeks`, `w`.

The offending question from defog.ai that triggered this need: `How many transactions were made by customers from the USA last week (exclusive of the current week)? Return the number of transactions and total transaction amount.` **(IMPORTANT: the defog.ai questions appears to assume that the start of the week is MONDAY instead of SUNDAY)**

The sqlite code:
```sql
SELECT COUNT(DISTINCT sb.sbTxId) AS num_transactions, SUM(sb.sbTxAmount) AS total_transaction_amount
FROM sbTransaction AS sb
JOIN sbCustomer AS sc
ON sb.sbTxCustId = sc.sbCustId
WHERE LOWER(sc.sbCustCountry) = 'usa'
AND sb.sbTxDateTime >= DATETIM('now',  '-' || ((strftime('%w', 'now') + 6) % 7) || ' days', '-7 days')
AND sb.sbTxDateTime < DATE('now',  '-' || ((strftime('%w', 'now') + 6) % 7) || ' days')
```

How it should be written for PyDough:
```py
selected_transactions.WHERE(
  (LOWER(customer.country) == "usa") &
  (date_time < DATETIME("now", "start of week")) & 
  (date_time >= DATETIME("now", "start of week", "-7 days")) & 
)
result = Broker(
 num_transactions=COUNT(selected_transactions),
 total_transaction_amount=SUM(selected_transactions.amount)
)
```

Under the hood: during SQLGlot conversion, convert `DATETIME(..., 'start of week')` into `DATE(...,  '-' || ((strftime('%w', ...) + 6) % 7) || ' days', 'start of day')` (for sqlite, if the start of the week is Monday). This isn't valid PyDough syntax since we require the modifiers to be string literals, but if it isn't post-conversion that's fine.

Also need to add a PyDough config to control the convention for dates.

**For `start of week` in `DATETIME`**:
- Snowflake has a more complicated configurable setup with a `WEEK_START` parameter which we can be inspired by: https://docs.snowflake.com/en/sql-reference/functions-date-time#label-calendar-weeks-weekdays
- Should make the config value for the start of the week an enum that has enumerations for each day of the week -> the default should be `SUNDAY`?
- The binding for `start of week` should change depending on the dialect vs `start_of_week` config value. For example, since sqlite starts on Sunday, if the `start_of_week` value is set to the values below, the `start of week` formula parameterizes the `X` section in `DATETIME(..., '-' || X || ' days', 'start of day')` as follows:
  - Sunday: `STRFTIME('%w', DATETIME(...))`
  - Monday: `((STRFTIME('%w', DATETIME(...)) + 6) % 7)`
  - Tuesday: `((STRFTIME('%w', DATETIME(...)) + 5) % 7)`
  - Wednesday: `((STRFTIME('%w', DATETIME(...)) + 4) % 7)`
  - Thursday: `((STRFTIME('%w',DATETIME(...)) + 3) % 7)`
  - Friday: `((STRFTIME('%w', DATETIME(...)) + 2) % 7)`
  - Saturday: `((STRFTIME('%w', DATETIME(...)) + 1) % 7)`

**For the `DAYOFWEEK` function**:
  - POSX convention: Sunday = 0, Monday = 1, ..., Saturday = 6 (PostgreSQL follows this convention, as does sqlite)
  - ISO 8601 convention: Sunday = 1, Monday = 2, ..., Saturday = 7
  - Should depend on the `start_of_week` config as well as another config `start_week_as_zero` (True = do POSIX, False = do ISO 8601, the starting term is controlled by `start_of_week`), default is True
  - The idea is that the formula we use for generating the SQLGlot code has to convert from what the dialect does to what the config says. E.g. if the config & dialect are both Sunday, its fine, but if the dialect is Sunday and the config says Monday, the implementation should match that.

**For the `DAYNAME` function**: build a `CASE` statement on `DAYOFWEEK`, parameterizing each of the 7 numbers to one of the 7 dow strings (`"Monday"`, `"Tuesday"`, etc.), where the mapping depends on `start_of_week` and `start_week_as_zero`. For certain dialects can use a function as a workaround.

**For `DATEDIFF` with `week` unit**: 
 - The options are to either do the number of days & divide by 7 (rounding down), or to do it based on the difference in weeks where Friday vs the following Tuesday are considered 1 week apart if the week start is Sat/Sun/Mon/Tue.
 - Current idea: if you want day/7, just do day/7, and have `DATEDIFF` always do week start-based by implementing `DATEDIFF('week', A, B)` as `DATEDIFF('day', DATETIME(A, 'start of week'), DATETIME(A, 'start of week')) / 7`, which will depend on the `start_of_week` config parameter.

**All of these function changes, additions, and config changes need to be documented in the function list & usage documentation**

vineetg3:
Resolves #271. Please see the issue for more details:
This PR introduces several enhancements to the PyDough datetime functions, adding support for week-based operations. The changes include new functionality for calculating days of the week, week names, and week-based differences, as well as configuration options to customize week behavior.  

1. **New Features:**  
   - **DAYOFWEEK Function:**  
     Returns the integer representation of the day of the week for a given date/timestamp. Behavior is configurable via `start_of_week` and `start_week_as_zero` session configs.  
     Example: `DAYOFWEEK("2025-03-20")` returns `3` or `4` depending on configs.  

   - **DAYNAME Function:**  
     Returns the string name of the day of the week for a given date/timestamp.  
     Example: `DAYNAME("2025-03-20")` returns `"Thursday"`.  Note that this does not use any configs as mentioned in the issue. 

   - **DATETIME with "start of week" and "+<amt> weeks/week/w":**  
     Adds support for calculating the start of the week for a given date/timestamp, relative to the configurable `start_of_week`.  
     Example: `DATETIME("2025-03-20", "start of week")` returns `"2025-03-17"` if `start_of_week` is Monday.  
Also, added support for adding or subtracting weeks to/from a given date using the DATETIME function. Weeks are calculated as days * 7.

   - **DATEDIFF with "week":**  
     Adds support for calculating the difference between two dates in terms of weeks, relative to the configurable `start_of_week`.  
     Example: `DATEDIFF("2025-03-17", "2025-03-20", "week")` returns `1` if `start_of_week` is Thursday.  

2. **Session Configurations:**  
   - **`start_of_week`:** Specifies which day is considered the first day of the week. Default is `DayOfWeek.SUNDAY`. Can be set to any value from the `DayOfWeek` enum (e.g., Monday, Tuesday, etc.).  
   - **`start_week_as_zero`:** Determines whether the first day of the week starts at 0 or 1. Default is `True`.  
   
**Auxiliary Change**: Added UTC timezone to tests which uses pandas.now() so as to ensure the default timezone that SQLite uses.


documentation/functions.md
```diff
@@ -511,7 +519,29 @@ Orders.CALCULATE(
...
 ```
 
-The first argument in the `DATEDIFF` function supports the following aliases for each unit of time. The argument is **case-insensitive**, and if a unit is not one of the provided options, an error will be thrown. See [`DATETIME`](#datetime) for the supported units and their aliases. Invalid or unrecognized units will result in an error. 
+The first argument in the `DATEDIFF` function supports the following aliases for each unit of time. The argument is **case-insensitive**, and if a unit is not one of the provided options, an error will be thrown. See [`DATETIME`](#datetime) for the supported units and their aliases. Invalid or unrecognized units will result in an error.
+
+<!-- TOC --><a name="dayofweek"></a>
+
+### DAYOFWEEK
+
+The `DAYOFWEEK` function returns the day of the week for a given date/timestamp. It takes a single argument, which is a date/timestamp, and returns an integer between 1 and 7,or between 0 and 6, depending on the `start_of_week` config and `start_week_as_zero` config. Please see the [Session Configs](./usage.md#session-configs) documentation for more details.
```
knassre-bodo, line 528: Missing a space after `7,`.

pydough/sqlglot/transform_bindings.py
```diff
@@ -72,6 +76,11 @@
 The valid string representations of the second unit.
 """
 
+week_units = ("weeks", "week", "w")
+"""
+The valid string representations of the week unit.
+"""
```
knassre-bodo, line 82: NIT: move this between month & day

pydough/sqlglot/transform_bindings.py
```diff
@@ -185,8 +200,134 @@ def impl(
     return impl
 
 
+def convert_sqlite_days_from_start_of_week(
```
knassre-bodo, line 203: This can be simplified if we find a way to rewrite this so it uses `convert_days_from_start_of_week` as a helper function.

pydough/sqlglot/transform_bindings.py
```diff
@@ -367,9 +556,226 @@ def impl(
...
+                    sqlglot_expressions.Literal.string("Friday"),
+                )
+                .when(
+                    ansi_if_eq_cond(
+                        this=ansi_base_week_day,
+                        expression=sqlglot_expressions.Literal.number(6),
+                    ),
+                    sqlglot_expressions.Literal.string("Saturday"),
+                )
+                .else_(sqlglot_expressions.Null())
```
knassre-bodo, line 768: This isn't actually needed (`else NULL` is implied if there is no `else` in a `CASE` statement`).

pydough/sqlglot/transform_bindings.py
```diff
@@ -367,9 +556,226 @@ def impl(
...
+            ansi_if_eq_cond: SQLGlotExpression = sqlglot_expressions.EQ
+            ansi_answer: SQLGlotExpression = apply_parens(
+                sqlglot_expressions.Case()
+                .when(
+                    ansi_if_eq_cond(
+                        this=ansi_base_week_day,
+                        expression=sqlglot_expressions.Literal.number(0),
+                    ),
+                    sqlglot_expressions.Literal.string("Sunday"),
+                )
```
knassre-bodo, line 725: This can be simplified with a loop:
```py
ansi_answer: SQLGlotExpression = sqlglot_expressions.Case()
for dow, dayname in enumerate(["Sunday", "Monday", ...]):
  ansi_answer = ansi_answer.when(
   sqlglot_expressions.EQ(
    this=ansi_base_week_day,
    expression=sqlglot_expressions.Literal.number(dow),
   ),
  sqlglot_expressions.Literal.string(dayname))
ansi_answer = apply_parens(ansi_answer)
  
```

pydough/sqlglot/transform_bindings.py
```diff
@@ -1366,23 +1806,67 @@ def convert_sqlite_datediff(
...
+                ),
+                to=sqlglot_expressions.DataType(
+                    this=sqlglot_expressions.DataType.Type.INT
+                ),
+            )
+            return answer
         case "days" | "day" | "d":
             # Extracts the start of date from the datetime and subtracts the dates.
+            # We use Date instead of Datetime as the converted SQL does not have
+            # the 'start of day' function for Datetime.
```
knassre-bodo, line 1869: I'm unsure what this means?

documentation/usage.md
```diff
@@ -233,6 +233,34 @@ Suppliers.ORDER_BY(
...
+
+```py
+import pydough
+from pydough.configs import DayOfWeek
+
+# The configs of the active session
+configs = pydough.active_session.config
+
+# Set the start of the week to Monday
+configs.start_of_week = DayOfWeek.MONDAY
```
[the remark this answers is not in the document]
vineetg3: I've done it on the previous lines. I did not do it for DATEDIFF and DATETIME as`start_week_as_zero` config does not these functions.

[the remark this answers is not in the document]
knassre-bodo: Ahh, sorry didn't see

pydough/sqlglot/transform_bindings.py
vineetg3: I made it different so that, when the `transform_bindings` refactor happens, SQLite will have its own helper.

pydough/sqlglot/transform_bindings.py
vineetg3: My bad, this comment should've been deleted. Thanks for pointing it out.

tests/conftest.py
```diff
@@ -35,6 +35,24 @@ def default_config() -> PyDoughConfigs:
...
+
+@pytest.fixture
+def defog_config() -> PyDoughConfigs:
+    """
+    The configuration of PyDoughConfigs used in testing defog.ai standard
+    queries. This is re-created with each request since a test function can
+    mutate this.
+    """
+    config: PyDoughConfigs = PyDoughConfigs()
+    # Set the defaults manually, in case they ever change.
```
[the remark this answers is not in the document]
vineetg3: The function docstring mentions that. The comment on L51 is for the same reason its on L35. Thoughts?

tests/test_pydough_to_sql.py
```diff
@@ -123,15 +132,168 @@ def test_pydough_to_sql_defog(
...
+    empty_context_database: DatabaseContext,
+    update_tests: bool,
+    default_config: PyDoughConfigs,
 ) -> None:
     """
     Tests that a PyDough unqualified node can be correctly translated to its
     sql, with the correct string representation.
     """
+    setattr(default_config, "start_of_week", start_of_week_config)
+    setattr(default_config, "start_week_as_zero", start_week_as_zero_config)
```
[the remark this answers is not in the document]
vineetg3: Awesome! Thanks!

documentation/functions.md
```diff
@@ -511,7 +520,44 @@ Orders.CALCULATE(
...
+<!-- TOC --><a name="dayofweek"></a>
+
+### DAYOFWEEK
+
+The `DAYOFWEEK` function returns the day of the week for a given date/timestamp. It takes a single argument, which is a date/timestamp, and returns an integer between 1 and 7, or between 0 and 6, depending on the `start_of_week` config and `start_week_as_zero` config. Please see the [Session Configs](./usage.md#session-configs) documentation for more details and in-depth examples.
+
+In other words, `DAYOFWEEK` returns which day of the week is the given date/timestamp, where the first day of the give date/timestamp is decided by the `start_of_week` config.
+
+```py
+# Returns the day of the week for the current date
```
hadia206, line 534: The code is not current date. Either chance example to be current date or change this to `order_date`

documentation/functions.md
```diff
@@ -511,7 +520,44 @@ Orders.CALCULATE(
...
+
+
+<!-- TOC --><a name="dayname"></a>
+
+### DAYNAME
+
+The `DAYNAME` function returns the name of the day of the week for a given date/timestamp. It takes a single argument, which is a date/timestamp, and returns a string, corresponding to the name of the day of the week. This returns one of the following: `"Monday"`, `"Tuesday"`, `"Wednesday"`, `"Thursday"`, `"Friday"`, `"Saturday"`, or `"Sunday"`.
+
+```py
+# Returns the name of the day of the week for the current date
```
hadia206, line 558: Same.

documentation/usage.md
```diff
@@ -233,6 +233,95 @@ Suppliers.ORDER_BY(
   key.DESC(),
 )
 ```
+
+The following configs are used in the behavior of `DAYOFWEEK`, `DATETIME`, and `DATEDIFF`:
+
+5. `start_of_week` (default=`DayOfWeek.SUNDAY`): Determines which day is considered the first day of the week. This affects functions like
```
hadia206, line 239:
```suggestion
5. `start_of_week` (default=`DayOfWeek.SUNDAY`): Determines which day is considered the first day of the week. This affects the following functions:
```

documentation/usage.md
```diff
@@ -233,6 +233,95 @@ Suppliers.ORDER_BY(
   key.DESC(),
 )
 ```
+
+The following configs are used in the behavior of `DAYOFWEEK`, `DATETIME`, and `DATEDIFF`:
+
+5. `start_of_week` (default=`DayOfWeek.SUNDAY`): Determines which day is considered the first day of the week. This affects functions like 
+- `DAYOFWEEK` : An integer that returns the number of days since the start of the week. Start of week is relative to the `start_of_week` config.
```
hadia206, line 240:
```suggestion
- `DAYOFWEEK` : A function that returns the number of days since the start of the week. Start of week is relative to the `start_of_week` config.
```

tests/defog_outputs.py
```diff
@@ -214,7 +214,7 @@ def defog_sql_text_broker_adv8() -> str:
     total transaction amount.
     """
     return """
-    SELECT COUNT(DISTINCT sb.sbTxId) AS num_transactions, SUM(sb.sbTxAmount) AS total_transaction_amount
+    SELECT COUNT(DISTINCT sb.sbTxId) AS num_transactions, COALESCE(SUM(sb.sbTxAmount), 0) AS total_transaction_amount
```
hadia206, line 217: why `coalesce` here?

hadia206 (approved):
Overall looks good to me.
Left minor comments and will leave final approval to Kian.

tests/conftest.py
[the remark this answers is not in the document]
knassre-bodo: Well the problem is these **aren't** the defaults, these settings are different bc they are set to match defog.ai. The one from earlier is setting the values to their normal defaults.

tests/defog_outputs.py
knassre-bodo: This is incorrect; the SQL code here is the reference answer, so we need to change the PyDough code to match it instead of changing the reference answer. If you need the sum value to be zero, use `KEEP_IF`. E.g.:
```
KEEP_IF(SUM(x), COUNT(x) > 0)
```

vineetg3: Oh okay, I was not aware. Thanks for pointing it out. I'll make the changes.

PR #293, opened 2025-03-12
Fix bug with downstreaming shadowing over join keys

knassre-bodo:
Consider a PyDough snippet like this:
```py
result = Customers.CALCULATE(
  customer_key=key,
  n_orders=COUNT(orders),
).TOP_K(5, by=(COUNT(orders).DESC(), customer_key.ASC()))
```

The name `customer_key` is also used in `orders` and is the join key between customers & orders. It is fine to use that name in customers because the downstreaming is not overwriting anything explicitly used by `orders` in the `TOP_K` (since the use for join keys is _implicit_). The problem is that currently, during hybrid tree conversion, the name gets shadowed by the downstream-ed use so it is not recognize correctly in the second `COUNT(orders)`. The fix is to slightly adjust how the terms are created for a collection/subcollection in the hybrid tree to allow terms that may have been overwritten by downstreaming just in case they are needed. This also requires a slight change in the API for the collection access QDAG nodes since `get_expr` will not allow accessing these potentially shadowed terms due to its own (correct) error handling.


pydough/conversion/hybrid_tree.py
```diff
@@ -445,20 +445,15 @@ class HybridCollectionAccess(HybridOperation):
...
 
     def __init__(self, collection: CollectionAccess):
-        expr: PyDoughExpressionQDAG
         self.collection: CollectionAccess = collection
         terms: dict[str, HybridExpr] = {}
         for name in collection.calc_terms:
-            # Skip columns that are overloaded with a name from an ancestor,
-            # since they should not be used.
-            if name in collection.ancestral_mapping:
-                continue
```
knassre-bodo, line 455: This was the bad part.

tests/test_pipeline_tpch_custom.py
```diff
@@ -769,15 +770,15 @@
             ),
             id="avg_gap_prev_urgent_same_clerk",
         ),
-        # pytest.param(
```
knassre-bodo, line 772: Not sure how I let this get merged onto main :facepalm:

PR #300, opened 2025-03-17
Add filter pushdown optimizations

### LINKED ISSUE #302: Enable filter pushdown optimizations
For the sake of cleaning up the generated SQL (and also potentially improving its performance), filter should be pushed down as far as possible. This includes the following:
- Pushing filter conditions beneath a project when the project has no window functions and the condition depends on pass-through columns from the project
- Combining adjacent filters and pushing them down further (unless there are window functions)
- Pushing filters through an aggregate when they only depend on the aggregate keys (thus will delete entire groups)
- Pushing filters into the inputs of an inner join (or the left side of a left join) when the conditions only depend on that input
- Pushing filters into the right side of a left join when the filter ocndition is always False if the RHS colmns are null (also transforms the left join into an inner join)

knassre-bodo:
Resolves #302. Set up a relational optimization pipeline after relational conversion, and add an optimization to push filters down as far as possible. Also allows conditions to be pushed into the RHS of a left join (turning it into an inner join) if the RHS being null implies that the filter would output False. Additionally, converted many of the e2e tests (tpch queries, correlated queries) to also test the generated SQL.


tests/test_plan_refsols/correl_14.txt
```diff
@@ -1,18 +1,16 @@
 ROOT(columns=[('n', n)], orderings=[])
  PROJECT(columns={'n': agg_0})
   AGGREGATE(keys={}, aggregations={'agg_0': COUNT()})
-   FILTER(condition=True:bool, columns={'account_balance': account_balance})
```
knassre-bodo, line 4: Nice side effect: these `True` filters also get deleted.

tests/test_plan_refsols/correl_2.txt
```diff
@@ -11,8 +11,8 @@ ROOT(columns=[('name', name_12), ('n_selected_custs', n_selected_custs)], orderi
       FILTER(condition=SLICE(comment_7, None:unknown, 1:int64, None:unknown) == LOWER(SLICE(region_name, None:unknown, 1:int64, None:unknown)), columns={'key': key, 'key_5': key_5})
        JOIN(conditions=[t0.key_5 == t1.nation_key], types=['inner'], columns={'comment_7': t1.comment, 'key': t0.key, 'key_5': t0.key_5, 'region_name': t0.region_name})
         JOIN(conditions=[t0.key == t1.region_key], types=['inner'], columns={'key': t0.key, 'key_5': t1.key, 'region_name': t0.region_name})
-         FILTER(condition=NOT(STARTSWITH(name, 'A':string)), columns={'key': key, 'region_name': region_name})
-          PROJECT(columns={'key': key, 'name': name, 'region_name': name})
+         PROJECT(columns={'key': key, 'region_name': name})
+          FILTER(condition=NOT(STARTSWITH(name, 'A':string)), columns={'key': key, 'name': name})
```
knassre-bodo, line 15: Simple example: filter got pushed before a project

tests/test_plan_refsols/tpch_q2.txt
```diff
@@ -2,31 +2,31 @@ ROOT(columns=[('S_ACCTBAL', S_ACCTBAL), ('S_NAME', S_NAME), ('N_NAME', N_NAME),
...
-           FILTER(condition=name_13 == 'EUROPE':string, columns={'key': key, 'n_name': n_name})
-            JOIN(conditions=[t0.region_key == t1.key], types=['left'], columns={'key': t0.key, 'n_name': t0.n_name, 'name_13': t1.name})
-             PROJECT(columns={'key': key, 'n_name': name, 'region_key': region_key})
-              SCAN(table=tpch.NATION, columns={'key': n_nationkey, 'name': n_name, 'region_key': n_regionkey})
+        JOIN(conditions=[t0.part_key == t1.key], types=['inner'], columns={'key_9': t1.key, 'supplycost': t0.supplycost})
+         JOIN(conditions=[t0.key_5 == t1.supplier_key], types=['inner'], columns={'part_key': t1.part_key, 'supplycost': t1.supplycost})
+          JOIN(conditions=[t0.key == t1.nation_key], types=['inner'], columns={'key_5': t1.key})
+           FILTER(condition=name_3 == 'EUROPE':string, columns={'key': key})
+            JOIN(conditions=[t0.region_key == t1.key], types=['left'], columns={'key': t0.key, 'name_3': t1.name})
```
knassre-bodo, line 13: This filter cant get pushed beneath the join because it would apply to the right side of a left join, potentially changing which rows get matches vs returning a null. This is an example where using `HAS` would positively affect the query, without changing the result, because it enables optimizations (e.g. changing the condition to `HAS(region) & (region.name == "EUROPE")`

tests/test_plan_refsols/triple_partition.txt
```diff
@@ -8,20 +8,20 @@ ROOT(columns=[('region', supp_region), ('avgpct', avg_percentage)], orderings=[(
        AGGREGATE(keys={'cust_region': cust_region, 'part_type': part_type, 'supp_region': supp_region}, aggregations={'agg_0': COUNT()})
         PROJECT(columns={'cust_region': name_15, 'part_type': part_type, 'supp_region': supp_region})
          JOIN(conditions=[t0.customer_key == t1.key], types=['left'], columns={'name_15': t1.name_15, 'part_type': t0.part_type, 'supp_region': t0.supp_region})
-          FILTER(condition=YEAR(order_date) == 1992:int64, columns={'customer_key': customer_key, 'part_type': part_type, 'supp_region': supp_region})
-           JOIN(conditions=[t0.order_key == t1.key], types=['inner'], columns={'customer_key': t1.customer_key, 'order_date': t1.order_date, 'part_type': t0.part_type, 'supp_region': t0.supp_region})
-            PROJECT(columns={'order_key': order_key, 'part_type': part_type, 'supp_region': name_7})
-             JOIN(conditions=[t0.supplier_key == t1.key], types=['left'], columns={'name_7': t1.name_7, 'order_key': t0.order_key, 'part_type': t0.part_type})
-              FILTER(condition=MONTH(ship_date) == 6:int64 & YEAR(ship_date) == 1992:int64, columns={'order_key': order_key, 'part_type': part_type, 'supplier_key': supplier_key})
```
knassre-bodo, line 15: Good examples: these year/month filters got pushed way down to the scans to order/lineitem.

tests/test_plan_refsols/tpch_q2.txt
```diff
@@ -2,31 +2,31 @@ ROOT(columns=[('S_ACCTBAL', S_ACCTBAL), ('S_NAME', S_NAME), ('N_NAME', N_NAME),
...
-           FILTER(condition=name_13 == 'EUROPE':string, columns={'key': key, 'n_name': n_name})
-            JOIN(conditions=[t0.region_key == t1.key], types=['left'], columns={'key': t0.key, 'n_name': t0.n_name, 'name_13': t1.name})
-             PROJECT(columns={'key': key, 'n_name': name, 'region_key': region_key})
-              SCAN(table=tpch.NATION, columns={'key': n_nationkey, 'name': n_name, 'region_key': n_regionkey})
+        JOIN(conditions=[t0.part_key == t1.key], types=['inner'], columns={'key_9': t1.key, 'supplycost': t0.supplycost})
+         JOIN(conditions=[t0.key_5 == t1.supplier_key], types=['inner'], columns={'part_key': t1.part_key, 'supplycost': t1.supplycost})
+          JOIN(conditions=[t0.key == t1.nation_key], types=['inner'], columns={'key_5': t1.key})
+           JOIN(conditions=[t0.region_key == t1.key], types=['inner'], columns={'key': t0.key})
+            SCAN(table=tpch.NATION, columns={'key': n_nationkey, 'region_key': n_regionkey})
+            FILTER(condition=name == 'EUROPE':string, columns={'key': key})
```
knassre-bodo, line 14: This is an interesting case. IN the original version, the filter `name_3 == EUROPE` happened **after** a left-join, and `name_3` returned to a column from the RHS of the left-join. In the new version, we've pushed this filter into the RHS of the join, which has become an inner join, because any row that would have been a null match due to the left join would get deleted by the filter anyway.

tests/test_plan_refsols/singular2.txt
```diff
@@ -2,8 +2,8 @@ ROOT(columns=[('name', name), ('okey', okey)], orderings=[])
  PROJECT(columns={'name': name, 'okey': key_2})
   JOIN(conditions=[t0.key == t1.nation_key], types=['left'], columns={'key_2': t1.key_2, 'name': t0.name})
    SCAN(table=tpch.NATION, columns={'key': n_nationkey, 'name': n_name})
-   FILTER(condition=key_2 == 454791:int64, columns={'key_2': key_2, 'nation_key': nation_key})
```
knassre-bodo, line 5: This is another simple example: previously this filter happened after the join, and now it got pushed into the RHS input.

tests/test_plan_refsols/rank_with_filters_a.txt
```diff
@@ -1,5 +1,4 @@
 ROOT(columns=[('n', n), ('r', r)], orderings=[])
- FILTER(condition=r <= 30:int64, columns={'n': n, 'r': r})
-  FILTER(condition=ENDSWITH(name, '0':string), columns={'n': n, 'r': r})
-   PROJECT(columns={'n': name, 'name': name, 'r': RANKING(args=[], partition=[], order=[(acctbal):desc_first])})
-    SCAN(table=tpch.CUSTOMER, columns={'acctbal': c_acctbal, 'name': c_name})
+ FILTER(condition=r <= 30:int64 & ENDSWITH(name, '0':string), columns={'n': n, 'r': r})
+  PROJECT(columns={'n': name, 'name': name, 'r': RANKING(args=[], partition=[], order=[(acctbal):desc_first])})
+   SCAN(table=tpch.CUSTOMER, columns={'acctbal': c_acctbal, 'name': c_name})
```
knassre-bodo, line 4: Importantly: these filters got merged but neither one got pushed **beneath** the window function call (which would change the outputs to the window function).

tests/test_plan_refsols/correl_15.txt
```diff
@@ -1,22 +1,20 @@
 ROOT(columns=[('n', n)], orderings=[])
  PROJECT(columns={'n': agg_1})
   AGGREGATE(keys={}, aggregations={'agg_1': COUNT()})
-   FILTER(condition=True:bool, columns={'account_balance': account_balance})
-    JOIN(conditions=[t0.key == t1.supplier_key], types=['semi'], columns={'account_balance': t0.account_balance}, correl_name='corr3')
-     PROJECT(columns={'account_balance': account_balance, 'global_avg_price': global_avg_price, 'key': key, 'supplier_avg_price': agg_0})
-      JOIN(conditions=[t0.key == t1.supplier_key], types=['left'], columns={'account_balance': t0.account_balance, 'agg_0': t1.agg_0, 'global_avg_price': t0.global_avg_price, 'key': t0.key})
-       FILTER(condition=nation_key == 19:int64, columns={'account_balance': account_balance, 'global_avg_price': global_avg_price, 'key': key})
```
knassre-bodo, line 8: This also got pushed all the way to the scan to SUPPLIER

tests/test_plan_refsols/correl_16.txt
```diff
@@ -1,14 +1,13 @@
...
-       PROJECT(columns={'key': key, 'rname': name_3})
-        JOIN(conditions=[t0.region_key == t1.key], types=['left'], columns={'key': t0.key, 'name_3': t1.name})
-         SCAN(table=tpch.NATION, columns={'key': n_nationkey, 'region_key': n_regionkey})
-         SCAN(table=tpch.REGION, columns={'key': r_regionkey, 'name': r_name})
-       SCAN(table=tpch.CUSTOMER, columns={'acctbal': c_acctbal, 'key': c_custkey, 'nation_key': c_nationkey})
+   JOIN(conditions=[t0.nation_key == t1.key], types=['semi'], columns={'account_balance': t0.account_balance}, correl_name='corr7')
+    PROJECT(columns={'account_balance': account_balance, 'nation_key': nation_key, 'tile': PERCENTILE(args=[], partition=[], order=[(account_balance):asc_last, (key):asc_last], n_buckets=10000)})
+     SCAN(table=tpch.SUPPLIER, columns={'account_balance': s_acctbal, 'key': s_suppkey, 'nation_key': s_nationkey})
+    FILTER(condition=rname == 'EUROPE':string & PERCENTILE(args=[], partition=[], order=[(acctbal):asc_last, (key_5):asc_last], n_buckets=10000) == corr7.tile, columns={'key': key})
```
knassre-bodo, line 7: The europe filter can NOT get pushed further than this, because it would change the results of the window function call.

tests/test_plan_refsols/tpch_q11.txt
```diff
@@ -7,17 +7,17 @@ ROOT(columns=[('PS_PARTKEY', PS_PARTKEY), ('VALUE', VALUE)], orderings=[(orderin
...
-         FILTER(condition=name_3 == 'GERMANY':string, columns={'availqty': availqty, 'supplycost': supplycost})
-          JOIN(conditions=[t0.supplier_key == t1.key], types=['left'], columns={'availqty': t0.availqty, 'name_3': t1.name_3, 'supplycost': t0.supplycost})
-           SCAN(table=tpch.PARTSUPP, columns={'availqty': ps_availqty, 'supplier_key': ps_suppkey, 'supplycost': ps_supplycost})
-           JOIN(conditions=[t0.nation_key == t1.key], types=['inner'], columns={'key': t0.key, 'name_3': t1.name})
-            SCAN(table=tpch.SUPPLIER, columns={'key': s_suppkey, 'nation_key': s_nationkey})
+         JOIN(conditions=[t0.supplier_key == t1.key], types=['inner'], columns={'availqty': t0.availqty, 'supplycost': t0.supplycost})
+          SCAN(table=tpch.PARTSUPP, columns={'availqty': ps_availqty, 'supplier_key': ps_suppkey, 'supplycost': ps_supplycost})
+          JOIN(conditions=[t0.nation_key == t1.key], types=['inner'], columns={'key': t0.key})
+           SCAN(table=tpch.SUPPLIER, columns={'key': s_suppkey, 'nation_key': s_nationkey})
+           FILTER(condition=name == 'GERMANY':string, columns={'key': key})
```
knassre-bodo, line 14: These `GERMANY` filters got pushed all the way to the scans

tests/test_plan_refsols/tpch_q9.txt
```diff
@@ -6,14 +6,14 @@ ROOT(columns=[('NATION', NATION), ('O_YEAR', O_YEAR), ('AMOUNT', AMOUNT)], order
      PROJECT(columns={'nation_name': nation_name, 'o_year': YEAR(order_date), 'value': extended_price * 1:int64 - discount - supplycost * quantity})
       JOIN(conditions=[t0.order_key == t1.key], types=['left'], columns={'discount': t0.discount, 'extended_price': t0.extended_price, 'nation_name': t0.nation_name, 'order_date': t1.order_date, 'quantity': t0.quantity, 'supplycost': t0.supplycost})
        JOIN(conditions=[t0.part_key == t1.part_key & t0.supplier_key == t1.supplier_key], types=['inner'], columns={'discount': t1.discount, 'extended_price': t1.extended_price, 'nation_name': t0.nation_name, 'order_key': t1.order_key, 'quantity': t1.quantity, 'supplycost': t0.supplycost})
-        FILTER(condition=CONTAINS(name_7, 'green':string), columns={'nation_name': nation_name, 'part_key': part_key, 'supplier_key': supplier_key, 'supplycost': supplycost})
```
knassre-bodo, line 9: This condition is also pushable beneath a left join (turning it to inner) since if `name_7` is null, the `CONTAINS` condition will be False.

pydough/sqlglot/sqlglot_relational_visitor.py
```diff
@@ -476,7 +476,7 @@ def visit_aggregate(self, aggregate: Aggregate) -> None:
             )
         if keys:
             if aggregations:
-                query = query.group_by(*keys)
+                query = query.group_by(*sorted(keys))
```
knassre-bodo, line 479: This fixes a nondeterminism issue in the generated SQL

tests/test_relational_to_sql.py
```diff
@@ -709,55 +704,6 @@ def test_convert_relation_to_sqlite_sql(
...
-            id="tpch_q3",
-        ),
-        pytest.param(
-            tpch_query_6_plan(),
-            "tpch_q6",
-            id="tpch_q6",
-        ),
-    ],
-)
-def test_tpch_relational_to_sqlite_sql(
```
knassre-bodo, line 732: Redundant with `test_pipeline_until_sql_tpch`

pydough/conversion/filter_pushdown.py
```diff
@@ -0,0 +1,138 @@
...
+def push_filters(
+    node: RelationalNode, filters: set[RelationalExpression]
+) -> RelationalNode:
+    """
+    Transpose filter conditions down as far as possible.
+
+    Args:
+        `node`: The current node of the relational tree.
+        `filters`: The set of filter conditions to push down representing
+        predicates from ancestor nodes that can be pushed this far down.
```
vineetg3, line 41: The `Returns:` part of the docstring is missing.

pydough/relational/rel_util.py
```diff
@@ -0,0 +1,297 @@
...
+    pydop.JOIN_STRINGS,
+    pydop.ADD,
+    pydop.SUB,
+    pydop.MUL,
+    pydop.DIV,
+}
+"""
+A set of operators with the property that the output is null if any of the
+inputs are null.
+"""
```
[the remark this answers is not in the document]
vineetg3: I think the reasoning is so that when you hover on the variable in VSCode the details come up on the tip tool .

vineetg3 (approved):
LGTM!

pydough/conversion/filter_pushdown.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    remaining_filters: set[RelationalExpression]
+    pushable_filters: set[RelationalExpression]
+    match node:
+        case Filter():
+            remaining_filters = set()
```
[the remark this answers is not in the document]
knassre-bodo: Whoops, good catch

pydough/conversion/relational_converter.py
```diff
@@ -1091,6 +1092,77 @@ def make_relational_ordering(
...
+    ordered_columns: list[tuple[str, RelationalExpression]] = []
+    orderings: list[ExpressionSortInfo] | None = None
+    renamings: dict[str, str] = hybrid.pipeline[-1].renamings
+    hybrid_expr: HybridExpr
+    rel_expr: RelationalExpression
+    name: str
+    original_name: str
+    if columns is None:
+        for original_name in node.calc_terms:
+            name = renamings.get(original_name, original_name)
```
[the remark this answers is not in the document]
knassre-bodo: Yes, its saying that if `original_name `is not a key in `renamings`, then just use `original_name ` instead of doing the dictionary lookup.

pydough/relational/rel_util.py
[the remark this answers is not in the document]
knassre-bodo: Yeah, the convention is for it to be below.

pydough/relational/rel_util.py
```diff
@@ -0,0 +1,297 @@
...
+
+    Args:
+        `expr`: The expression to extract conjunctions from.
+
+    Returns:
+        The set of filter conditions whose conjunction forms `expr`.
+    """
+    if isinstance(expr, LiteralExpression) and expr.value:
+        return set()
+    if isinstance(expr, CallExpression) and expr.op == pydop.BAN:
```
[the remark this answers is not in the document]
knassre-bodo: Fair, both work.

pydough/conversion/filter_pushdown.py
```diff
@@ -0,0 +1,138 @@
...
+                pushable_filters, remaining_filters = partition_expressions(
+                    filters,
+                    lambda expr: only_references_columns(expr, allowed_cols),
+                )
+                pushable_filters = {
+                    transpose_expression(expr, node.columns)
+                    for expr in pushable_filters
+                }
+            node._input = push_filters(node.input, pushable_filters)
+            return build_filter(node, remaining_filters)
```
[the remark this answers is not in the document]
knassre-bodo: The idea is that the predicate in `partition_expressions` tells us **which** filters we can push (the rest get materialized at this stage). E.g. for aggregates, we can only push filters that only depend on key columns (since those will delete entire groups, instead of deleting data within a group_), so we pass in a predicate that returns True if the expression only depends on the key columns. `build_filter` materializes whatever **can't** be pushed.

tests/test_pipeline_defog.py
```diff
@@ -375,6 +375,38 @@ def defog_test_data(
     return request.param
 
 
+def test_defog_until_sql(
```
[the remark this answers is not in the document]
knassre-bodo: The idea is that it tests each defog query by running it through the pipeline up until SQL conversion. The ones that go up to relational plans are called `_until_relational`, and the ones that go all the way to runtime execution are called `e2e`.

PR #304, opened 2025-03-20
Adjust how datetime arguments are handled for literals

knassre-bodo:
Making several changes to how datetime functions & arguments are handled for the sake of robustness & cleanliness of SQL:
- Ensured that DATEDIFF and the datetime extraction functions (year, etc.) work when the input is a string literal (e.g. `"now"`, or `"2024-03-01 12:30:59"`)
- Changed how datetime literals are converted into SQLGlot in the ANSI dialect to always directly create a CAST call instead of using the `TIME_STR_TO_TIME` or `DATE_STR_TO_DATE` functions (which don't exist and are just SQLGlot internals)
- Updated SQL generation tests to increase coverage for this behavior
- Ancillary change: moved some string parsing logic for datetimes to a regex for simplicity/reuse
- Ancillary change: removed a specific use `@cache` decorator in the QDAG nodes that a profiler revealed to be causing performance issues on very specific tests


pydough/qdag/collections/calculate.py
```diff
@@ -148,8 +148,7 @@ def get_expression_position(self, expr_name: str) -> int:
             raise PyDoughQDAGException(f"Unrecognized CALCULATE term: {expr_name!r}")
         return self.calc_term_indices[expr_name]
 
-    @cache
-    def get_term(self, term_name: str) -> PyDoughQDAG:
+    def get_term(self, term_name: str) -> PyDoughQDAG:  # type: ignore
         if term_name in self.calc_term_values:
```
knassre-bodo, line 152: I ran a profiler and it turns out this **specific** use of `@cache` accounted for a disproportionate performance bottleneck, to the point that certain extreme tests took over a second to convert to SQL when run with multiple dialects. We still get a speedup here because the **superclass** already caches. The `# type: ignore` is so mypy will ignore the fact that the signature is different than the superclass (bc the superclass version is cached but this one isn't)

pydough/sqlglot/sqlglot_relational_expression_visitor.py
```diff
@@ -160,6 +162,26 @@ def visit_literal_expression(self, literal_expression: LiteralExpression) -> Non
         literal: SQLGlotExpression = sqlglot_expressions.convert(
             literal_expression.value
         )
+
+        # Special handling: insert cast calls for ansi casting of date/time
+        # instead of relying on SQLGlot conversion functions.
+        if self._bindings.dialect == DatabaseDialect.ANSI:
```
knassre-bodo, line 168: This ensures we **always** generate the cast calls instead of that nonsensical `TIME_STR_TO_TIME` function for the ANSI files.

pydough/sqlglot/transform_bindings.py
```diff
@@ -293,19 +303,17 @@ def handle_datetime_base_arg(
...
     # datetime should be used.
     if isinstance(arg, sqlglot_expressions.Literal) and arg.is_string:
-        if str(arg.this).lower().strip() in (
-            "now",
-            "current_timestamp",
-            "current_date",
-            "current timestamp",
-            "current date",
-        ):
+        if current_ts_pattern.fullmatch(arg.this):
```
knassre-bodo, line 306: Moved this into a regex for compactness & future extensions.

tests/simple_pydough_functions.py
```diff
@@ -610,6 +610,60 @@ def datetime_sampler():
             " \n \t312 \nD \n\r ",
         ),
         DATETIME(" Current_Date ", "+ 45 MM", "-135 s"),
+        YEAR("Current Date"),
```
knassre-bodo, line 613: Added these patterns to make sure the various edge cases with datetime & literalhandling were covered.

tests/test_sql_refsols/datediff_ansi.sql
```diff
@@ -22,16 +22,16 @@ FROM (
     years_diff
   FROM (
     SELECT
-      DATEDIFF(TIME_STR_TO_TIME('2025-05-02 11:00:00'), date_time, DAY) AS days_diff,
```
knassre-bodo, line 25: `TIME_STR_TO_TIME` is a dummy function inside of SQLGlot that each dialect has its own logic for translating into valid SQL, but when we are doing ANSI there is no translation so it just returns this dummy function. My changes in this PR forced it to always use a `CAST` instead, just for ANSI, so the ansi sql files are more legible.

pydough/sqlglot/transform_bindings.py
```diff
@@ -293,19 +303,17 @@ def handle_datetime_base_arg(
...
             if dialect == DatabaseDialect.SQLITE:
                 return sqlglot_expressions.Datetime(
                     this=[sqlglot_expressions.convert("now")]
                 )
             else:
                 return sqlglot_expressions.CurrentTimestamp()
+    if dialect != DatabaseDialect.SQLITE:
```
J-Solano-bodo-ai, line 313: For simplicity's sake I would recommend unless the opposite is needed
```
if dialect == DatabaseDialect.SQLITE:
        return sqlglot_expressions.Datetime(this=[arg])

    return sqlglot_expressions.Cast(
            this=arg, to=sqlglot_expressions.DataType.build("TIMESTAMP")
    )
```

pydough/sqlglot/transform_bindings.py
```diff
@@ -1207,20 +1215,19 @@ def convert_ndistinct(
...
     """
-    Creates a function that extracts a specific time unit
-    (e.g., HOUR, MINUTE, SECOND) from a SQLGlot expression.
+    Creates a function that extracts a specific date/time unit from a SQLGlot
+    expression.
 
     Args:
-        `unit`: The time unit to extract. Must be one of 'HOUR', 'MINUTE',
-                or 'SECOND'.
+        `unit`: The time unit to extract.
```
J-Solano-bodo-ai, line 1224: I think having the available options in the comment (second, hour, minute, year, etc.) would be useful unless they're stated elsewhere

J-Solano-bodo-ai (approved):

pydough/sqlglot/sqlglot_relational_expression_visitor.py
hadia206: Add reason of not relying on SQLGlot functions or link to issue (if there's one to explain that)

hadia206: Also, what if the dialect is not ANSI?

hadia206 (approved):
Looks good to me

pydough/sqlglot/sqlglot_relational_expression_visitor.py
knassre-bodo: > Add reason of not relying on SQLGlot functions

Will do.

> Also, what if the dialect is not ANSI?

The real dialects have actual implementations of `TIME_STR_TO_TIME` (e.g. they cast the string to a timestamp). The problem is when doing ansi, we're just using the SQLGlot base class for the dialects (that isn't meant to be used directly for SQL generation), which doesn't have any such function, so it just leaves the dummy function in the SQL text.

vineetg3 (approved):
LGTM. Added a small note about design.

pydough/sqlglot/sqlglot_relational_expression_visitor.py
```diff
@@ -160,6 +162,31 @@ def visit_literal_expression(self, literal_expression: LiteralExpression) -> Non
...
+            if isinstance(literal_expression.value, datetime.datetime):
+                dt: datetime.datetime = literal_expression.value
+                if dt.tzinfo is not None:
+                    raise ValueError(
+                        "PyDough does not yet support datetime values with a timezone"
+                    )
+                literal = sqlglot_expressions.Cast(
+                    this=sqlglot_expressions.convert(dt.isoformat(sep=" ")),
+                    to=sqlglot_expressions.DataType.build("TIMESTAMP"),
+                )
```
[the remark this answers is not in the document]
knassre-bodo: No, because this is for handling **literals**, and transform bindings handles **function calls**.

PR #306, opened 2025-03-21
Add support for MEDIAN function in PyDough

### LINKED ISSUE #305: Add support for MEDIAN aggregation function in PyDough
Goal: allow calling `MEDIAN` as an aggregation function in PyDough. This is complicated by the fact that many dialects don't have a median equivalent, so a convoluted rewrite must be achieved via window functions.

The rewrite for `MEDIAN(subcollection.data_column)` should be a follows (using the window funcitons from #160):

```py
# levels = height of `subcollection`
# (e.g. 1 for `MEDIAN(orders.total_price)` or 3 for `MEDIAN(nations.customers.orders.total_price)`
n = RELCOUNT(data_column, levels=levels)
r = RANKING(by=data_column.DESC(na_pos="last"), levels=levels)
is_median_row = ABS(r - 1 - ((n - 1.0) / 2.0)) < 1
selected_data = KEEP_IF(data_column, is_median_row)
median = AVG(subcollection.CALCULATE(_expr=selected_data)._expr)
```

The logic for this rewrite is as follows:
- `n` counts how many non-null values of `data_column` exist
- `r` assigns each value a ranking based on its position out of all values when sorted by `data_column` (nulls placed at the end so they don't influence the rankings)
- If `n` is odd, then `ABS(r - 1 - ((n - 1.0) / 2.0))` is 0 for the median row
- If `n` is even, then `ABS(r - 1 - ((n - 1.0) / 2.0))` is 0.5 for the two rows that should be averaged to take the median
- Taking the average of `KEEP_IF(data_column, is_median_row)` means either selecting the single row with where it is 0, or taking the average of the two rows where it is 0.5

**IMPORTANT**: need to ensure that the result is null even if the `avg_default_zero` config is True. The rewrite as-written would coalesce the `AVG` output value with 0, which is undesired here.

The tricky part is that this will almost certainly need to happen during the qdag or hybrid stages, since that is the only way to generate some of these terms correctly. However, this rewrite should not happen for a dialect that has an actual median function (e.g. Snowflake), so the choice to make the should be dialect-aware

The function must be documented in functions.md.

knassre-bodo:
Resolves #305. Adds the `MEDIAN` function via convoluted rewrite during hybrid conversion. As a side effect, the need arises to ensure that any inputs to aggregate calls in the `aggs` dictionary of a hybrid connection are hybrid references, since if correlated references exist inside of them (which can happen due to the median rewrite) this can cause bugs during decorrelation. As a result, a step is added to eject any non-reference inputs into a calculate in the subtree. Also fixes a bunch of typos.


tests/test_plan_refsols/aggregate_on_function_call.txt
```diff
@@ -2,5 +2,6 @@ ROOT(columns=[('nation_name', nation_name), ('avg_consumer_value', avg_consumer_
  PROJECT(columns={'avg_consumer_value': agg_0, 'nation_name': key})
   JOIN(conditions=[t0.key == t1.nation_key], types=['left'], columns={'agg_0': t1.agg_0, 'key': t0.key})
    SCAN(table=tpch.NATION, columns={'key': n_nationkey})
-   AGGREGATE(keys={'nation_key': nation_key}, aggregations={'agg_0': MAX(IFF(acctbal < 0.0:float64, 0.0:float64, acctbal))})
-    SCAN(table=tpch.CUSTOMER, columns={'acctbal': c_acctbal, 'nation_key': c_nationkey})
+   AGGREGATE(keys={'nation_key': nation_key}, aggregations={'agg_0': MAX(expr_1)})
+    PROJECT(columns={'expr_1': IFF(acctbal < 0.0:float64, 0.0:float64, acctbal), 'nation_key': nation_key})
+     SCAN(table=tpch.CUSTOMER, columns={'acctbal': c_acctbal, 'nation_key': c_nationkey})
```
knassre-bodo, line 7: The changes to these files (& corresponding sql files) is a result of expressions getting ejected out of aggregates into a child projection. For most of these queries it is fine either way, but the ejection is necessary for certain patterns so it is simpler to just _always_ eject.

pydough/conversion/hybrid_tree.py
```diff
@@ -1801,6 +1871,103 @@ def make_agg_call(
...
+        assert len(converted_args) == 1
+        data_expr: HybridExpr = converted_args[0]
+        one: HybridExpr = HybridLiteralExpr(Literal(1.0, Float64Type()))
+        two: HybridExpr = HybridLiteralExpr(Literal(2.0, Float64Type()))
+        partition_args: list[HybridExpr] = []
+        self.stack.append(hybrid)
+        self.add_unique_terms(
+            child_connection.subtree, child_height, 0, partition_args, child_idx
+        )
+        self.stack.pop()
```
knassre-bodo, line 1919: This is the procedure for identifying the partition arguments for a window call (elsewhere in the same file)

pydough/conversion/hybrid_tree.py
```diff
@@ -2071,7 +2251,9 @@ def make_hybrid_expr(
                             inside_agg or expr.operator.is_aggregation,
                         )
                     )
-                if expr.operator.is_aggregation:
+                if expr.operator == pydop.MEDIAN and self.rewrite_median:
```
hadia206, line 2254: what if `self.rewrite_median` is False?

pydough/pydough_operators/expression_operators/README.md
```diff
@@ -119,10 +119,11 @@ These functions can be called on plural data to aggregate it into a singular exp
 
 ##### Simple Aggregations
 
-- `SUM`: returns the result of adding all of the values of a plural expression.
-- `AVG`: returns the result of taking the average of the values of a plural expression.
-- `MIN`: returns the largest out of the values of a plural expression.
-- `MAX`: returns the smallest out of the values of a plural expression.
+- `SUM`: returns the result of adding all of the non-null values of a plural expression.
+- `AVG`: returns the result of taking the average of the non-null values of a plural expression.
+- `AVG`: returns the result of taking the median of the non-null values of a plural expression.
```
hadia206, line 124: Typo?
```suggestion
- `MEDIAN`: returns the result of taking the median of the non-null values of a plural expression.
```

tests/defog_test_functions.py
```diff
@@ -1057,15 +1057,15 @@ def impl_defog_ewallet_gen1():
     whose category contains 'retail'
     """
     active_merchants = Merchants.WHERE(
-        (CONTAINS(LOWER(category), "%retail%")) & (status == "active")
+        (CONTAINS(LOWER(category), "retail")) & (status == "active")
```
hadia206, line 1060: why remove `%`?

tests/defog_test_functions.py
```diff
@@ -1057,15 +1057,15 @@ def impl_defog_ewallet_gen1():
...
     )
 
-    latest_balance_today = balances.WHERE(
-        DATE(updated_at)
-        == DATE(DATETIME("now")) & (RANKING(updated_at.DESC(), levels=1) == 1)
+    latest_balance_today = active_merchants.balances.WHERE(
+        (DATETIME(updated_at, "start of day") == DATETIME("now", "start of day"))
```
hadia206, line 1064: what `DATETIME` ?

hadia206 (approved):
Looks good to me.
I skimmed the files in `test_sql_refsols`, let me know if I need to focus on them.

pydough/conversion/hybrid_tree.py
knassre-bodo: Then it just leaves `MEDIAN` alone as a regular function call that gets converted to SQLGlot. The point of this rewrite is that we need to intercept a lot sooner since this kind of rewrite _isn't_ possible during the SQLGlot conversion step (too much information has been lost by then).

pydough/pydough_operators/expression_operators/README.md
knassre-bodo: Whoops, thanks

tests/defog_test_functions.py
knassre-bodo: This was a typo in the original. The query says `LOWER(category) LIKE '%retail'`, which is  the same as `CONTAINS(LOWER(category), "retail")`.

tests/defog_test_functions.py
knassre-bodo: `DATETIME` is a PyDough function that does a lot of things, including truncation/offsets. However, there is no `DATE` function in PyDough (in SQLite, date/datetime = same function except date truncates to y/m/d, while datetime always includes h/m/s as well). In PyDough, if you want to truncate to y/m/d, you need to include `"start of day"`

pydough/conversion/hybrid_tree.py
```diff
@@ -1801,6 +1896,80 @@ def make_agg_call(
...
         return self.postprocess_agg_output(hybrid_call, result_ref, joins_can_nullify)
 
+    def rewrite_median_call(
+        self,
+        child_connection: HybridConnection,
+        expr: HybridFunctionExpr,
+        create_new_calc: bool,
+    ) -> HybridFunctionExpr:
+        """
+        Transforms a MEDIAN call into an into an AVG of the 1-2 median rows
```
vineetg3, line 1906:
```suggestion
        Transforms a MEDIAN call into an AVG of the 1-2 median rows
```

pydough/conversion/hybrid_tree.py
```diff
@@ -1801,6 +1896,80 @@ def make_agg_call(
...
+        (obtained via window functions). This step must be done after
+        de-correlation because it invokes the aggregation keys use for the
```
vineetg3, line 1908:
```suggestion
        de-correlation because it invokes the aggregation keys used for the
```

pydough/qdag/abstract_pydough_qdag.py
```diff
@@ -10,8 +10,8 @@
 
 class PyDoughQDAG(ABC):
     """
-    Base class used for PyDough collection, expression, and opertor QDAG
```
vineetg3, line 13: For my understanding, which super class are operator nodes inheriting from?

vineetg3 (approved):
LGTM. Minor typos and a query.

pydough/qdag/abstract_pydough_qdag.py
knassre-bodo: The abstract `PyDoughOperator` base class (which **used** to be a subclass of `PyDoughQDAG`, but hasn't been for some time).

PR #309, opened 2025-03-24
Integrate SQLGlot's optimizer

vineetg3:
Resolves #209. This PR involved utilizing the `optimize` API provided by SQLGlot. This required few changes to "standardize" the AST generation which includes
- giving table alias all the time
- keeping brackets for few specific cases etc. 
- eliminating semi and anti join with sqlglot's `eliminate_semi_and_anti_joins` API.
Few rules(SQLGlot optimize applies rules to the AST one after the another) were skipped in this implementation as mentioned in issue #313 .


pydough/sqlglot/join_type_relational_visitor.py
```diff
@@ -0,0 +1,146 @@
+"""
```
knassre-bodo, line 1: Two things:
1. Let's move this to the `relational/relational_nodes` folder
2.. Lets add this file to that module's readme (see how `TreeStringVisitor` is mentioned).

pydough/sqlglot/execute_relational.py
```diff
@@ -40,14 +48,90 @@ def convert_relation_to_sql(
...
+    glot_expr = normalize(glot_expr)
+    glot_expr = pushdown_predicates(glot_expr, dialect=dialect)
+    glot_expr = optimize_joins(glot_expr)
+    glot_expr = eliminate_subqueries(glot_expr)
+
+    join_types = set(JoinTypeRelationalVisitor().get_join_types(relational))
+    if JoinType.ANTI not in join_types and JoinType.SEMI not in join_types:
+        glot_expr = merge_subqueries(glot_expr)
+
+    glot_expr = eliminate_joins(glot_expr)
```
knassre-bodo, line 99: Let's add a comment to say what each of these does (& comments for any missing ones)

knassre-bodo (commented):
Hi Vineet! Here's my first round of reviews. I anticipate we'll need a few rounds of back & forth as you continue to revise, particularly with my comment about passing in the dialect to `parse_one`.

pydough/sqlglot/execute_relational.py
```diff
@@ -40,14 +48,90 @@ def convert_relation_to_sql(
...
         str: The SQL string representing the relational tree.
     """
-    # TODO (gh #205): use simplify/optimize from sqlglot to rewrite the
-    # generated SQL.
     glot_expr: SQLGlotExpression = SQLGlotRelationalVisitor(
         dialect, bindings, config
     ).relational_to_sqlglot(relational)
+
+    glot_expr = parse_one(glot_expr.sql(dialect))
+    glot_expr = apply_sqlglot_optimizer(glot_expr, relational, dialect)
```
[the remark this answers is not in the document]
vineetg3: Yes, this worked. I reverted my Datetime changes :).

tests/test_logging.py
```diff
@@ -182,80 +182,48 @@ def test_execute_df_logging(
...
+    SUM(
+      "lineitem"."l_extendedprice" * (
+        1 - "lineitem"."l_discount"
+      ) * (
+        1 + "lineitem"."l_tax"
+      )
+    ) AS "sum_charge",
+    "lineitem"."l_linestatus" AS "l_linestatus",
+    "lineitem"."l_returnflag" AS "l_returnflag"
+  FROM "lineitem" AS "lineitem"
```
knassre-bodo, line 205: This kind of thing is ugly. Can you add your own extra pass at the end that recursively visits/transforms all of these SQLGlot nodes & does the following:
- Remove any aliases in the form `x AS x`
- Within each `SELECT`, if there is only a single input, drop all of the qualified prefixes, so it just uses the column names  (e.g. `lineitem.l_linestatus` in this `SELECT` clause becomes just `l_linestatus`, same for `_t0.sum_charge` in the one below)

pydough/sqlglot/sqlglot_relational_visitor.py
```diff
@@ -82,7 +83,7 @@ def _generate_table_alias(self) -> str:
         Returns:
             str: A unique table alias.
         """
-        alias = f"_table_alias_{self._alias_counter}"
+        alias = f"_t{self._alias_counter}"
```
knassre-bodo, line 86: I know I suggested this, but it looks like this change resulted in some weird name generations by SQLGlot (e.g. `_t3_2`). Can you try experimenting with `_s` so its the same overall visual look but without these weird collisions? I'll leave it to your judgement if the overall result looks better that way or the way it is right now with the `_t` aliases with mixed numberings.

knassre-bodo (commented):
Fantastic job @vineetg3! There are a couple more things I want you to experiment with iterating upon, for the sake of making the generated SQL as nice as possible, but this PR is definitely close to being done!

pydough/sqlglot/execute_relational.py
```diff
@@ -37,15 +56,129 @@ def convert_relation_to_sql(
...
+    # Remove unused joins from an expression.
+    # This only removes joins when we know that the join condition doesn't
+    # produce duplicate rows.
+    glot_expr = eliminate_joins(glot_expr)
+
+    # Remove unused CTEs from an expression.
+    glot_expr = eliminate_ctes(glot_expr)
+
+    # NEW: Makes sure all identifiers that need to be quoted are quoted.
+    glot_expr = quote_identifiers(glot_expr, dialect=dialect)
```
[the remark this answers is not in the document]
vineetg3: I kept it because the optimizer suggests that we apply rules one by one and I'd like to think that it would break some rule down the line.

pydough/sqlglot/sqlglot_relational_visitor.py
vineetg3: This occurs because `eliminate_subqueries` take the previous alias of the subquery as the base alias for rest of the subqueries.

knassre-bodo: 🤦 ok let's just keep this `_t` version then.

pydough/sqlglot/execute_relational.py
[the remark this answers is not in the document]
knassre-bodo: Can we try w/o this? Its kinda ugly to have all of these extra quotes when we don't need them. Let's just quote the `AS xxx` part of the final SELECT.

tests/test_sql_refsols/defog_broker_adv2_ansi.sql
```diff
@@ -1,57 +1,28 @@
...
-          ticker_id
-      )
-        ON _id = ticker_id
-    )
-  )
+    COALESCE(_t1.agg_0, 0) AS ordering_1,
+    sbticker.sbtickersymbol AS symbol,
+    COALESCE(_t1.agg_0, 0) AS tx_count
+  FROM main.sbticker AS sbticker
+  LEFT JOIN _t1 AS _t1
```
vineetg3, line 17: This could technically become un-aliased, for now I'm leaving that case, but open to doing it if needed right now.

pydough/sqlglot/execute_relational.py
```diff
@@ -37,15 +56,181 @@ def convert_relation_to_sql(
...
+    Returns:
+        None (The AST is modified in place.)
+    """
+    if isinstance(expr, Select) and (
+        expr.args.get("joins") is None or len(expr.args.get("joins")) == 0
+    ):
+        from_clause = expr.args.get("from")
+        if from_clause is not None:
+            if not isinstance(from_clause.this, Table):
+                return
```
knassre-bodo, line 202: It shouldn't `return` in this case, since that prevents us from visiting anything further inside. Instead, just ensure lines 204-222 are not executed if it isn't a table, but still do lines 225-231.

pydough/sqlglot/execute_relational.py
```diff
@@ -37,15 +56,181 @@ def convert_relation_to_sql(
...
+            # Table(this=Identifier(this=..),..)
+            table = from_clause.this
+            actual_table_name = table.name
+            # Table(this=..,alias=TableAlias(this=Identifier(this=..)))
+            alias = table.alias
+            if str(actual_table_name) == str(alias):
+                # Remove cases like `..FROM t1 as t1..`
+                table.args.pop("alias")
```
knassre-bodo, line 210: The removing of aliases can also be done even outside of the line 200 condition. What I recommend is that between this main block from 200-222, and the recursive part in 225-231, you add a condition that checks if `expr` is an Alias with the same input & output name, and if so you get rid of the alias. To do so, you'll probably need to make it so this function returns the answer, and make 225-231 modify the arguments of `expr` accordingly

knassre-bodo (commented):
Great job Vineet! We're almost done here, but lets go a step further with the removing of the `_t1 AS _t1` aliases (both for tables & for columns).

tests/test_sql_refsols/tpch_q8_ansi.sql
```diff
@@ -1,168 +1,42 @@
...
+  JOIN tpch.region AS region
+    ON nation_2.n_regionkey = region.r_regionkey AND region.r_name = 'AMERICA'
   GROUP BY
-    o_year
+    EXTRACT(YEAR FROM orders.o_orderdate)
 )
+SELECT
+  _t0.o_year AS O_YEAR,
+  COALESCE(_t0.agg_0, 0) / COALESCE(_t0.agg_1, 0) AS MKT_SHARE
+FROM _t0_2 AS _t0
```
knassre-bodo, line 42: Why didn't this one get the `_t0` alias deleted?

pydough/sqlglot/execute_relational.py
```diff
@@ -37,15 +56,181 @@ def convert_relation_to_sql(
...
+        from_clause = expr.args.get("from")
+        if from_clause is not None:
+            if not isinstance(from_clause.this, Table):
+                return
+            # Table(this=Identifier(this=..),..)
+            table = from_clause.this
+            actual_table_name = table.name
+            # Table(this=..,alias=TableAlias(this=Identifier(this=..)))
+            alias = table.alias
+            if str(actual_table_name) == str(alias):
```
knassre-bodo, line 208: Also, why do we need this condition here? We should do lines 209-222 even in a situation like `SELECT ... FROM t1 AS t2`

pydough/sqlglot/execute_relational.py
vineetg3: Is this for the cases in the Select clause?

knassre-bodo (approved):
LGTM! Fantastic job with this rather intense issue!

PR #312, opened 2025-03-26
Unskipping defog EWALLET test after added WEEK support

knassre-bodo:
See title.


tests/test_pipeline_defog.py
```diff
@@ -754,6 +751,7 @@ def test_defog_until_sql(
     defog_test_data: PyDoughSQLComparisonTest,
     defog_graphs: graph_fetcher,
     empty_context_database: DatabaseContext,
+    defog_config: PyDoughConfigs,
```
knassre-bodo, line 754: This is ensuring the SQL generation tests use the same configs as the e2e tests for defog (start of week is Monday).

J-Solano-bodo-ai (approved):

PR #314, opened 2025-03-28
Add projection merging to PyDough

knassre-bodo:
Adding a step to the relational optimization pipeline to merge adjacent projection/root/scan nodes, where possible. Also makes an ancillary change to how subquery names are generated by changing the prefix from `_s` to `_t` to avoid a strange bug caused by naming collisions with SQLGlot that was exposed during this PR.


pydough/conversion/hybrid_tree.py
```diff
@@ -2316,12 +2316,17 @@ def process_hybrid_collations(
...
-            )
-            new_expressions[name] = expr
+            if type(collation.expr) is Reference:
+                name = collation.expr.term_name
+            else:
+                name = self.get_ordering_name(hybrid)
+                expr = self.make_hybrid_expr(
+                    hybrid, collation.expr, child_ref_mapping, False
+                )
+                new_expressions[name] = expr
```
knassre-bodo, line 2329: This is because after these changes, it is possible for collation terms to be non-references (e.g. a function call got pushed into the ordering term of the root)

hadia206 (approved):
Looks good to me.
Thanks Kian

pydough/relational/relational_nodes/abstract_node.py
```diff
@@ -6,7 +6,7 @@
 """
 
 from abc import ABC, abstractmethod
-from collections.abc import MutableMapping, MutableSequence
+from collections.abc import MutableMapping
```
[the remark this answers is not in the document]
knassre-bodo: The imports are auto managed by pre-commit so this means it is still used `MutableMapping` somewhere, but while I'm here I might as well purge the rest of them.

pydough/relational/rel_util.py
```diff
@@ -301,3 +317,36 @@ def transpose_expression(
...
             )
+
+
+def add_expr_uses(
+    expr: RelationalExpression,
+    n_uses: defaultdict[RelationalExpression, int],
+    top_level: bool,
+) -> None:
+    """
+    Count the number of times nontrivial expressions are used in an expression
```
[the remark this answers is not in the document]
knassre-bodo: Will add it. In this case, its a pretty arbitrary rule to avoid complicated duplicate terms.

pydough/conversion/relational_converter.py
```diff
@@ -1161,6 +1162,11 @@ def optimize_relational_tree(root: RelationalRoot) -> RelationalRoot:
     # Step 2: prune unused columns
     root = ColumnPruner().prune_unused_columns(root)
 
+    # Step 3: merge adjacent projections, when clearcut
```
vineetg3, line 1165: I think its good to define what clearcut here means for my understanding and future purposes.

vineetg3 (approved):
LGTM! The comments in the merge steps were very descriptive and that helped.

pydough/conversion/relational_converter.py
knassre-bodo: Will do. The definition is a bit vague, as it will avoid merging projects when the result would have a lot of duplicate subexpressions.

PR #317, opened 2025-04-01
Add verification for has/hasnot to ensure input is a collection

### LINKED ISSUE #316: Ensure HAS/HASNOT have a proper error message when the argument is not a collection
If `HAS` or `HASNOT` is called on a non-collection, should raise an exception explaining as such. This should be done by adding a proper type verifier for those two operators. The new verifier should be tested accordingly.

knassre-bodo:
Resolves #316 . Ensures the argument is a collection, and provides a clean error message if it isn't.


pydough/pydough_operators/expression_operators/registered_expression_operators.py
```diff
@@ -148,9 +149,11 @@
     "SIGN", False, RequireNumArgs(1), ConstantType(Int64Type())
 )
 COUNT = ExpressionFunctionOperator("COUNT", True, AllowAny(), ConstantType(Int64Type()))
-HAS = ExpressionFunctionOperator("HAS", True, AllowAny(), ConstantType(BooleanType()))
+HAS = ExpressionFunctionOperator(
+    "HAS", True, RequireCollection(), ConstantType(BooleanType())
+)
 HASNOT = ExpressionFunctionOperator(
-    "HASNOT", True, AllowAny(), ConstantType(BooleanType())
+    "HASNOT", True, RequireCollection(), ConstantType(BooleanType())
```
knassre-bodo, line 156: These verifier objects are assigned to each operator, and provide much higher quality error messages when the operands have an issue. Right now, most operators have weak/nonexistent verifiers, but eventually we will add proper verifiers for everything.

tests/test_qdag_collection_errors.py
```diff
@@ -89,6 +89,34 @@
...
+                FunctionInfo(
+                    "HAS",
+                    [
+                        ChildReferenceExpressionInfo("order_date", 0),
+                    ],
+                ),
+            ),
+            "Invalid operator invocation 'HAS(orders.order_date)': Expected a collection as an argument, received an expression",
+            id="has_on_expression",
+        ),
```
knassre-bodo, line 105: This test is the the equivalent of `Customers.WHERE(HAS(orders.order_date))`, and shows the error message we would receive in this case. The `XxxInfo` classes are helpers that exist for creating QDAG nodes (one of the IR datastructures). The `**` operator is used to chain these qdag nodes together when they are built (e.g. `X ** WhereInfo(...)` means build `X`, then build & return a `WHERE` clause with `X` as its predecessor)

vineetg3 (approved):
LGTM!

J-Solano-bodo-ai (approved):

PR #320, opened 2025-04-02
Dummy column added if no columns present 

### LINKED ISSUE #196: Optimize column pruning in the COUNT(*) scenario to avoid keeping extra column
To fix #187, a workable solution (#195) was found that fixed the bug by picking an arbitrary column (the first one, alphabetically) and keeping it to prevent every column from being pruned in the `COUNT(*)` scenario, which would cause bugs in the generated SQL. However, a more optimal solution is theoretically possible by inserting a dummy column, such as a null literal, which is kept instead of one of the real data columns.

vineetg3:
Resolves #196 . 
This PR keeps a NULL column if the subquery does not have any expressions it's selecting.


pydough/sqlglot/sqlglot_relational_visitor.py
```diff
@@ -308,6 +308,8 @@ def _build_subquery(
         """
         if sort:
             column_exprs = sorted(column_exprs, key=repr)
+        if len(column_exprs) == 0:
+            column_exprs = [sqlglot_convert((None,))]
```
knassre-bodo, line 312: Add a comment why this is needed & its effect (same for line 346).

knassre-bodo (approved):

tests/test_plan_refsols/correl_13.txt
```diff
@@ -1,10 +1,10 @@
 ROOT(columns=[('n', n)], orderings=[])
  PROJECT(columns={'n': agg_0})
   AGGREGATE(keys={}, aggregations={'agg_0': COUNT()})
-   FILTER(condition=DEFAULT_TO(agg_1, 0:int64) > 0:int64, columns={'account_balance': account_balance})
-    JOIN(conditions=[t0.key == t1.supplier_key], types=['left'], columns={'account_balance': t0.account_balance, 'agg_1': t1.agg_1})
-     FILTER(condition=nation_key <= 3:int64, columns={'account_balance': account_balance, 'key': key})
-      SCAN(table=tpch.SUPPLIER, columns={'account_balance': s_acctbal, 'key': s_suppkey, 'nation_key': s_nationkey})
+   FILTER(condition=DEFAULT_TO(agg_1, 0:int64) > 0:int64, columns={})
```
hadia206, line 4: For my understanding, a NULL column will not show up in the plans?

hadia206 (approved):
One question. Otherwise Looks good to me.
Thanks Vineet.

tests/test_plan_refsols/correl_13.txt
vineetg3: It's not a NULL column but an empty list of columns. We added a Null column in the SQLGlot AST.

PR #331, opened 2025-04-14
Add support for general join conditions

### LINKED ISSUE #330: Support general join conditions in PyDough
The user-facing metadata specification for these general join conditions may change when metadata v2 is implemented. The goals of this issue:
- Add a usable metadata syntax for general join conditions that works just like the simple join conditions, except instead of the keys mapping it uses a string that can be converted to PyDough text, where if the columns are referred to as `self.xxx` or `other.yyy` depending on if they come from the parent vs child collection. For example:
  - `condition: "LOWER(self.name) == LOWER(other.name)"`
  - `condition: "CONTAINS(self.name, other.name) | CONTAINS(other.name, self.name)"`
  - `condition: "(self.customer_key == other.customer_key) & (self.key != other.key)"`
- The metadata is handled the same as regular simple joins up until hybrid conversion. For regular parent->successor accesses, or aggregate parent->child accesses, it is still normal. But for parent->child accesses with aggregation, the same HybridPullUp trick must be done to convert into a parent->successor access then aggregate on the parent's uniqueness keys.
- The relational conversion will change slightly from normal simple joins.
- The new joins need to be heavily tested in relational conversion, sql conversion, and on e2e data, which means having a metadata knowledge graph that uses them.
  - Sub-task: fully flesh out the Amazon schema with synthetic data (just like defog) in a way that uses these general join conditions in the metadata alongside regular joins, then test e2e.
- The general joins must be properly documented.

knassre-bodo:
Resolves #330. See issue for more details. Also creates a new testing database in the knowledge graphs + sqlite databases called `Epoch` with the following characteristics:
- An events table containing various notable events from the past century & the moment they occurred.
- An eras table mapping each era from that period (e.g. WWI, Cold War, etc.)to a start/end year (latter exclusive) that can be joined onto the events via an interval condition.
- A times-of-day table mapping each time of day (e.g. morning, evening, etc.) from that period to a start/end hour (latter exclusive) that can be joined onto the events via an interval condition.
- A seasons table mapping each season to the 3 months that belong to it that can be joined onto the events via a disjunction (the month has to be any one of the three months).
- A searches table containing various query searches in search engines by various users. These can be joined onto the season/time-of-day with the same interval condition as the events, and onto the events in an N->N join on whether the search string contains the event name as a substring (case-insensitive).
- A users table containing various users who have search profiles and have made searches.

Some key limitations on general join conditions, for the time being:
- They cannot reference subcollections (e.g. `self.region.name == other.region.name` or `COUNT(self.orders_made) > other.threshold`)
- They cannot contain window functions


pydough/conversion/hybrid_tree.py
```diff
@@ -1563,13 +1686,45 @@ def populate_children(
         """
         self.stack.append(hybrid)
         for child_idx, child in enumerate(child_operator.children):
+            # Infer how the child is used by the current operation based on
+            # the expressions that the operator uses.
+            reference_types: set[ConnectionType] = set()
```
knassre-bodo, line 1691: Just switched around the order so this happens first in the loop body instead of at the end

documentation/metadata.md
```diff
@@ -10,6 +10,7 @@ This page document the exact format that the JSON files containing PyDough metad
 - [Properties](#properties)
```
knassre-bodo, line 10: This entire file will be drastically re-written in the immediate future when metadata V2 is released

pydough/conversion/hybrid_decorrelater.py
```diff
@@ -285,13 +285,18 @@ def decorrelate_child(
...
             child.subtree.agg_keys = new_agg_keys
         # If the child is such that we don't need to keep rows from the parent
         # without a match, replace the parent & its ancestors with a
         # HybridPullUp node (and replace any other deleted nodes with no-ops).
-        if child.connection_type.is_semi:
+        # This is done in-place, but only if the child is the first child of
+        # the parent.
+        if child.connection_type.is_semi and child_idx == min(
+            old_parent.correlated_children
+        ):
```
knassre-bodo, line 299: The body of this if statement is a **powerful** optimization that avoids duplicate subtrees, but also a tricky one with lots of edge cases. This extra condition bypasses the optimization under more complicated patterns.

pydough/metadata/properties/general_join_metadata.py
```diff
@@ -0,0 +1,203 @@
...
+        """
+        return self._self_name
+
+    @property
+    def other_name(self) -> str:
+        """
+        The name used to refer to columns from the child collection in the
+        condition string.
+        """
+        return self._other_name
```
knassre-bodo, line 83: When the property in the JSON file is created, these are `"self"` and `"other"` respectively. However, when `build_reverse_relationship` is called to build the flipside relationship, the self/other names are switched so that the `condition` string will work correctly.

pydough/conversion/hybrid_tree.py
```diff
@@ -2465,16 +2639,40 @@ def make_hybrid_tree(
...
+                            sub_property: SubcollectionRelationshipMetadata = (
+                                node.child_access.subcollection_property
                             )
+                            if isinstance(sub_property, SimpleJoinMetadata):
+                                # For a simple join access, populate the join
+                                # keys of the tree which will be bubbled
+                                # down throughout the entire child tree.
+                                join_key_exprs = HybridTranslator.get_join_keys(
+                                    parent,
+                                    node.child_access.subcollection_property,
```
hadia206, line 2657: why note use `sub_property` here?

pydough/conversion/hybrid_tree.py
```diff
@@ -2525,8 +2721,58 @@ def make_hybrid_tree(
...
+                            current_level = current_level.parent
+                        # Insert the calculate to access these correlated
+                        # keys, then add references to the new terms to
+                        # the agg keys.
+                        successor_hybrid.pipeline.append(
+                            HybridCalculate(
+                                successor_hybrid.pipeline[0], lhs_unique_keys, []
+                            )
+                        )
+                        for key_name, expr in lhs_unique_keys.items():
```
hadia206, line 2766: why is it renamed here?

pydough/conversion/relational_converter.py
```diff
@@ -271,12 +274,154 @@ def translate_expression(
...
+    def build_general_join_condition(
+        self,
+        condition: HybridExpr,
+        lhs_result: TranslationOutput,
+        rhs_result: TranslationOutput,
+        lhs_alias: str | None,
+        rhs_alias: str | None,
+    ) -> RelationalExpression:
+        """
+        Converts a the condition for a non-equijoin from a hybrid expression
```
hadia206, line 286:
```suggestion
        Converts the condition for a non-equijoin from a hybrid expression
```

pydough/metadata/parse.py
```diff
@@ -307,8 +307,8 @@ def add_dependency(property: PropertyKey, dependency: PropertyKey) -> None:
     # The set of all properties that are table columns
     table_columns: set[PropertyKey] = set()
 
-    # The set of all properties that are cartesian products
-    cartesian_products: set[PropertyKey] = set()
+    # The set of all properties that are cartesian products or general joins.
+    cartesian_or_general: set[PropertyKey] = set()
```
hadia206, line 311: For my understanding, why merge cartesian and general join in same? Will the properties always be the same for both?

pydough/unqualified/qualification.py
```diff
@@ -351,6 +355,192 @@ def qualify_collation(
...
+                            term = raw_term
+                            assert isinstance(term, (Reference, ColumnProperty))
+                            term_name = (
+                                term.term_name
+                                if isinstance(term, Reference)
+                                else term.column_property.name
+                            )
+                            return SidedReference(term_name, access, False)
+                raise PyDoughUnqualifiedException(
+                    "Accessing sub-collection terms is currently unsupported in PyDough general join conditions"
```
hadia206, line 524: Suggestion add `condition` statement to the error message
Same to all other errors.

pydough/unqualified/qualification.py
```diff
@@ -351,6 +355,192 @@ def qualify_collation(
...
+                return self.qualify_literal(condition)
+            case UnqualifiedBinaryOperation():
+                # For binary operators, invoke the same logic as for normal
+                # qualification of binary operators except with using
+                # `qualify_join_condition` on the inputs instead of
+                # `qualify_expression`.
+                operator = condition._parcel[0]
+                for _, op in BinOp.__members__.items():
+                    if operator == op.value:
+                        operation = op.name
```
hadia206, line 457: Any thoughts on making this more efficient? Repeating multiple times might slow things down if items are too large.

hadia206, line 457: Is it ok. to add `break` after finding the operation name. Do we need to keep looping after finding it?

pydough/unqualified/qualification.py
```diff
@@ -351,6 +355,192 @@ def qualify_collation(
...
+                # For accesses, verify that the access is in the form of
+                # `root.xxx.expr_name` where `xxx` is either the "self name" or
+                # the "other name" of the general join condition, and generate
+                # an appropriate sided reference indicating whether the term
+                # comes from the parent side or the child side. Also verify
+                # that the term is an expression.
+                if isinstance(condition._parcel[0], UnqualifiedAccess):
+                    predecessor: UnqualifiedAccess = condition._parcel[0]
+                    if isinstance(predecessor._parcel[0], UnqualifiedRoot):
+                        if predecessor._parcel[1] == self_name:
```
hadia206, line 493: The code looks very similar in this if block and the next one.
Only change is in `raw_term` and `True/False` pass at the end. Can we merge and clean it up?

hadia206: A thought
```Python
is_self = predecessor._parcel[1] == self_name
    raw_term = (
        access.ancestor_context.get_term(condition._parcel[1])
        if is_self
        else access.get_term(condition._parcel[1])
    )
....
 return SidedReference(term_name, access, is_self)
```

hadia206 (approved):
Well done Kian.
I skimmed over test_metadata folders and refsols ones.
Let me know if you want me review those in depth

pydough/unqualified/qualification.py
```diff
@@ -351,6 +355,192 @@ def qualify_collation(
...
+        sub-collection access.
+
+        Args:
+            `access`: the sub-collection access containing the join condition
+            that is being parsed.
+        """
+        assert isinstance(access.subcollection_property, GeneralJoinMetadata)
+        property: GeneralJoinMetadata = access.subcollection_property
+        self_name: str = property.self_name
+        other_name: str = property.other_name
```
vineetg3, line 372: I assume that when there are reverse relationships, the join condition has `self` and `other` reversed right?

pydough/qdag/expressions/sided_reference.py
```diff
@@ -0,0 +1,96 @@
...
+    The QDAG node implementation class representing a reference to a term from
+    one of two sides of connection between a parent collection and a child
+    connection. This type of expression node is only allowed to exist inside of
+    a general join condition.
+    """
+
+    def __init__(
+        self, term_name: str, collection: PyDoughCollectionQDAG, is_parent: bool
+    ):
+        self._collection: PyDoughCollectionQDAG = collection
```
vineetg3, line 28: Nit: Docstring for init

pydough/unqualified/qualification.py
vineetg3: Discard this, saw the other comment.

pydough/qdag/expressions/sided_reference.py
knassre-bodo: The convention is usually to do a class-level docstring (right above) instead of one for the constructor.

pydough/unqualified/qualification.py
knassre-bodo: Yeah we can do a version of this, but we still need to check if it is `self_name` or `other_name` because if it is neither then we have a problem.

pydough/unqualified/qualification.py
knassre-bodo: Sure, while I'm at it I should probably move this to be a method of the enum class itself.

knassre-bodo: This is just the list of all binary operations:
Screenshot 2025-04-24 at 12 00 48 AM
Anything with dictionaries etc. is probably just as slow due to constant offsets.

pydough/unqualified/qualification.py
knassre-bodo: It already is; the try-except in the calling site wraps the exception to include the _original_ condition in its entire form.

pydough/conversion/hybrid_decorrelater.py
```diff
@@ -436,18 +441,18 @@ def decorrelate_hybrid_tree(self, hybrid: HybridTree) -> HybridTree:
...
-            child.subtree = self.decorrelate_hybrid_tree(child.subtree)
         # Iterate across all the children, identify any that are correlated,
         # and transform any of the correlated ones that require decorrelation
         # due to the type of connection.
-        child_idx: int = 0
-        while child_idx < len(hybrid.children):
+        child_idx: int = len(hybrid.children) - 1
+        original_parent: HybridTree
+        if len(hybrid.correlated_children) > 0:
+            original_parent = copy.deepcopy(hybrid)
```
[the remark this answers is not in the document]
knassre-bodo: The point is we don't need `orignal_parent` if the length is 0 because it is used inside a loop that only does anything to correlated children.

pydough/conversion/hybrid_decorrelater.py
```diff
@@ -436,18 +439,18 @@ def decorrelate_hybrid_tree(self, hybrid: HybridTree) -> HybridTree:
         if hybrid.parent is not None:
             hybrid._parent = self.decorrelate_hybrid_tree(hybrid.parent)
             hybrid._parent._successor = hybrid
-        # Iterate across all the children and recursively decorrelate them.
-        for child in hybrid.children:
-            child.subtree = self.decorrelate_hybrid_tree(child.subtree)
```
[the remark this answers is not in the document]
knassre-bodo: Will do. The rationale is rather messy :(

pydough/conversion/hybrid_tree.py
knassre-bodo: Good point here, this should actually use the original not the renamed version.

Context: the pipeline within each level of the hybrid tree has a renaming dictionary for each step since each operator can cause certain names to be overridden, so we need to keep the original value for certain operations (like joins) and the new value (which is the one that gets used whenever the name is references) and need to distinguish the two. The way this is implemented is whenever a property is added (e.g. via calculate in the pipeline) and its name matches a name already being used, it instead gets added with a dummy name and the original name gets mapped to the dummy name in `renamings`. That way, if we ever need the original, we use the original name, and if we need the _most recent_ version we use the renaming.

pydough/metadata/parse.py
knassre-bodo: They are placed in the same set because the way this dependency-tracking algorithm works does not need to make any distinction between cartesian/general joins (a quirk of the soon-to-be removed "compound" properties)

pydough/unqualified/qualification.py
```diff
@@ -355,6 +354,176 @@ def qualify_collation(
...
+                return self.builder.build_expression_function_call(
+                    operation, qualified_operands
+                )
+            case UnqualifiedAccess():
+                # For accesses, verify that the access is in the form of
+                # `root.xxx.expr_name` where `xxx` is either the "self name" or
+                # the "other name" of the general join condition, and generate
+                # an appropriate sided reference indicating whether the term
+                # comes from the parent side or the child side. Also verify
+                # that the term is an expression.
```
vineetg3, line 484: What does "coming" from parent/child side mean?

PR #333, opened 2025-04-16
Add Quarter based date time support to PyDough

### LINKED ISSUE #322: Add support for QUARTER-based datetime manipulations
Goal: add the following functionality to PyDough that deal with datetime quarters.

- `QUARTER(X)` extracts what quarter of the year `X` belongs to (just like the `YEAR` or `MONTH` functions), as an integer between 1 and 4. By default this can just generate a case statement that switches on `MONTH(X).
- `DATETIME(X, 'start of quarter')` (& similar aliases) should truncate to the start of the quarter (jan-mar to jan 1st, apr-jun to apr 1st, etc.). Can implement by truncating to the month, then subtracting a number of months equal to `(MONTH(X) - 1) % 3`.
- `DATETIME(X, '±n quarters)` (& similar aliases) should be the same as `DATETIME(X, '±3n months)` (like how `"±n weeks"` is expanded to `"±7n days"`.
- `DATEDIFF("quarters", x, y)` (& similar aliases) should return the number of quarters between  `x` and `y`. Can do the same kind of trick as `DATEDIFF` with `months`: `4 * DATEDIFF("years", x, y) + QUARTER(y) - QUARTER(x)`.

These feature need to be documented in the [function list documentation](https://github.com/bodo-ai/PyDough/blob/main/documentation/functions.md).

vineetg3:
Resolves #322. Please check the issue for more details.


knassre-bodo (approved):
LGTM, just a few super-minor nitpicks

PR #335, opened 2025-04-19
Implemented Var and Std aggregation functions

### LINKED ISSUE #220: Add support for VAR/STD functions in PyDough, with keyword arguments that change with operator is used
Goal: add the following functionalities to PyDough, which should be tested e2e (see `test_pipeline.py` for examples):

- `VAR(x, type="population")` returns the population variance of `x`
- `VAR(x, type="sample")` returns the sample variance of `x`
- `STD(x, type="population")` returns the population population standard deviation of `x`
- `STD(x, type="sample")` returns the sample standard deviation of `x`

This should involve creating new aggregation operators for `VAR` and `STD`, but allowing multiple different operators to share the same name and choosing one of them based on keyword arguments. New operator classes & special handling in the UnqualifiedNode step will need to be added. If the `type` is not specified, the default for both of these functions should be "population".

What this should look like when the operators are defined (with one new operator class defined for `VARIANCE`, and a different one created for the implementing operators):
```
VARIANCE = KeywordBranchingExpressionFunctionOperator("VARIANCE", True, <verifier>, <deducer>, kwarg_defaults={"type": "population"})
POPULATION_VARIANCE = VARIANCE.with_kwarg({"type": "population"})
SAMPLE_VARIANCE = VARIANCE.with_kwarg({"type": "sample"})
```

The idea will be that whenever `builtin_registered_operators` is used, the sample/population versions are ignored (in the same way that binary operators are ignored in the definition of `UnqualifiedRoot`). Instead, they will be stored inside of a list of implementing operators inside `VARIANCE` (populated when `with_kwarg` is called). Whenever an invocation of `VARIANCE` is called (see `__call__` of `UnqualifiedOperator`), special logic should be done if the operator is an instance of `KeywordBranchingExpressionFunctionOperator`. The logic should iterate across the list of implementations & choose the one that matches the keyword arguments (throwing an exception if none of them match). From that point onward, `VARIANCE` will not be used as it should just be `POPULATION_VARIANCE` or `SAMPLE_VARIANCE` (which get cased on in the transform bindings).

All of these features need to be documented in the [function list documentation](https://github.com/bodo-ai/PyDough/blob/main/documentation/functions.md)

vineetg3:
Resolves #220.
This PR primarily does two things:
- It allows support for Functions to have keyword arguments( `VAR`, `STD`) by creating a class `KeywordBranchingExpressionFunctionOperator`, which allows us to create different `ExpressionFunctionOperator`s based on keywords by calling the `with_kwarg` function of the `KeywordBranchingExpressionFunctionOperator` class. 
- In `UnqualifiedOperator` class, we check if the operator is of type `KeywordBranchingExpressionFunctionOperator`, and if found, we find the concrete `ExpressionFunctionOperator` by comaring the kwargs the user provided with the kwargs defined internally, via `find_matching_implementation`.

This way we support sample variance, sample std, population variance, population std.


documentation/functions.md
```diff
@@ -886,6 +888,44 @@ The `HASNOT` function is called on a sub-collection and returns `True` if no rec
...
+<!-- TOC --><a name="var"></a>
+
+### VAR
+
+The `VAR` function returns the variance of the set of numerical values it is called on. This operation supports the `type` keyword argument, which can be used to specify the type of variance to compute. The following variances are supported:
+
+- `"population"`: The population variance.
+- `"sample"`: The sample variance.
+
+By default, the `VAR` function computes the population variance.
```
knassre-bodo, line 900: NIT: delete line 900 and change to this:
```
- `"population"` (default value): The population variance.
- `"sample"`: The sample variance.
```

documentation/functions.md
```diff
@@ -886,6 +888,44 @@ The `HASNOT` function is called on a sub-collection and returns `True` if no rec
...
+<!-- TOC --><a name="std"></a>
+
+### STD
+
+The `STD` function returns the standard deviation of the set of numerical values it is called on. This operation supports the `type` keyword argument, which can be used to specify the type of standard deviation to compute. The following standard deviations are supported:
+
+- `"population"`: The population standard deviation.
+- `"sample"`: The sample standard deviation.
+
+By default, the `STD` function computes the population standard deviation.
```
knassre-bodo, line 919: Same here

pydough/pydough_operators/expression_operators/keyword_branching_operators.py
```diff
@@ -0,0 +1,126 @@
...
+        # Create a new operator with same properties but different name
+        impl_name: str = (
+            f"{self.function_name}_{self._get_suffix_from_kwargs(kwargs)}".upper()
+        )
+        impl: ExpressionFunctionOperator = ExpressionFunctionOperator(
+            impl_name,
+            self.is_aggregation,
+            self.verifier,
+            self.deducer,
+        )
```
knassre-bodo, line 78: We should create a new subclass of `ExpressionFunctionOperator` just for this. My reason is that in the definition of `builtin_registered_operators`, we should make it so that objects of this new type are **NOT** included in the returned dictionary => ensuring that the user cannot just type `SAMPLE_STD` and have it work.

Also, can you include a test in `test_qualification_errors.py` to make sure that writing out `SAMPLE_STD` (and the others) as a function is NOT allowed?

pydough/unqualified/unqualified_node.py
```diff
@@ -537,7 +568,7 @@ def __call__(self, *args, **kwargs):
                 per,
                 kwargs,
             )
-        return UnqualifiedOperation(self._parcel[0], operands)
+        return UnqualifiedOperation(func_str, operands)
```
knassre-bodo, line 571: What I said in our earlier conversations is that we should change `UnqualifiedOperation` so it stops taking in the `func_str` and instead just takes in the operator value directly (instead of relying on the AST node builder to lookup the operator based on the string).

pydough/unqualified/unqualified_node.py
```diff
@@ -520,10 +519,42 @@ def __call__(self, *args, **kwargs):
...
                 is_window = False
-                if len(kwargs) > 0:
-                    raise PyDoughUnqualifiedException(
-                        f"PyDough function call {func} does not support keyword arguments at this time"
-                    )
+                # Get all registered operators
+                all_operators = pydop.builtin_registered_operators()
+                operator = all_operators.get(func)
+
+                if operator is not None:
```
knassre-bodo, line 526: Instead of this, which will regenerate the entire dictionary every time, let's add a new function to the `pydop` module in `operator_registry.py` which will take in the name of the operator, the kwargs, and do all of this & the stuff below, returning the operator in question. Instead of having it call `builtin_registered_operators()`, have it replicate the same idea with calling `inspect.getmembers` to find the single operator with the specified name that is also in `__all__`.

tests/test_documentation.py
```diff
@@ -14,7 +14,14 @@ def test_function_list():
...
     """
     # Identify every function that should be documented
-    special_names = {"NOT", "SLICE"}
+    special_names = {
+        "NOT",
+        "SLICE",
+        "POPULATION_STD",
+        "POPULATION_VARIANCE",
+        "SAMPLE_STD",
+        "SAMPLE_VARIANCE",
```
knassre-bodo, line 23: My earlier comment about ensuring that these aren't legal names will mean these should be removed because they won't be in the registry.

tests/simple_pydough_functions.py
```diff
@@ -1763,3 +1763,15 @@ def odate_and_rdate_avggap():
...
+
+def simple_var_std():
+    return Nations.WHERE(ISIN(name, ("ALGERIA", "ARGENTINA"))).CALCULATE(
+        name,
+        var=VAR(suppliers.account_balance),
+        std=STD(suppliers.account_balance),
+        sample_var=VAR(suppliers.account_balance, type="sample"),
+        sample_std=STD(suppliers.account_balance, type="sample"),
+        pop_var=VAR(suppliers.account_balance, type="population"),
+        pop_std=STD(suppliers.account_balance, type="population"),
```
knassre-bodo, line 1776: We also need to test a case where there are 0, 1 and 2 non-null records that are being aggregated for all 4 variants of var/std. Lets create another test that does this by doing a global aggregation of some data that gets partially/completely null-ed out by `KEEP_IF`

knassre-bodo (commented):
Hey @vineetg3, great job so far with this! Some further things to refine, but overall in the right direction.

tests/simple_pydough_functions.py
```diff
@@ -2167,3 +2167,47 @@ def odate_and_rdate_avggap():
...
+    )
+
+
+def simple_var_std_with_nulls():
+    first_customers = Customers.WHERE(ISIN(key, (1, 2, 3)))
+    return TPCH.CALCULATE(
+        var_samp_0_nnull=VAR(
+            KEEP_IF(first_customers.acctbal, first_customers.key > 3), type="sample"
+        ),
```
knassre-bodo, line 2189: Minor nit: you can simplify this a lot if you put the modified columns in a calculate of `first_customers`:
```
first_customers = Customers.WHERE(ISIN(key, (1, 2, 3))).CALCULATE(
  key_0=KEEP_IF(acctbal, key > 3),
  key_1=KEEP_IF(acctbal, key > 2),
  key_2=KEEP_IF(acctbal, key > 1),
)
return TPCH.CALCULATE(
  var_samp_0_nnull=VAR(first_customers.key_0, type="sample"),
  ...
)
```

knassre-bodo (approved):
One comment, otherwise LGTM :)

PR #336, opened 2025-04-21
Create documentation for internal workings of PyDough

knassre-bodo:
Includes documentation for the various IRs and conversions between them.

### DIFF
--- a/documentation/ir.md
+++ b/documentation/ir.md
@@ -1,0 +1,86 @@
+# PyDough Internal Representations Guide
+
+This document describes the various IRs used by PyDough to convert raw PyDough code into SQL text, as well as the conversion processes between them.
+
+<!-- TOC start (generated with https://github.com/derlin/bitdowntoc) -->
+
+- [Overview](#overview)
+- [Unqualified Nodes](#unqualified-nodes)
+  - [Unqualified Transform](#unqualified-transform)
+- [QDAG Nodes](#qdag-nodes)
+  - [Qualification](#qualification)
+- [Hybrid Tree](#hybrid-tree)
+  - [Hybrid Conversion](#hybrid-conversion)
+  - [Hybrid Decorrelation](#hybrid-decorrelation)
+- [Relational Tree](#relational-tree)
+  - [Relational Conversion](#relational-conversion)
+  - [Relational Optimization](#relational-conversion)
+- [SQLGlot AST](#sqlglot-ast)
+  - [SQLGlot Conversion](#sqlglot-conversion)
+
+<!-- TOC end -->
+
+<!-- TOC --><a name="overview"></a>
+## Overview
+
+TODO
+
+<!-- TOC --><a name="unqualified-nodes"></a>
+## Unqualified Nodes
+
+TODO
+
+<!-- TOC --><a name="unqualified-transform"></a>
+### Unqualified Transform
+
+TODO
+
+<!-- TOC --><a name="qdag-nodes"></a>
+## QDAG Nodes
+
+TODO
+
+<!-- TOC --><a name="qualification"></a>
+### Qualification
+
+TODO
+
+<!-- TOC --><a name="hybrid-tree"></a>
+## Hybrid Tree
+
+TODO
+
+<!-- TOC --><a name="hybrid-conversion"></a>
+### Hybrid Conversion
+
+TODO
+
+<!-- TOC --><a name="hybrid-decorrelation"></a>
+### Hybrid Decorrelation
+
+TODO
+
+<!-- TOC --><a name="relational-tree"></a>
+## Relational Tree
+
+TODO
+
+<!-- TOC --><a name="relational-conversion"></a>
+### Relational Conversion
+
+TODO
+
+<!-- TOC --><a name="relational-optimization"></a>
+### Relational Optimization
+
+TODO
+
+<!-- TOC --><a name="sqlglot-ast"></a>
+## SQLGlot AST
+
+TODO
+
+<!-- TOC --><a name="sqlglot-conversion"></a>
+### SQLGlot Conversion
+
+TODO


documentation/ir.md
```diff
@@ -0,0 +1,694 @@
...
+- `H2` has two children (`$0` and `$1`) and its pipeline has 3 operations:
+  - An access to the nations collection (how the step-down from `H1` to `H2` begins)
+  - A calculate that defines `region_name`, `nation_name`, and `n_customers_in_debt`
+  - An order-by that sorts by `nation_name` in ascending order.
+- `$0` is a singular access, meaning the data from `H2` and `$0` can be directly joined without needing to worry about changes in cardinality. They are joined on the condition that the `region_key` term from `H2` equals the `key` term from the bottom subtree of `$0` (which is `H3`).
+  - The contents of `$0` is just a single tree level `H3` which does not have any children and has a pipeline containing only an access to the `regions` collection.
+- `$1` is an aggregation access, meaning the data from `$1` must first be aggregated before it is joined with `H2`. The aggregation is done by grouping on the `nation_key` field of the bottom subtree of `$1` (which is `H5`) and computes the term `agg_0` as `COUNT()`. Then, the result is joined with `H2` on the condition that the `key` term from `H2` equals the `nation_key` field just used to aggregate.
```
hadia206, line 282: Suggestion to make $0 and $1 points have sublists describing steps rather than paragraph.
E.g.:
$1 is an aggregation access over customers.orders, with:
Grouped by nation_key and joined to H2 on key = nation_key
COUNT() aggregation over the orders collection
Join on condition ....

documentation/ir.md
```diff
@@ -0,0 +1,694 @@
...
+  - The contents of `$1` is two tree levels `H4` and `H5`. `H4` does not have any children and has a pipeline containing an access to the `customers` collection followed by a filter on the condition that `acctbal < 0`. `H5` does not have any children and only contains a single operation accessing the `orders` sub-collection of the parent level.
```
hadia206, line 283:
```suggestion
  - The contents of `$1` is two levels `H4` and `H5`. `H4` does not have any children and has a pipeline containing an access to the `customers` collection followed by a filter on the condition that `acctbal < 0`. `H5` does not have any children and only contains a single operation accessing the `orders` sub-collection of the parent level.
```

documentation/ir.md
```diff
@@ -0,0 +1,694 @@
...
+### Hybrid Conversion
+
+> [!IMPORTANT]
+> TODO: FINISH THIS SECTION.
+
+
+<!-- TOC --><a name="hybrid-decorrelation"></a>
+### Hybrid Decorrelation
+
+To understand why de-correlation matters, first consider the slightly more complex PyDough code example below. This PyDough code finds, for each nation in Europe, the total purchase quantity made by customers in that nation from suppliers in the same nation. 5 nations with the largest number of orders made by customers in that nation in the building market segment where the total price of the order is at least double the average of the total prices of **all** orders.
```
hadia206, line 295:
```suggestion
To understand why de-correlation matters, first consider the slightly more complex PyDough code example below. This PyDough code finds, for each nation in Europe, the total quantity of purchases made by customers from suppliers in the same nation. 5 nations with the largest number of orders made by customers in that nation in the building market segment where the total price of the order is at least double the average of the total prices of **all** orders.
```

hadia206: Is this part of the example? I don't see top 5 in the code
> 5 nations with the largest number of orders made by customers in that nation in the building market segment where the total price of the order is at least double the average of the total prices of **all** orders.

documentation/ir.md
```diff
@@ -0,0 +1,694 @@
...
+    L1 --> R1
+    R1 --> L2
+    L2 -.->|Aggregation| C2
+    L4 --> L5
+    L5 --> L6
+    L6 -.->|Singular| C4
+    Q1 --> Q2
+```
+
+Notice that `H6` contains a filter condition `$0.name == CORREL(nation_name)`. This means that even though `H6` is inside child `$0` of `H3` (not to be confused with child `$0` of `H6`), it still references the `nation_name` field from `H3`. This is a called a correlated reference, because it means that child `$0` of `H3` requires information from `H3` in order to be derived, but the whole point is that the data from child `$0` is calculated then joined onto he data from `H3`, so there is a catch-22.
```
hadia206, line 385: This was confusing to me. Here's a suggestion to edit. Feel free to update
```suggestion
Notice that `H6` contains a filter condition `$0.name == CORREL(nation_name)`, indicating a correlated reference. Although `H6` is nested within child `$0` of `H3`, it still references the `nation_name` field from `H3`. This is called a correlated reference — child `$0` of `H3` depends on `H3`'s data (`nation_name`) to compute its result. But since this data is meant to be computed before joining with `H3`, this creates a catch-22.
```

documentation/ir.md
```diff
@@ -0,0 +1,694 @@
...
...rocedure. This will recursively traverse the entire tree and look for hybrid nodes that have a correlated child (where the child type is NOT semi/anti join, since those two do allow correlated references). This procedure will reach tree `H3` and see that its child `$0` is correlated and is an `AGGREGATION` connection, and will therefore de-correlate it. It does so by copying the entire hybrid tree so far (`H1`, `H2`, and the first two operators from `H3`) and attaching them to the top of child `$0`, above `H4`. This now means that the `CORREL(nation_name)` term in `H6` can be rephrased into a back-reference to the `nation_name` field in the copied version of `H3` (which is 3 levels above `H6`). This does change the join/aggregation keys when connecting H3 to child `$0`, since now we must connect each unique record of `H3` to the corresponding prepended section of `$0`. This is done by changing the join keys into the uniqueness keys of `H1`/`H2`/`H3`, which is aka `key` from `regions` (`BACK(1).key` from the perspective of `H3` and `BACK(4).key` from the perspective of `H6`) and `key` from `nations` (`key` from the perspective of `H3` and `BACK(3).key` from the perspective of `H6`).
```
hadia206, line 387:
```suggestion
To fix this, we next run the de-correlation procedure. This procedure recursively traverses the entire tree, looking for hybrid nodes with correlated children (excluding semi and anti joins, which do allow correlated references). This procedure will reach tree `H3` and see that its child `$0` is correlated and is an `AGGREGATION` connection, and will therefore de-correlate it. To do this, it copies the hybrid tree built so far (`H1`, `H2`, and the first two operators of `H3`) and attaches them above child `$0`, specifically before `H4`. This now means that the `CORREL(nation_name)` term in `H6` can be rephrased into a back-reference to the `nation_name` field in the copied version of `H3` (which is 3 levels above `H6`). This changes the join and aggregation keys used when connecting `H3` to child `$0`, as each unique record from `H3` must now align with the prepended section of `$0`. This is done by changing the join keys into the uniqueness keys of `H1`/`H2`/`H3`, the `key` from `regions` (accessed as `BACK(1).key` from`H3` and `BACK(4).key` from `H6`) and the `key` from `nations` (`key` from  `H3` and `BACK(3).key` from `H6`).
```

documentation/ir.md
```diff
@@ -0,0 +1,694 @@
...
+    R1' --> L2'
+    L2' -->L4
+    L2 -.->|Aggregation| C2
+    L4 --> L5
+    L5 --> L6
+    L6 -.->|Singular| C4
+    Q1 --> Q2
+```
+
+However, we can go one step further if we make an observation about this hybrid tree. Notice how now, we are computing the logic of `H1`, `H2`, and most of `H3` twice. This must happen if we intend to keep *every* nation, including the ones without any entries of `$0`. However, if we don't intend to keep nations where there are no records of `$0` to join onto, we can modify the original PyDough code as follows:
```
hadia206, line 492:
```suggestion
However, we can go one step further if we make an observation about this hybrid tree. Notice how now, we are computing the logic of `H1`, `H2`, and most of `H3` twice. This must happen if we intend to keep *every* nation, including the ones without any entries of `$0`. However, if we don't intend to keep nations that lack corresponding `$0` records to join with, we can modify the original PyDough code as follows:
```

documentation/ir.md
```diff
@@ -0,0 +1,694 @@
...
+).nations.CALCULATE(nation_name=name)
+domestic_lines = customers.orders.lines.WHERE(supplier.nation.name == nation_name)
+nation_domestic_purchase_info = selected_nations
+.WHERE(HAS(domestic_lines)).CALCULATE(
+    nation_name,
+    domestic_quantity=SUM(domestic_lines.quantity),
+)
+```
+
+With this modification, the original hybrid tree changes so the access to child `$0` of `H3` is now an `AGGREGATION_ONLY_MATCH` access, meaning we do an inner join instead of a left join. The de-correlation procedure will notice this and, after transforming child `$0` into its de-correlated form, realize it can do an optimization to delete the original `H1`, `H2` and prefix of `H3` since all of the data required from them is accessible inside `$0`. This is done via the hybrid "pull up" node, which specifies that the data comes from the specified child, instead of having existing data that gets joined with data from the child. This will look as follows:
```
hadia206, line 506:
```suggestion
With this modification, the original hybrid tree is modified so that access to child `$0` of `H3` is now an `AGGREGATION_ONLY_MATCH` access, meaning we do an inner join instead of a left join. The de-correlation procedure will notice this and, after transforming child `$0` into its de-correlated form, recognize an opportunity to optimize by removing the original `H1`, `H2` and prefix of `H3` since all of the data required from them is accessible inside `$0`. This is done via the hybrid "pull up" node, which specifies that the data comes from the specified child, instead of having existing data that gets joined with data from the child. This will look as follows:
```

documentation/ir.md
```diff
@@ -0,0 +1,694 @@
...
+    R1' --> L2'
+    L2' -->L4
+    L2 -.->|AggregationOnlyMatch| C2
+    L4 --> L5
+    L5 --> L6
+    L6 -.->|Singular| C4
+    Q1 --> Q2
+```
+
+The way to interpret this is that the entirety of child `$0` of H3 is evaluated, from `H1'` to `H6`, then grouped on `BACK(3).key` and `BACK(4).key` to calculate `agg_0` and `agg_1` (renaming the latter to `nation_name`), and that result is passed along from the PullUp node to the calculate in `H3` which builds on it.
```
hadia206, line 591:
```suggestion
This means the entire child `$0` subtree of `H3` is evaluated, from `H1'` to `H6`, then grouped on `BACK(3).key` and `BACK(4).key` to calculate `agg_0` and `agg_1` (renaming the latter to `nation_name`), and that result is passed along from the PullUp node to the calculate in `H3` which builds on it.
```

documentation/ir.md
```diff
@@ -0,0 +1,694 @@
...
+
+<!-- TOC --><a name="relational-tree"></a>
+## Relational Tree
+
+> [!IMPORTANT]
+> TODO: FINISH THIS SECTION.
+
+For an example of the relational tree, consider the `nation_info` example from earlier.
```
hadia206, line 599:
```suggestion
To illustrate the relational tree, consider the `nation_info` example from earlier.
```

documentation/ir.md
```diff
@@ -0,0 +1,694 @@
...
+```py
+nation_info = nations.CALCULATE(
+  region_name=region.name,
+  nation_name=name,
+  n_orders_from_debt_customers=COUNT(customers.WHERE(acctbal < 0).orders)
+).ORDER_BY(nation_name.ASC())
+```
+
+Using the hybrid tree from earlier, the following is how the relational tree created by relational conversion could look:
```
hadia206, line 608:
```suggestion
Using the hybrid tree from earlier, here’s how the relational tree created by the conversion would look:
```

documentation/ir.md
```diff
@@ -0,0 +1,694 @@
...
+3. Next, the QDAG nodes are run through a process called [hybrid conversion](#hybrid-conversion) which restructures the logic into the datastructure known as the [hybrid tree](#hybrid-tree) in order to better organize the types of ways different subtrees of the data are linked together.
+4. The hybrid tree is further transformed by the [decorrelation procedure](#hybrid-decorrelation) to remove correlated references created by the hybrid conversion process.
+5. The transformed hybrid tree is converted into a [relational tree](#relational-tree) highly reminiscent of the datastructure used to represent relational algebra in frameworks such as Apache Calcite. The conversion to this datastructure is called [relational conversion](#relational-conversion).
```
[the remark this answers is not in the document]
knassre-bodo: It is used more as a macrocosm example of what kind of datastructure it is, since this analogy would be more familiar to a lot of people with some database experience when explaining how PyDough works. But yeah I can link it.

documentation/ir.md
```diff
@@ -0,0 +1,694 @@
...
+  n_orders_from_debt_customers=COUNT(customers.WHERE(acctbal < 0).orders)
+).ORDER_BY(nation_name.ASC())
+```
+
+This has the following structure as QDAG nodes:
+
+```
+──┬─ TPCH
+  ├─── TableCollection[nations]
+  ├─┬─ Calculate[region_name=$1.name, nation_name=name, n_customers_in_debt=COUNT($2)]
```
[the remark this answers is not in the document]
knassre-bodo: Whoops, this should have been updated but wasn't.

documentation/ir.md
```diff
@@ -0,0 +1,694 @@
...
+> [!IMPORTANT]
+> TODO: FINISH THIS SECTION.
+
+For an example of the Hybrid Tree structure, consider the `nation_info` example earlier from the unqualified nodes:
+
+```py
+nation_info = nations.CALCULATE(
+  region_name=region.name,
+  nation_name=name,
+  n_customers_in_debt=COUNT(customers.WHERE(acctbal < 0).orders)
```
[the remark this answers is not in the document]
knassre-bodo: Yeah this is out of date and should be updated.

documentation/ir.md
knassre-bodo: Whoops, more holdover from an older version I forgot to delete.

PR #338, opened 2025-04-23
Add prefix frame cumulative window functions for sum/count/size/avg

### LINKED ISSUE #337: Add cumulative functionality to aggregation window functions
Goal: allow RELSUM, RELCOUNT, RELSIZE and RELAVG to have the functionality of `ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW` via two extra keyword arguments:
- `by`: same as other window functions (requires `cumulative=True` if provided)
- `cumulative`: if True, the window function is done with the specified frame instead of frame-less (requires `by` to be provided if True), default is False
These changes must be properly documented in the function list documentation.

knassre-bodo:
Resolves #337. Allows `RELSUM`, `RELAVG`, `RELCOUNT` and `RELSIZE` to take in a `by` argument and a `cumulative` argument (`by` is allowed only if `cumulative=True`, and `cumulative` can only be True if `by` is provided) specifying to do a prefix sum/avg/etc.


pydough/pydough_operators/expression_operators/expression_window_operators.py
```diff
@@ -25,24 +25,24 @@ def __init__(
         function_name: str,
         verifier: TypeVerifier,
         deducer: ExpressionTypeDeducer,
-        allows_order: bool = True,
```
knassre-bodo, line 28: Replaced this because there are no longer any window functions that do _not_ allow an order term in PyDough. If there are once again, we can add this term back.

documentation/functions.md
```diff
@@ -1046,6 +1048,10 @@ Customers.CALCULATE(ratio=acctbal / RELSUM(acctbal))
 # Finds the ratio between each customer's account balance and the sum of all
 # all customers' account balances within that nation.
 Nations.customers.CALCULATE(ratio=acctbal / RELSUM(acctbal, per="Nations"))
+
+# Finds, for each customer, the wealth of that customer combined with all
+# poorer customers.
+Customers.CALCULATE(cumulative_wealth=RELSUM(acctbal, by=acctbal.ASC(), cumulative=True))
```
vineetg3, line 1054: I think an another example combined with `per` argument would nail the power of this function even more.

knassre-bodo: The `per` argument is the same for all window functions, all of which have an example with `per`.

vineetg3 (approved):
LGTM!

PR #343, opened 2025-05-14
Using new `always matches` metadata feature to optimize plans

knassre-bodo:
Add utilization of the `always matches` property of the new metadata so any left joins that always match are promoted to inner joins, along with several other changes to relational plan logic to avoid regressions in other areas:
- Changing column pruning so the removal of redundant joins can also occur with inner joins (by passing around some extra information)
- Allowing merging prune/rename projects into aggregates, filters, limits

The net effects of these changes are:
- Some left joins become inner joins
- Some instances with 2 CTEs are merged into one
- Some WHERE conditions are merged into the JOIN condition
- Some instances of duplicate subquery computations are deleted
- Some COALESCE calls are removed because they are redundant


pydough/sqlglot/sqlglot_relational_visitor.py
```diff
@@ -489,7 +489,13 @@ def visit_aggregate(self, aggregate: Aggregate) -> None:
...
         if keys:
             if aggregations:
-                query = query.group_by(*sorted(keys, key=repr))
+                grouping_keys: list[SQLGlotExpression] = []
+                for key in sorted(keys, key=repr):
+                    while isinstance(key, SQLGlotAlias):
+                        key = key.this
+                    if key not in grouping_keys:
+                        grouping_keys.append(key)
+                query = query.group_by(*grouping_keys)
```
knassre-bodo, line 498: This ensures we don't have `X as Y` inside of a `GROUP BY` clause (invalid SQL syntax)

pydough/conversion/filter_pushdown.py
```diff
@@ -114,6 +114,10 @@ def push_filters(
                         remaining_filters,
                         lambda expr: only_references_columns(expr, input_cols[idx]),
                     )
+                # Ensure that if any filter is pushed into an input (besides
+                # the first input) that the join is marked as not prunable.
```
hadia206, line 118: Why excluding the first input?

pydough/conversion/hybrid_tree.py
```diff
@@ -1403,6 +1410,70 @@ def add_successor(self, successor: "HybridTree") -> None:
...
+        # there are any operations that could remove a row.
+        for operation in self.pipeline[1:]:
+            match operation:
+                case HybridCalculate() | HybridNoop() | HybridRoot():
+                    continue
+                case HybridFilter() | HybridLimit():
+                    return False
+                case operation:
+                    raise NotImplementedError(
+                        f"Invalid intermediary pipeline operation: {operation.__class__.__name__}"
```
hadia206, line 1472: For my understanding, do we have a list of what is considered an intermediary op?

pydough/conversion/hybrid_tree.py
```diff
@@ -1403,6 +1410,70 @@ def add_successor(self, successor: "HybridTree") -> None:
...
+        There is no need to check the children data (except for partitions &
+        pull-ups) since the only way a child could cause the current context
+        to reduce records is if there is a HAS/HASNOT somewhere, which
+        would mean there is a filter in the pipeline.
+        """
+        # Verify that the first operation in the pipeline guarantees a match
+        # with every record from the previous level (or parent context if it
+        # is the top level)
+        start_operation: HybridOperation = self.pipeline[0]
+        match start_operation:
```
hadia206, line 1430: Is there restrictions on what can be a start operation?
I don't see a case for some of the hybrid operations: e.g. `HybridRoot`, `HybridCalculate`, `HybridFilter`, ...etc
and see the default with error "invalid start of pipeline"

pydough/conversion/merge_projects.py
```diff
@@ -219,19 +222,28 @@ def merge_adjacent_projects(node: RelationalRoot | Project) -> RelationalNode:
                 # Otherwise, halt the merging process since it is no longer
                 # possible to merge the children of this project into it.
                 break
-    # Final round: if there is a project on top of a scan that only does
-    # column pruning/renaming, just push it into the scan.
+    # Final round: if there is a project on top of a scan/aggregate/filter that only does
```
hadia206, line 225: add limit?

pydough/conversion/relational_converter.py
```diff
@@ -670,6 +675,7 @@ def handle_children(
                             child.subtree.join_keys,
                             child.subtree.general_join_condition,
                             child_idx,
+                            True,
```
hadia206, line 678: Can you explain why `True` in this case and the below? but `False` otherwise.

tests/test_plan_refsols/access_partition_child_backref_calc.txt
```diff
@@ -1,6 +1,5 @@
 ROOT(columns=[('part_name', name), ('part_type', part_type_1), ('retail_price_versus_avg', retail_price - avg_price)], orderings=[])
  JOIN(conditions=[t0.part_type == t1.part_type], types=['inner'], columns={'avg_price': t0.avg_price, 'name': t1.name, 'part_type_1': t1.part_type, 'retail_price': t1.retail_price})
-  PROJECT(columns={'avg_price': agg_0, 'part_type': part_type})
-   AGGREGATE(keys={'part_type': part_type}, aggregations={'agg_0': AVG(retail_price)})
-    SCAN(table=tpch.PART, columns={'part_type': p_type, 'retail_price': p_retailprice})
+  AGGREGATE(keys={'part_type': part_type}, aggregations={'avg_price': AVG(retail_price)})
```
hadia206, line 3: What triggered moving the projection (`'avg_price': agg_0, `) to aggregation ? 
I can see how this is obvious and correct behavior but wonder what part of the code in this PR triggered it.

hadia206 (approved):
LGTM. Most comments are questions to understand.

pydough/conversion/hybrid_tree.py
knassre-bodo: No, but the key distinction here is that certain kinds of operators can be the _start_ of a pipeline, and others must build off of something.
- Can be the start: root, partition, collection access, child pull up, partition child
- Must build on something: calculate, filter, limit, noop

pydough/conversion/hybrid_tree.py
knassre-bodo: Calculate/Filter can't be the start (since they augment something else), but good point I should move root into this one (it doesn't come up so far because roots usually aren't present inside the children, and `always_exists` is only ever called on children of something else.

pydough/conversion/relational_converter.py
knassre-bodo: These kinds of joins are accessing child data and bringing it into the current context, so if the data is not being used (and the rhs does not filter the lhs), then we can just delete the rhs. The ones where it is False actually do have an effect even if we only use columnsf rom one side. E.g. if I just have `orders.customers`, this will return all the columns from `customers`, none of the data from `orders`, and will return each rows from `customers` once for each order that customer made (or delete the row if that customer never made any orders), so in this scenario I can't prune the join even if I don't use one side.

PR #346, opened 2025-05-19
Add support for sliding window function frames

### LINKED ISSUE #345: Add window function frame feature for relsum/relavg/relcount/relsize
Goal: build another path for window function keyword arguments where if `by` is provided, you can either have `cumulative=True`, or `frame=(lower,upper)` where the following are true:
- `lower` and `upper` are both either integer literals or None.
- Positive integer literals = `n following`
- Negative integer literals = `n preceding`
- Zero = `current row`
- None = `unbounded preceding` (for lower) or `unbounded following` for upper
- The frames must be valid, so `lower` represents an ordinal position that is `<=` the ordinal position of upper
So for example:
```py
nations.customers.orders.CALCULATE(
  w1=RELSUM(account_balance, by=order_date.ASC(), frame=(None, 0)),
  w2=RELSUM(account_balance, by=order_date.ASC(), frame=(0, None), per="customers"),
  w3=RELSUM(account_balance, by=order_date.ASC(), frame=(-3, 5)),
  w4=RELSUM(account_balance, by=order_date.ASC(), frame=(-10, -1), per="nations"),
)
```
In this example:
- `w1` is equivalent to `SUM(c_acctbal) OVER (ORDER BY o_orderdate ASC ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW)`, which is the same as `RELSUM(account_balance, by=order_date.ASC(), cumulative=True)`
- `w2` is equivalent to `SUM(c_acctbal) OVER (PARTITION BY c_custkey ORDER BY o_orderdate ASC ROWS BETWEEN CURRENT ROW AND UNBOUNDED FOLLOWING)`
- `w3` is equivalent to `SUM(c_acctbal) OVER (ORDER BY o_orderdate ASC ROWS BETWEEN 3 PRECEDING AND 5 FOLLOWING)`
- `w4` is equivalent to `SUM(c_acctbal) OVER (PARTITION BY n_nationkey ORDER BY o_orderdate ASC ROWS BETWEEN 10 PRECEDING AND 1 PRECEDING)`

This feature must be well documented in the function documentation. The implementation should closely track how `cumulative=True` was implemented in #388.

knassre-bodo:
Resolves #345. See issue for more details.  Additional changes:
- Changed the SQLGlot generation for intermediary `SELECT` clauses to sort the columns by the column **alias** (falling back on the entire column string if there isn't one) instead of just using the column string. This results in SQL that is sometimes easier to read when comparing to the original PyDough code.
- Minor change to how the SQLGlot AST is generated for the `RelationalRoot` to avoid an extra `SELECT` clause just to add a redundant `ORDER BY` on top of a sub-query that already sorts by the same values.
- Fixed the naming of the generated plan files for the EPOCH dataset to include the `epoch_` prefix


pydough/sqlglot/sqlglot_relational_visitor.py
```diff
@@ -45,6 +45,23 @@
...
+    Args:
+        expr (SQLGlotExpression): The expression to sort.
+
+    Returns:
+        str: The string representation of the expression.
+    """
+    if isinstance(expr, SQLGlotAlias):
+        return repr(expr.alias)
+    else:
+        return repr(expr)
```
knassre-bodo, line 62: Cosmetic effect: outside of the final `SELECT` clause, all other `SELECT` clauses are ordered alphabetically by **alias** (if they have one) instead of just by the entire expression's string representation.

tests/test_sql_refsols/tpch_q5_sqlite.sql
```diff
@@ -1,9 +1,9 @@
 WITH _t1 AS (
   SELECT
-    MAX(nation.n_name) AS agg_3,
     SUM(lineitem.l_extendedprice * (
       1 - lineitem.l_discount
-    )) AS agg_0
+    )) AS agg_0,
+    MAX(nation.n_name) AS agg_3
```
knassre-bodo, line 6: Example of the cosmetic change: before it was sorting by the string `MAX(...) AS agg_3` vs `SUM(...) AS agg_0`, now it is sorting by `agg_3` vs `agg_0`

pydough/sqlglot/sqlglot_relational_visitor.py
```diff
@@ -544,10 +561,14 @@ def visit_root(self, root: RelationalRoot) -> None:
...
-
         if self._is_mergeable_ordering(ordering_exprs, input_expr):
+            # The columns used for ordering must only be noted as dependencies
+            # if the root has an ordering but the input expression does not.
+            ordering_dependencies: set[Identifier] = set()
+            if len(ordering_exprs) > 0 and "order" not in input_expr.args:
+                ordering_dependencies = find_identifiers_in_list(ordering_exprs)
             query = self._merge_selects(
-                exprs, input_expr, find_identifiers_in_list(ordering_exprs), sort=False
+                exprs, input_expr, ordering_dependencies, sort=False
```
knassre-bodo, line 571: This change makes it so we no longer have final queries like this:
```sql
with _t0 AS (
SELECT ...
  FROM ...
  ORDER BY x, y, z
  LIMIT 5
)
SELECT ..
FROM _t0
ORDER BY x, y, z
```
Since they can be replaced with:
```sql
SELECT ...
FROM ...
ORDER BY x, y, z
LIMIT 5
```

This method  `_merge_selects` already performs the compaction discussed here, but not if any of the columns in the order by clause are NOT in the final select clause (since it is bad form to do `select a from ... order by c`). The third arguments to `_merge_selects` is used to pass in any columns like this.

The change here is we no longer pass in the columns used by the `order by` unless there is an `order by` in the root but NOT in the input. The effect here is if we get the pattern of double-orderby, we can now delete the 2nd one bc it is redundant (we don't have to worry about the order bys being different because that is what `_is_mergeable_ordering` checks for)

tests/test_pipeline_epoch.py
```diff
@@ -400,7 +400,8 @@ def test_pipeline_until_relational_epoch(
     Tests the conversion of the PyDough queries on the custom epoch dataset
     into relational plans.
     """
-    unqualified_impl, file_name, _ = pydough_pipeline_test_data_epoch
+    unqualified_impl, test_name, _ = pydough_pipeline_test_data_epoch
+    file_name: str = f"epoch_{test_name}"
```
knassre-bodo, line 404: This change was made for consistency & clarity; now all of the generated files for the epoch tests have the same naming scheme.

tests/test_sql_refsols/epoch_overlapping_event_searches_per_user_sqlite.sql
```diff
@@ -20,26 +20,16 @@ WITH _t3 AS (
...
-  SELECT
-    agg_2,
-    n_searches,
-    user_name
-  FROM _t1
-  ORDER BY
-    n_searches DESC,
-    agg_2
-  LIMIT 4
 )
```
knassre-bodo, line 33: This is an example of the cosmetic optimization created by the change to the call to `_merge_selects`

tests/test_sql_refsols/epoch_overlapping_event_searches_per_user_ansi.sql
```diff
@@ -16,26 +16,16 @@ WITH _t3 AS (
...
-), _t0 AS (
-  SELECT
-    agg_2,
-    n_searches,
-    user_name
-  FROM _t1
-  ORDER BY
-    n_searches DESC,
-    agg_2
-  LIMIT 4
```
knassre-bodo, line 38: This is another example of the cosmetic optimization created by the change to the call to _merge_selects

tests/test_sql_refsols/window_sliding_frame_relsize_ansi.sql
```diff
@@ -0,0 +1,16 @@
...
+  COUNT(*) OVER (ORDER BY sbtransaction.sbtxdatetime NULLS LAST, sbtransaction.sbtxid NULLS LAST ROWS BETWEEN UNBOUNDED PRECEDING AND 1 PRECEDING) AS w5,
+  COUNT(*) OVER (PARTITION BY sbcustomer.sbcustid ORDER BY sbtransaction.sbtxdatetime NULLS LAST, sbtransaction.sbtxid NULLS LAST ROWS BETWEEN UNBOUNDED PRECEDING AND 1 PRECEDING) AS w6,
+  COUNT(*) OVER (ORDER BY sbtransaction.sbtxdatetime NULLS LAST, sbtransaction.sbtxid NULLS LAST ROWS BETWEEN 3 PRECEDING AND 5 FOLLOWING) AS w7,
+  COUNT(*) OVER (PARTITION BY sbcustomer.sbcustid ORDER BY sbtransaction.sbtxdatetime NULLS LAST, sbtransaction.sbtxid NULLS LAST ROWS BETWEEN 3 PRECEDING AND 5 FOLLOWING) AS w8
+FROM main.sbcustomer AS sbcustomer
+JOIN main.sbtransaction AS sbtransaction
+  ON sbcustomer.sbcustid = sbtransaction.sbtxcustid
+ORDER BY
+  sbtransaction.sbtxdatetime
+LIMIT 8
```
knassre-bodo, line 16: Before the two cosmetic changes made in this PR, this SQL file was extremely nasty to read for two reasons:
- It had the same pattern of CTE with limit folowed by the redundant `SELECT ... FROM ... ORDER BY`
- Within the CTE, the columns were in a crazy order because they were being sorted lexigraphically by the entire string, which meant that whatever was in the `OVER (...)` was being used to sort, instead of `w1` vs `w2` vs ...

Now, in this version, it is much easier to look at each column in this select clause and compare it 1:1 with the expression in the PyDough code.

tests/test_sql_refsols/epoch_culture_events_info_ansi.sql
```diff
@@ -1,34 +1,22 @@
...
+  OR seasons.s_month3 = EXTRACT(MONTH FROM events.ev_dt)
+JOIN times AS times
+  ON times.t_end_hour > EXTRACT(HOUR FROM events.ev_dt)
+  AND times.t_start_hour <= EXTRACT(HOUR FROM events.ev_dt)
+WHERE
+  events.ev_typ = 'culture'
 ORDER BY
-  date_time
+  events.ev_dt
+LIMIT 6
```
knassre-bodo, line 22: This is another example of the cosmetic optimization created by the change to the call to _merge_selects

tests/test_sql_refsols/epoch_overlapping_event_search_other_users_per_user_ansi.sql
```diff
@@ -19,16 +19,6 @@ WITH _s1 AS (
...
-), _t0 AS (
-  SELECT
-    agg_2,
-    n_other_users,
-    user_name
-  FROM _t1
-  ORDER BY
-    n_other_users DESC,
-    agg_2
-  LIMIT 7
```
knassre-bodo, line 31: This is another example of the cosmetic optimization created by the change to the call to `_merge_selects`

pydough/sqlglot/sqlglot_relational_expression_visitor.py
```diff
@@ -71,6 +71,91 @@ def visit_call_expression(self, call_expression: CallExpression) -> None:
...
+        - None: `UNBOUNDED PRECEDING` (for lower) or `UNBOUNDED FOLLOWING` (for
+        upper)
+        - Zero: `CURRENT ROW`
+        - `+N`: `N FOLLOWING`
+        - `-N`: `N PRECEDING`
+
+        Args:
+            kwargs (dict[str, object]): The keyword arguments to parse, which
+            may include a `frame` argument or a `cumulative` argument. It is
+            assumed the keyword arguments are the correct types/formats.
```
knassre-bodo, line 94: This assumption is all of the checks in `unqualified_node.py` (verified in the new `bad_window` tests in `test_unqualified_node.py`

documentation/functions.md
```diff
@@ -1086,8 +1086,16 @@ The `RELSUM` function returns the sum of multiple rows of a singular expression
...
+- `by` (optional): 1+ collation values, either as a single expression or an iterable of expressions, used to order the records of the current context. Can only be provided if `cumulative` is True or `frame` is provided.
+- `cumulative` (optional): optional argument (default `False`) that can only be `True` if the `by` argument is provided. If `True`, then instead of returning the sum of all of the data in the context, returns the cumulative sum of all rows up to and including the current row when sorted according to the keys in the `by` argument. This argument cannot be provided if `frame` is also provided.
+- `frame` (optional): optional argument that, if provided, must be a tuple of two values `(lower, upper)` that are each either integer literals or None. If provided, it means the relative sum is performed on a subset of the selected records based on the values of `lower` and `upper` when sorted according to the keys in the `by` argument:
```
hadia206, line 1091: nit:
```suggestion
- `frame` (optional): optional argument that, if provided, must be a tuple of two values `(lower, upper)`, where each value is either an integer literal or None. If provided, it means the relative sum is performed on a subset of the selected records based on the values of `lower` and `upper` when sorted according to the keys in the `by` argument:
```

documentation/functions.md
```diff
@@ -1114,8 +1126,16 @@ The `RELAVG` function returns the average of multiple rows of a singular express
...
+- `by` (optional): 1+ collation values, either as a single expression or an iterable of expressions, used to order the records of the current context. Can only be provided if `cumulative` is True or `frame` is provided.
+- `cumulative` (optional): optional argument (default `False`) that can only be `True` if the `by` argument is provided. If `True`, then instead of returning the average of all of the data in the context, returns the cumulative average of all rows up to and including the current row when sorted according to the keys in the `by` argument. This argument cannot be provided if `frame` is also provided.
+- `frame` (optional): optional argument that, if provided, must be a tuple of two values `(lower, upper)` that are each either integer literals or None. If provided, it means the relative average is performed on a subset of the selected records based on the values of `lower` and `upper` when sorted according to the keys in the `by` argument:
```
hadia206, line 1131: Same
```suggestion
- `frame` (optional): optional argument that, if provided, must be a tuple of two values `(lower, upper)`, where each value is either an integer literal or None. If provided, it means the relative average is performed on a subset of the selected records based on the values of `lower` and `upper` when sorted according to the keys in the `by` argument:
```

documentation/functions.md
```diff
@@ -1140,8 +1165,16 @@ The `RELCOUNT` function returns the number of non-null records in multiple rows
...
+- `by` (optional): 1+ collation values, either as a single expression or an iterable of expressions, used to order the records of the current context. Can only be provided if `cumulative` is True or `frame` is provided.
+- `cumulative` (optional): optional argument (default `False`) that can only be `True` if the `by` argument is provided. If `True`, then instead of returning the count of all of the data in the context, returns the cumulative count of all rows up to and including the current row when sorted according to the keys in the `by` argument. This argument cannot be provided if `frame` is also provided.
+- `frame` (optional): optional argument that, if provided, must be a tuple of two values `(lower, upper)` that are each either integer literals or None. If provided, it means the relative count is performed on a subset of the selected records based on the values of `lower` and `upper` when sorted according to the keys in the `by` argument:
```
hadia206, line 1170: Same
```suggestion
- `frame` (optional): optional argument that, if provided, must be a tuple of two values `(lower, upper)`, where each value is either an integer literal or None. If provided, it means the relative count is performed on a subset of the selected records based on the values of `lower` and `upper` when sorted according to the keys in the `by` argument:
```

documentation/functions.md
```diff
@@ -1156,18 +1189,30 @@ Nations.customers.CALCULATE(ratio = acctbal / RELCOUNT(KEEP_IF(acctbal, acctbal
...
+- `by` (optional): 1+ collation values, either as a single expression or an iterable of expressions, used to order the records of the current context. Can only be provided if `cumulative` is True or `frame` is provided.
+- `cumulative` (optional): optional argument (default `False`) that can only be `True` if the `by` argument is provided. If `True`, then instead of returning the number records in the context, returns the cumulative number of records up to and including the current row when sorted according to the keys in the `by` argument. This argument cannot be provided if `frame` is also provided.
+- `frame` (optional): optional argument that, if provided, must be a tuple of two values `(lower, upper)` that are each either integer literals or None. If provided, it means the relative size is performed on a subset of the selected records based on the values of `lower` and `upper` when sorted according to the keys in the `by` argument:
```
hadia206, line 1208: Same
```suggestion
- `frame` (optional): optional argument that, if provided, must be a tuple of two values `(lower, upper)`, where each value is either an integer literal or None. If provided, it means the relative size is performed on a subset of the selected records based on the values of `lower` and `upper` when sorted according to the keys in the `by` argument:
```

pydough/sqlglot/sqlglot_relational_expression_visitor.py
```diff
@@ -71,6 +71,91 @@ def visit_call_expression(self, call_expression: CallExpression) -> None:
...
+
+        Returns:
+            The window specification if applicable, otherwise None.
+        """
+        lower: int | None = None
+        upper: int | None = None
+        if kwargs.get("cumulative", False):
+            # If cumulative=True, that is the same as the frame (None, 0)
```
hadia206, line 102: Shouldn't the comment be `If cumulative=False`? or is the if statement should be `"cumulative", True`?

pydough/sqlglot/sqlglot_relational_expression_visitor.py
```diff
@@ -71,6 +71,91 @@ def visit_call_expression(self, call_expression: CallExpression) -> None:
...
+            lower, upper = None, 0
+
+        elif "frame" in kwargs:
+            # If frame is provided, parse it to extract the lower and upper bounds
+            # which are assumed to be correctly formatted
+            frame = kwargs["frame"]
+            assert isinstance(frame, tuple) and len(frame) == 2
+            lower_raw, upper_raw = frame[0], frame[1]
```
hadia206, line 110: Is the name meant be be `raw` not `row`?

pydough/sqlglot/sqlglot_relational_expression_visitor.py
```diff
@@ -175,13 +260,8 @@ def visit_window_expression(self, window_expression: WindowCallExpression) -> No
             window_args["order"] = sqlglot_expressions.Order(
                 this=None, expressions=order_exprs
             )
-        if cumulative:
```
hadia206, line 178: Where did setting the start/end values if `cumulative` is `True` go?

pydough/sqlglot/sqlglot_relational_visitor.py
hadia206: I would like to see that explanation in the code as a comment but in a shorter version somehow

documentation/functions.md
hadia206, line 1091: We need to mention that if `frame` is provided, `by` must be provided as well.

tests/bad_pydough_functions.py
```diff
@@ -54,6 +54,28 @@ def bad_window_6():
     return Customers.CALCULATE(RELAVG(acctbal, cumulative=True))
 
 
+def bad_window_7():
+    # Relcount with frame but without by
+    return Customers.CALCULATE(RELCOUNT(acctbal, frame=(-10, 0)))
+
+
+def bad_window_8():
+    # Relsize with frame but AND cumulative
```
hadia206, line 63:
```suggestion
    # Relsize with frame AND cumulative
```

tests/test_logging.py
```diff
@@ -185,19 +185,19 @@ def test_execute_df_logging(
...
-      l_extendedprice * (
-        1 - l_discount
-      )
-    )) AS sum_disc_price,
     SUM(l_extendedprice * (
       1 - l_discount
     ) * (
       1 + l_tax
     )) AS sum_charge,
+    SUM(l_discount) AS sum_discount,
```
hadia206, line 194: Why these changes?

tests/test_sql_refsols/defog_broker_adv11_ansi.sql
```diff
@@ -1,7 +1,7 @@
 WITH _t1 AS (
   SELECT
-    sbcustomer.sbcustemail AS email,
-    sbcustomer.sbcustid AS _id
+    sbcustomer.sbcustid AS _id,
```
hadia206, line 3: Why this change?

tests/test_sql_refsols/window_sliding_frame_relsize_ansi.sql
hadia206: That's cool

hadia206 (approved):
Overall looks good to me.

tests/test_logging.py
knassre-bodo: The same alphabetization of column order outside of the final select.

tests/test_sql_refsols/defog_broker_adv11_ansi.sql
knassre-bodo: The same alphabetization of column order outside of the final select.

pydough/sqlglot/sqlglot_relational_expression_visitor.py
knassre-bodo: These are called "raw" because I'm using these temporary variables to get around mypy type restrictions, because mypy doesn't know what type `frame[0]` or `frame[1]` are yet.

pydough/sqlglot/sqlglot_relational_expression_visitor.py
knassre-bodo: The comment is correct, since this is what we do if `kwargs` contains `cumulaative` and its value is True. We use `False` as the default so that "cumulative isn't in the kwargs" has the same effect as "cumulative is in the kwargs but its value is False".

pydough/sqlglot/sqlglot_relational_expression_visitor.py
knassre-bodo: This is _also_ done in `get_window_spec` (which handles both `cumulative` and `frame`)

pydough/sqlglot/sqlglot_relational_visitor.py
knassre-bodo: I can expand on the comment on line 565-566 to say the intent.

PR #359, opened 2025-06-10
Attempt to increase effectiveness of aggregation splitting

### LINKED ISSUE #365: Refactor join node relational structure
Goals of the overhaul:
- Every `JOIN` node in the relational algebra should have exactly 2 inputs.
- The condition & join type are now single values, instead of lists of values.
- Every join has a `cardinality` enum describing the behavior from the perspective of the LHS being joined onto the RHS: singular vs plural, and does the join with the RHS effectively filter the LHS. The value of this enum depends on the circumstances the join is being created with, and can be affected by filter pushdown
- Several tests will need to be adjusted accordingly

knassre-bodo:
Resolves #365 and #366.

Several related changes to relational nodes & relational optimizations:
- Refactored `JOIN` nodes to always have 2 inputs, their condition/type be single values instead of lists, and have a CARDINALITY enum describing the type of join from the perspective of the LHS node: does it match to 1 or multiple rows of the RHS, and is it just accessing or does joining with the RHS also filter the LHS.
- Fixed minor bugs in projection merging.
- Adjusted how aggregation splitting/pushdown works so it can push down aggregates into either/both input simultaneously, but will not push down aggregations into the LHS if the cardinality information tells us that the join with the RHS will first filter the LHS, thus causing fewer rows to get aggregated.
- Adjust filter pushdown to also attempt to push filters into a join condition (if legal), which will also affect their cardinality. The cardinality of the join now becomes the means to detect whether the join is prunable (cannot prune the RHS of a join that matches onto multiple rows, or one that filters the LHS). This feature replaces the existing `is_prunable` checks.
- Fixed a bug with SQLGlot CTE merging that caused left joins to get flattened incorrectly alongside inner joins in a way that broke the correctness of the left join. The bug was fixed by manually copying and overriding a section of the SQLGlot optimizations.

## Example: removing potentially sub-optimal aggregation splitting

Consider a relational plan such as this that computes the total account balance of all German suppliers:
```
AGGREGATE(keys={}, aggregations={german_total=SUM(account_balance)})
  JOIN(condition=t0.nation_key == t1.key, type=INNER, cardinality=SINGULAR_FILTER, columns={account_balance=t0.account_balance})
    SCAN(suppliers)
    FILTER(name == "GERMANY")
      SCAN(nations)
```
In this instance, we could push the aggregate into the input as follows:
```
AGGREGATE(keys={}, aggregations={german_total=SUM(agg_0)})
  JOIN(condition=t0.nation_key == t1.key, type=INNER, cardinality=SINGULAR_FILTER, columns={agg_0=t0.agg_0})
    AGGREGATE(keys={nation_key}, aggregations={agg_0=SUM(account_balance)})
      SCAN(suppliers)
    FILTER(name == "GERMANY")
      SCAN(nations)
```
However, this plan is potentially LESS optimal than the original because we are aggregating the entire table by nation key,  instead of first using the join with the filtered nations to filter the suppliers. This PR will make it so the aggregation transformation does not occur in situations like the one above (although the implementing database may still choose to do so if it believe it to be optimal, e.g. what Snowflake sometimes does with its relational optimizer).

## Example: adding more aggregation splitting

Now consider this other relational plan that counts the total supplier wealth per nation, but also includes both the key and name of the nation.
```
AGGREGATE(keys={nation_key}, aggregations={nation_name=ANYTHING(nation_name), supp_wealth=SUM(account_balance)})
  JOIN(condition=t0.nation_key == t1.key, type=INNER, cardinality=SINGULAR_ACCESS, columns={nation_key=t1.key, nation_name=t1.name, account_balance=t0.account_balance})
    SCAN(suppliers)
    SCAN(nations)
```

Previously, aggregate splitting could not happen here because `SUM(account_balance)` would be pushed into the LHS while `ANYTHING(nation_name)` would be pushed into the RHS, but now with this PR the following can be done:
```
AGGREGATE(keys={nation_key}, aggregations={nation_name=ANYTHING(nation_name), supp_wealth=SUM(agg_0)})
  JOIN(condition=t0.nation_key == t1.key, type=INNER, cardinality=SINGULAR_ACCESS, columns={nation_key=t1.key, nation_name=t1.nation_name, agg_0=t0.agg_0})
    AGGREGATE(keys={nation_key}, aggregations={agg_0=SUM(account_balance)})
      SCAN(suppliers)
    AGGREGATE(keys={key}, aggregations={nation_name=ANYTHING(name)})
      SCAN(nations)
```

Then, with existing optimizations down the line, this plan will be re-written as follows:
```
JOIN(condition=t0.nation_key == t1.key, type=INNER, cardinality=SINGULAR_ACCESS, columns={nation_key=t1.key, nation_name=t1.nation_name, supp_wealth=t0.agg_0})
  AGGREGATE(keys={nation_key}, aggregations={agg_0=SUM(account_balance)})
    SCAN(suppliers)
  PROJECT(key=key, nation_name=name)
    SCAN(nations)
```


pydough/conversion/agg_split.py
```diff
@@ -160,33 +164,25 @@ def transpose_aggregate_join(
     Returns:
         The transformed node. The transformation is also done-in-place.
     """
+    count_ref: RelationalExpression | None = None
     agg_input_name: str | None = join.default_input_aliases[agg_side]
-    # Keep a dictionary for the projection columns that will be used to post-process
-    # the output of the aggregates, if needed.
     need_projection: bool = False
-    projection_columns: dict[str, RelationalExpression] = {**node.keys}
```
knassre-bodo, line 167: This logic is now passed in by the calling site.

pydough/relational/relational_nodes/column_pruner.py
```diff
@@ -152,8 +152,14 @@ def _prune_node_columns(
...
-            (output.join_types == [JoinType.LEFT])
-            or (output.join_types == [JoinType.INNER] and output.is_prunable)
+            (output.join_type == JoinType.LEFT)
+            or (
+                output.join_type == JoinType.INNER
+                and not (
+                    output.cardinality.potentially_filters
+                    or output.cardinality.potentially_plural
+                )
+            )
```
knassre-bodo, line 162: We are deciding whether to remove an inner join from A to B when all of the columns of B are unused. WE cannot do so if joining with B can cause rows of A to either be added or deleted.

pydough/conversion/relational_converter.py
```diff
@@ -1369,24 +1390,18 @@ def optimize_relational_tree(
     # happens underneath the join.
     root = confirm_root(split_partial_aggregates(root, configs))
 
-    # Step 4: re-run projection merging.
-    root = confirm_root(merge_projects(root))
```
knassre-bodo, line 1373: No need to do this before redundant agg removal.

pydough/sqlglot/override_merge_subqueries.py
```diff
@@ -0,0 +1,119 @@
...
+)
+from sqlglot.optimizer.merge_subqueries import _mergeable as _old_mergeable
+from sqlglot.optimizer.scope import Scope, traverse_scope
+
+if t.TYPE_CHECKING:
+    from sqlglot._typing import E
+    from sqlglot.optimizer.merge_subqueries import FromOrJoin
+
+
+def merge_subqueries(expression: E, leave_tables_isolated: bool = False) -> E:
```
knassre-bodo, line 30: Everything in this file is copied 1:1 from `merge_subqueries.py` except the `PYDOUGH CHANGE` part.

pydough/sqlglot/execute_relational.py
```diff
@@ -34,6 +33,7 @@
     RelationalExpression,
 )
 
+from .override_merge_subqueries import merge_subqueries
```
knassre-bodo, line 36: This change allowed fixing a bug with CTE merging in sqlglot. Currently, they allow merging `select * from t1 left join (select * from t2 inner join t3)` into `select * from t1 left join t2 inner join t3`, which is incorrect because the inner join must happen **before** `t2` gets joined onto `t1` (otherwise the results will change). This bug only manifested once some of the relational plans shifted during this PR.

pydough/relational/relational_nodes/join.py
```diff
@@ -12,14 +12,136 @@
...
+        return self in (
+            JoinCardinality.UNKNOWN_ACCESS,
+            JoinCardinality.UNKNOWN_FILTER,
+            JoinCardinality.UNKNOWN_UNKNOWN,
+        )
+
+
 class Join(RelationalNode):
     """
     Relational representation of all join operations. This single
```
knassre-bodo, line 147: NOTE TO SELF: FIX THIS COMMENT

hadia206 (commented):
Overall code is doing what it says. 
I have some questions/comments below

pydough/conversion/relational_converter.py
```diff
@@ -822,14 +840,22 @@ def translate_sub_collection(
                 assert node.general_condition is not None
                 join_cond = node.general_condition
 
+            case CartesianProductMetadata():
```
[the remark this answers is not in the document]
knassre-bodo: 1. Whitelisting the types that can be used in this function, raising an error for anything else (thus making it easier to catch gaps if we ever add more metadata types).
2. This one is `pass` because if its cartesian product, then we leave the join condition & general join both blank (which denotes a cross join). Will add as a comment

pydough/conversion/relational_converter.py
```diff
@@ -1397,24 +1422,18 @@ def optimize_relational_tree(
...
     root = confirm_root(merge_projects(root))
 
-    # Step 7: prune unused columns
+    # Step 6: prune unused columns.
     root = ColumnPruner().prune_unused_columns(root)
 
-    # Step 3: merge adjacent projections, unless it would result in excessive
-    # duplicate subexpression computations.
-    merged_root = merge_projects(root)
-    assert isinstance(merged_root, RelationalRoot)
```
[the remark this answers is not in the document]
knassre-bodo: `confirm_root` takes care of it

pydough/relational/rel_util.py
```diff
@@ -339,6 +374,77 @@ def transpose_expression(
...
+        references.
+    """
+    match expr:
+        case LiteralExpression() | CorrelatedReference():
+            return expr
+        case ColumnReference():
+            if expr.input_name == input_names[0]:
+                return left_columns.get(expr.name, expr)
+            else:
+                assert expr.input_name == input_names[1]
```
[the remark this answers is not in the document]
knassre-bodo: This should always be true, but good point will change it

pydough/relational/relational_nodes/join.py
```diff
@@ -12,14 +12,136 @@
...
 
 class JoinType(Enum):
+    """
+    Enum describing the type of join operation.
+    """
+
     INNER = "inner"
     LEFT = "left"
-    RIGHT = "right"
-    FULL_OUTER = "full outer"
```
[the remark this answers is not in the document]
knassre-bodo: We never ever generate these two, since joins are always generated as left/inner/semi/anti. If we ever have a situation where these two are possible, we can add them back in later.

PR #364, opened 2025-06-13
Improve error messages for accessing terms that do not exist

### LINKED ISSUE #358: Improve error messages for accessing terms that do not exist
The QDAG nodes `GlobalContext` and `CollectionAccess` both have a `get_term` API that returns an error message if you try to access a term that does not exist. For example, when using the `TPCH` graph, `customers.CALCULATE(n=c_name)` will raise a QDAG exception because there is no property called `c_name` in customers `"Unrecognized term of simple table collection 'customers' in graph 'TPCH': 'c_name'"`.

The goal of this issue is to improve the error message by including the name of the correct property in the collection, e.g.: `"Unrecognized term of simple table collection 'customers' in graph 'TPCH': 'c_name' (did you mean `name`?)"`.

How this should work:
- When the error message is raised in either file, call a function `find_possible_name_matches` (defined in another file) which takes in the bad term name, a set of valid term names (use `self.all_terms` to obtain such a set), and the API will return the list of names to include in the error message as possible alternatives.
- The new API `find_possible_name_matches(bad_name: str, good_names: set[str]) -> list[str]` will find which names from the set of legal terms (`good_names`) are the closest matches to `bad_name` via minimum edit distance:
  - For each name `good_name` in `good_names`, compute the minimum edit distance `me = min_edit_distance(bad_name, good_name)`. Augment this formula so that if two characters are the same but with different capitalization, then changing the capitalization counts as a 0.1 difference instead of a 1.0 difference (e.g. `NAME` and `name` should have a `me` value of `0.4` instead of `4.0`)
  - Identify the minimum `me` value `closest_match`, and do one of the following (whichever one is larger):
    - Return all names from `good_names` that have a `me <= closest_match + 2`
    - Return all names from `good_names` that have a `me <= closest_match * 1.1`
    - Return the three names from `good_names` with the smallest values of `me` (breaking ties by name)

Here is the optimized formula for minimum edit distance between two strings ([originally from here](https://en.wikipedia.org/wiki/Levenshtein_distance#Iterative_with_two_matrix_rows)) that is used by BodoSQL to compute minimum edit distance between two strings:

```py
def min_edit_distance(s: str, t: str) -> int:
   # Ensure that s is the shorter of the two strings
    if len(s) > len(t):
        s, t = t, s
    m, n = len(s), len(t)

    # Use a 2 x (m + 1) array to represent an n x (m + 1) array since you only
    # need to consider the previous row to generate the next row, therefore the
    # same two rows can be recycled
    row, otherRow = 1, 0
    arr = np.zeros((2, m + 1), dtype=np.uint32)

    # MED(X, "") = len(X)
    arr[0, :] = np.arange(m + 1)

    for i in range(1, n + 1):
        # MED("", X) = len(X)
        arr[row, 0] = i

        # Loop over the rest of s to see if it matches with the corresponding letter of t
        for j in range(1, m + 1):
            # If these two characters match, then the diagonal entry above them is the MED
            if s[j - 1] == t[i - 1]:
                arr[row, j] = arr[otherRow, j - 1]

            # Otherwise, it is the min of the diagonal entry and the one above / to the left
            else:
                arr[row, j] = 1 + min(
                    arr[row, j - 1], arr[otherRow, j], arr[otherRow, j - 1]
                )

        row, otherRow = otherRow, row

    return arr[n % 2, m]
```

The changes that should be made for the augmentation:
- Change the dtype from `np.int32` to `np.float32` and the return type to float
- Change the body of the inner loop to the following
```py
substitution_cost: float
if s[j-1] == t[i-1]: substitution_cost = 0.0
elif s[j-1].lower() == t[i-1].lower(): substitution_cost = 0.1
else: substitution_cost = 1.0
arr[row, j] = min(arr[row, j - 1] + 1.0, arr[otherRow, j] + 1.0, arr[otherRow, j - 1] + substitution_cost)
```

Note: existing error tests should be updated, and new error tests should be added to verify exactly which names are included in the error message.

john-sanchez31:
Resolves #358


pydough/qdag/collections/calculate.py
```diff
@@ -149,6 +149,15 @@ def get_expression_position(self, expr_name: str) -> int:
     def get_term(self, term_name: str) -> PyDoughQDAG:  # type: ignore
         if term_name in self.calc_term_values:
             return self.calc_term_values[term_name]
+        elif term_name not in self.all_terms:
+            error_message = f"Unrecognized term of {self.to_string()}: {term_name!r}"
+            suggestions = self.find_possible_name_matches(term_name=term_name)
```
hadia206, line 154: If possible, add type hints to suggestions

pydough/qdag/collections/collection_access.py
```diff
@@ -132,9 +132,16 @@ def get_term(self, term_name: str) -> PyDoughQDAG:
             return Reference(context, term_name)
 
         if term_name not in self.all_terms:
-            raise PyDoughQDAGException(
+            error_message = (
                 f"Unrecognized term of {self.collection.error_name}: {term_name!r}"
             )
+            suggestions = self.find_possible_name_matches(term_name=term_name)
```
hadia206, line 138: Here too

pydough/qdag/collections/collection_qdag.py
```diff
@@ -340,3 +342,114 @@ def to_tree_string(self) -> str:
...
+        name based on minimum edit distance.
+
+        Args:
+            name (str): The name to match against the list of candidates.
+
+        Returns:
+            List[str]: A list of candidate names, based on the closest matches.
+        """
+
+        terms_distance_list = []
```
hadia206, line 358: Also here, let's add type hint

pydough/qdag/collections/collection_qdag.py
```diff
@@ -340,3 +342,114 @@ def to_tree_string(self) -> str:
...
+        for tuple in sorted_list:
+            # all names that have a me <= closest_match + 2
+            if tuple[0] <= closest_match[0] + 2:
+                good_matches_1.append(tuple[1])
+
+            # all names that have a me <= closest_match * 1.1
+            if tuple[0] <= closest_match[0] * 1.1:
+                good_matches_2.append(tuple[1])
+
+        # returning the larger
```
hadia206, line 389: Why?

pydough/qdag/collections/collection_qdag.py
```diff
@@ -340,3 +342,114 @@ def to_tree_string(self) -> str:
...
+            return good_matches_2
+
+        else:
+            return good_matches_3
+
+    @staticmethod
+    def min_edit_distance(s: str, t: str) -> float:
+        """
+        Computes the minimum edit distance between two strings using the
+        Levenshtein distance algorithm.
```
hadia206, line 407: Did you implement that code yourself or used an external resource? 
If used an external resource, let's add link here (Copied or Adapted from link)

pydough/qdag/collections/collection_qdag.py
```diff
@@ -340,3 +342,114 @@ def to_tree_string(self) -> str:
...
+        # Ensures str1 is the shorter string
+        if len(s) > len(t):
+            s, t = t, s
+        m, n = len(s), len(t)
+
+        # Use a 2 x (m + 1) array to represent an n x (m + 1) array since you only
+        # need to consider the previous row to generate the next row, therefore the
+        # same two rows can be recycled
+
+        row, otherRow = 1, 0
```
hadia206, line 425: nit: replace `otherRow` with `previousRow` for better naming.

pydough/qdag/collections/collection_qdag.py
```diff
@@ -340,3 +342,114 @@ def to_tree_string(self) -> str:
...
+
+                arr[row, j] = min(
+                    arr[row, j - 1] + 1.0,
+                    arr[otherRow, j] + 1.0,
+                    arr[otherRow, j - 1] + substitution_cost,
+                )
+
+            row, otherRow = otherRow, row
+
+        return arr[n % 2, m]  # Return the last computed row's last element
```
hadia206, line 455: Since the code is doing a swap between the rows at the end, wouldn't it be easier to do `return arr[otherRow, m]`

pydough/qdag/collections/global_context.py
```diff
@@ -116,9 +116,17 @@ def get_expression_position(self, expr_name: str) -> int:
 
     def get_term(self, term_name: str) -> PyDoughQDAG:
         if term_name not in self.collections:
-            raise PyDoughQDAGException(
+            error_message = (
                 f"Unrecognized term of {self.graph.error_name}: {term_name!r}"
             )
+            suggestions = self.find_possible_name_matches(term_name=term_name)
```
hadia206, line 122: Type hint

pydough/qdag/collections/partition_by.py
```diff
@@ -188,7 +188,14 @@ def get_term(self, term_name: str) -> PyDoughQDAG:
         elif term_name == self.child.name:
             return PartitionChild(self.child, self.child.name, self)
         else:
-            raise PyDoughQDAGException(f"Unrecognized term: {term_name!r}")
+            error_message = f"Unrecognized term of {self}: {term_name!r}"
+            suggestions = self.find_possible_name_matches(term_name=term_name)
```
hadia206, line 192: Same

pydough/qdag/collections/partition_child.py
```diff
@@ -102,6 +103,17 @@ def get_term(self, term_name: str):
                     assert context.ancestor_context is not None
                     context = context.ancestor_context
             return Reference(context, term_name)
+
+        elif term_name not in self.all_terms:
+            error_message = f"Unrecognized term of {self.to_string()}: {term_name!r}"
+            suggestions = self.find_possible_name_matches(term_name=term_name)
```
hadia206, line 109: Same

hadia206 (commented):
Overall looks good to me.
Please see my comments below

pydough/qdag/collections/collection_qdag.py
```diff
@@ -358,3 +361,132 @@ def to_tree_string(self) -> str:
...
+
+        # filtering the result
+        for tuple in sorted_list:
+            # all names that have a me <= closest_match + 2
+            if tuple[0] <= closest_match[0] + 2:
+                good_matches_1.append(tuple[1])
+
+            # all names that have a me <= closest_match * 1.1
+            if tuple[0] <= closest_match[0] * 1.1:
+                good_matches_2.append(tuple[1])
```
knassre-bodo, line 406: Here is how to fix this:
```py
# Find all the names that are candidates to be near-best matches
for me, name in sorted_list:
  # Is the name within 2 of the best match
  if me <= closest_match[0] + 2:
    good_matches_1.append(name)
  # Is the name within 10% of the best match
  if me <= closest_match[0] * 1.1
    good_matches_2.append(name)
```

pydough/qdag/collections/calculate.py
john-sanchez31: Done

pydough/qdag/collections/calculate.py
```diff
@@ -149,6 +149,15 @@ def get_expression_position(self, expr_name: str) -> int:
     def get_term(self, term_name: str) -> PyDoughQDAG:  # type: ignore
         if term_name in self.calc_term_values:
             return self.calc_term_values[term_name]
+        elif term_name not in self.all_terms:
+            error_message = f"Unrecognized term of {self.to_string()}: {term_name!r}"
```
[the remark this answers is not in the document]
john-sanchez31: Implemented in the base class `collection_qdag.py`, It returns the string because I was having some problems with the returns where this function is called

tests/test_pydough_functions/bad_pydough_functions.py
```diff
@@ -317,6 +317,122 @@ def bad_unsupported_kwarg3():
...
+
+
+# collection.collection.property_of_first_collection
+def bad_name_5():
+    return customers.orders.CALCULATE(c_name)
+
+
+# collection suppliers doesn't exist
+def bad_name_6():
+    return customers.suppliers.CALCULATE(key)
```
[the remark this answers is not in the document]
john-sanchez31: Done

tests/test_pydough_functions/bad_pydough_functions.py
```diff
@@ -317,6 +317,122 @@ def bad_unsupported_kwarg3():
...
+def bad_name_1():
+    return customers.CALCULATE(c_name)
+
+
+def bad_name_2():
+    return TPCH.CUSTS
+
+
+def bad_name_3():
+    return TPCH.CALCULATE(foo=1, bar=2, fizz=3, BUZZ=4).CALCULATE(fizzbuzz)
```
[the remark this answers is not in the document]
john-sanchez31: Added

pydough/qdag/collections/collection_qdag.py
```diff
@@ -358,3 +361,132 @@ def to_tree_string(self) -> str:
...
+        return arr[previousRow, m]  # Return the last computed row's last element
+
+    def name_mismatch_error(self, term_name: str) -> str:
+        """
+        Raises a name mismatch error with suggestions if possible.
+        Args:
+            term_name (str): The name of the term that caused the error.
+        """
+
+        error_message: str = f"Unrecognized term of {self.to_string()}: {term_name!r}"
```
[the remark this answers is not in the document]
john-sanchez31: Added

pydough/qdag/collections/collection_qdag.py
```diff
@@ -358,3 +361,132 @@ def to_tree_string(self) -> str:
...
+
+        Returns:
+            list[str]: A list of candidate names, based on the closest matches.
+        """
+
+        terms_distance_list: list[tuple[float, str]] = []
+
+        for term in self.all_terms:
+            # get the minimum edit distance
+            me = self.min_edit_distance(term_name, term)
```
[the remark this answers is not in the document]
john-sanchez31: I've added the type hint to `me`, but I don't really know which `name` you are referring to.

juankx-bodo (approved):
LGTM

pydough/qdag/collections/collection_qdag.py
```diff
@@ -358,3 +361,135 @@ def to_tree_string(self) -> str:
...
+        sorted_list = sorted(terms_distance_list)
+
+        closest_match = sorted_list[0]
+
+        # List with all names that have a me <= closest_match + 2
+        good_matches_1: list[str] = []
+        # List with all names that have a me <= closest_match * 1.1
+        good_matches_2: list[str] = []
+        # List with the top 3 closest matches (me) breaking ties by name
+        good_matches_3: list[str] = [name for _, name in sorted_list[:3]]
```
knassre-bodo, line 396: Let's rename these slightly:
- `good_matches_1` -> `matches_within_2`
- `good_matches_2` -> `matches_within_10_pct`
- `good_matches_3` -> `matches_top_3`

pydough/qdag/collections/collection_qdag.py
```diff
@@ -358,3 +361,135 @@ def to_tree_string(self) -> str:
...
+
+        for term in self.all_terms:
+            # get the minimum edit distance
+            me: float = self.min_edit_distance(term_name, term)
+            terms_distance_list.append((me, term))
+
+        if terms_distance_list == []:
+            return []
+        # sort the list by minimum edit distance break ties by name
+        sorted_list = sorted(terms_distance_list)
```
knassre-bodo, line 387: Nitpick: instead of creating a new `sorted_list`, just call `terms_distance_list.sort()` to sort it in-place

pydough/qdag/collections/collection_qdag.py
```diff
@@ -358,3 +361,135 @@ def to_tree_string(self) -> str:
...
+        ):
+            return good_matches_1
+
+        elif len(good_matches_2) >= len(good_matches_1) and len(good_matches_2) >= len(
+            good_matches_3
+        ):
+            return good_matches_2
+
+        else:
+            return good_matches_3
```
knassre-bodo, line 420: NIT: you can use higher-order functions to make this more compact. The following code takes the maximum of the 3 lists, using the length of each list to determine which list is the "max" (with the values of the lists serving as a tiebreaker)
```
return max(
  [good_matches_1, good_matches_2, good_matches_3],
  key=lambda x: (len(x), x)
)
```

pydough/qdag/collections/collection_qdag.py
```diff
@@ -358,3 +361,135 @@ def to_tree_string(self) -> str:
...
+        with a 2-row array is used to save memory.
+        Link:
+        https://en.wikipedia.org/wiki/Levenshtein_distance#Iterative_with_two_matrix_rows
+
+        Args:
+            str1 (str): The first string.
+            str2 (str): The second string.
+
+        Returns:
+            float: The minimum edit distance between the two strings.
```
knassre-bodo, line 436: Let's mention the capitalization augmentation detail: substituting a character for the same character with different capitalization is considered 10% of the edit cost of replacing it with any other character.

knassre-bodo (approved):
A few final nitpicks, but otherwise LGTM @john-sanchez31!

PR #370, opened 2025-06-15
Add column bubbling, merging, and name cleanup to relation plans

knassre-bodo:
Adds a new relational optimization pass that alters the names of columns with the intention of making the final SQL text cleaner. Also adds further modifications to the subquery merging optimization from sqlglot to further compress the final SQL files for readability.
- "bubbles" up column names from the input relational nodes to the parent relational nodes, thus reducing the frequency of columns being renamed until the final root & eliminating many of the dummy intermediate names (e.g. `expr_4` or `agg_12`)
- Attempts to rename the output columns of aggregations from forms like `agg_5` to forms like the following: `SUM(val)` becomes `sum_val`, `AVG(x)` becomes `avg_x`, `COUNT()` becomes `n_rows`, etc.
- The `_mergeable` override for the subquery merigng optimization now includes the entire body of the original implementation because one section needs to be changed: instead of aggregation/groups **always** being banned from being merged from the inner scope into the outer scope, they are now **sometimes** allowed, unless:
   - The outer scope also has groupby/aggs
   - The outer scope has a join or a where clause
   - The outer scope has a join or a where clause
   - The inner scope has a window function
- Modifies the merging protocol slightly to account for the fact that it is possible for the inner scope to have a group by, which must be propagated to the outer scope


pydough/conversion/relational_converter.py
```diff
@@ -1435,7 +1436,15 @@ def optimize_relational_tree(
...
     root = ColumnPruner().prune_unused_columns(root)
 
-    # Step 7: re-run projection merging.
+    # Step 7: bubble up names from the leaf nodes to further encourage simpler
+    # naming without aliases, and also to delete duplicate columns where
+    # possible.
+    root = bubble_column_names(root)
+
+    # Step 8: re-run column pruning.
+    root = ColumnPruner().prune_unused_columns(root)
```
knassre-bodo, line 1445: Bubbling happens after column pruning because the name bubbling trick avoids name collisions, so if we remove the un-used column names it increases the amount of opportunities to bubble the names. However, we re-run pruning afterwards because the bubbling can also create more unused columns.

tests/test_sql_refsols/tpch_q17_ansi.sql
```diff
@@ -1,6 +1,6 @@
 WITH _t2 AS (
   SELECT
-    lineitem.l_extendedprice AS extended_price
+    lineitem.l_extendedprice
```
knassre-bodo, line 3: This is an instance of the "column bubbling" trick: the name `l_extendedprice` now gets used until `SUM` gets called on it, instead of being renamed along the way.

tests/test_sql_refsols/tpch_q13_sqlite.sql
```diff
@@ -1,27 +1,22 @@
 WITH _s1 AS (
   SELECT
-    COUNT(*) AS agg_0,
-    o_custkey AS customer_key
+    COUNT(*) AS n_rows,
+    o_custkey
```
knassre-bodo, line 4: This is an example of how column bubbling changed the name of the aggregation output from `agg_0` to `n_rows` to be more legible, and also avoided renaming `o_custkey`.

tests/test_sql_refsols/tpch_q18_ansi.sql
```diff
@@ -1,7 +1,7 @@
 WITH _t0 AS (
   SELECT
-    SUM(l_quantity) AS agg_0,
-    l_orderkey AS order_key
+    SUM(l_quantity) AS sum_l_quantity,
+    l_orderkey
```
knassre-bodo, line 4: This is a different example of the column bubbling altering the names of aggregations. Here, we are calling `SUM` on `l_quantity`, so instead of a nebulous name like `agg_0` we use `sum_l_quanity`.

tests/test_sql_refsols/simple_var_std_sqlite.sql
```diff
@@ -50,44 +50,21 @@ WITH _s1 AS (
...
-    CAST((
-      SUM((
-        POWER(s_acctbal, 2)
-      )) - (
-        CAST((
-          POWER(SUM(s_acctbal), 2)
-        ) AS REAL) / COUNT(s_acctbal)
-      )
-    ) AS REAL) / COUNT(s_acctbal) AS var,
-    s_nationkey AS nation_key
```
knassre-bodo, line 76: This is an example of the column bubbling removing duplicate computations (see further up in the file)

pydough/sqlglot/override_merge_subqueries.py
```diff
@@ -96,13 +96,57 @@ def merge_ctes(expression: E, leave_tables_isolated: bool = False) -> E:
...
+            # PyDough Change: merge groups
+            _merge_groups(outer_scope, inner_scope)
             _merge_order(outer_scope, inner_scope)
             _merge_hints(outer_scope, inner_scope)
             _pop_cte(inner_scope)
             outer_scope.clear_cache()
     return expression
 
 
+def _merge_groups(outer_scope: Scope, inner_scope: Scope) -> None:
```
knassre-bodo, line 108: This is based on existing functions in SQLGlot

pydough/conversion/column_bubbler.py
```diff
@@ -0,0 +1,336 @@
...
+
+    Returns:
+        A string representing the generated name, or `None` if no suitable
+        name can be generated based on the provided conditions.
+    """
+    if len(agg_expr.inputs) == 1:
+        input_expr = agg_expr.inputs[0]
+        if isinstance(input_expr, ColumnReference):
+            return f"{agg_expr.op.function_name.lower()}_{input_expr.name}"
+    if len(agg_expr.inputs) == 0 and agg_expr.op.function_name.lower() == "count":
```
juankx-bodo, line 76: What happens if there are 2 COUNT() from different tables?

pydough/sqlglot/override_merge_subqueries.py
```diff
@@ -122,5 +174,103 @@ def _mergeable(
...
+            )
+        ]
+        window_expressions_in_unmergable = [
+            column
+            for column in unmergable_window_columns
+            if column.table == inner_select_name and column.name in window_alias_names
+        ]
+        return any(window_expressions_in_unmergable)
+
+    def _outer_select_joins_on_inner_select_join():
```
john-sanchez31, line 200: Same here

pydough/sqlglot/override_merge_subqueries.py
```diff
@@ -122,5 +174,103 @@ def _mergeable(
...
+            return False
+        inner_from_table = inner_from.alias_or_name
+        inner_projections = {s.alias_or_name: s for s in inner_scope.expression.selects}
+        return any(
+            col.table != inner_from_table
+            for selection in selections
+            for col in inner_projections[selection].find_all(exp.Column)
+        )
+
+    def _is_recursive():
```
john-sanchez31, line 231: Could you explain what is this function for? Maybe adding a docstring. Also the type hint, if it applies

john-sanchez31 (approved):
LGTM, just a couple of questions below

pydough/sqlglot/override_merge_subqueries.py
knassre-bodo: Again this is from SQLGlot so I'm not touching this

pydough/sqlglot/override_merge_subqueries.py
```diff
@@ -122,5 +174,103 @@ def _mergeable(
...
+        cte = inner_scope.expression.parent
+        node = outer_scope.expression.parent
+
+        while node:
+            if node is cte:
+                return True
+            node = node.parent
+        return False
+
+    return (
```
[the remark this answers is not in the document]
knassre-bodo: Same here, also SQLGlot

PR #371, opened 2025-06-17
Enable de-correlation of pure SEMI/ANTI joins

knassre-bodo:
TODO: add description


tests/test_pipeline_correl.py
```diff
@@ -336,10 +336,11 @@
             PyDoughPandasTest(
                 correl_16,
                 "TPCH",
-                lambda: pd.DataFrame({"n": [929]}),
+                lambda: pd.DataFrame({"n": [230]}),
```
knassre-bodo, line 339: The change in the correct answer here is because I changed the test to be something that was better defined.

tests/test_plan_refsols/correl_16.txt
```diff
@@ -1,5 +1,19 @@
 ROOT(columns=[('n', n)], orderings=[])
  AGGREGATE(keys={}, aggregations={'n': COUNT()})
+<<<<<<< HEAD
```
hadia206, line 3: Looks like there's a unresolved merge conflict here

tests/test_pydough_functions/correlated_pydough_functions.py
```diff
@@ -267,19 +267,89 @@ def correl_16():
     # Count how many european suppliers have the exact same percentile value
     # of account balance (relative to all other suppliers) as at least one
     # customer's percentile value of account balance relative to all other
-    # customers. Percentile should be measured down to increments of 0.01%.
-    # (This is a correlated SEMI-joins)
-    selected_customers = nation.CALCULATE(rname=region.name).customers.WHERE(
-        (PERCENTILE(by=(account_balance.ASC(), key.ASC()), n_buckets=10000) == tile)
-        & (rname == "EUROPE")
+    # customers in the nation, int he same nation as the supplier. Percentile
```
hadia206, line 270:
```suggestion
    # customers in the nation, in the same nation as the supplier. Percentile
```

tests/test_relational_nodes_to_sqlglot.py
```diff
@@ -14,13 +14,11 @@
     Abs,
     Add,
     Binary,
-    Exists,
     Expression,
     From,
     Length,
     Literal,
     Lower,
-    Not,
```
hadia206, line 23: Why delete those?

tests/test_sql_refsols/cumulative_stock_analysis_ansi.sql
```diff
@@ -25,8 +25,8 @@ FROM main.sbtransaction AS sbtransaction
 JOIN main.sbticker AS sbticker
   ON sbticker.sbtickerid = sbtransaction.sbtxtickerid
 WHERE
-  EXTRACT(MONTH FROM sbtransaction.sbtxdatetime) = 4
-  AND EXTRACT(YEAR FROM sbtransaction.sbtxdatetime) = 2023
+  EXTRACT(MONTH FROM CAST(sbtransaction.sbtxdatetime AS DATETIME)) = 4
```
hadia206, line 28: For my understanding, how is this code affecting the generation of cast functions?

hadia206 (approved):
Looks good to me.
Left minor comments

pydough/conversion/hybrid_decorrelater.py
```diff
@@ -157,23 +161,39 @@ def remove_correl_refs(
...
                 # For window functions, recursively transform all of their
                 # arguments, partition keys, and order keys.
                 for idx, arg in enumerate(expr.args):
                     expr.args[idx] = self.remove_correl_refs(
-                        arg, parent, child_height, correl_level
+                        arg, parent, child_height, correl_level, new_parent_uni_keys
                     )
+                req_aug: bool = True
                 for idx, arg in enumerate(expr.partition_args):
+                    if (
```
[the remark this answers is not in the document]
knassre-bodo: Will do. The reason is that in this scenario, it is already correlated _further_ than the ancestor context we would link it to, so adding extra partition args would change the answer.

For example, if I am doing the following:
```py
suppliers.nation.CALCULATE(
  y=customers.orders.BEST(by=total_value.DESC(), per="nations").key,
  z=customers.orders.BEST(by=total_value.DESC()).key,
)
```
In this scenario, if we de-correlate the children containing `y` and `z`:
- The augmented partition keys passed in from the parent are `CORREL(BACK(1).key)` (referring to `suppliers.key`) and `CORREL(key)` (referring to `nation.key`)
- If we de-correlate, we need to add these new augmented partition keys to the partition for `z` or else it could change the answer once de-correlation occurs (since previously we had no per-clause when we were independant of `suppliers.nations`, but now if we are joining `suppliers.nations` onto `customers.orders` **BEFORE** computing the window function, we need to use the uniqueness keys of `suppliers.orders` to avoid double-counting entries from `customers.orders` in the window function.
- If we de-correlate, we can **NOT** add the new augmented partition keys to the partition for `y` or else it would change the answer bc we are now becoming MORE specific than we previously were (previously should have been with regards to just the nation, but now is with regards to the nation AND the supplier)

pydough/conversion/hybrid_decorrelater.py
```diff
@@ -246,24 +271,38 @@ def correl_ref_purge(
...
+                        child_height,
+                        correl_level,
+                        new_parent_uni_keys,
                     )
                 if isinstance(operation, HybridCalculate):
                     for str, expr in operation.new_expressions.items():
-                        operation.new_expressions[str] = self.remove_correl_refs(
-                            expr, old_parent, child_height, correl_level
-                        )
+                        operation.new_expressions[str] = operation.terms[name]
```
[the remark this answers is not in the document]
knassre-bodo: Minor optimization: can re-use `operation.terms[name]` since `new_expressions` is a subset of `terms` for calculate.

pydough/conversion/hybrid_decorrelater.py
```diff
@@ -351,6 +401,12 @@ def decorrelate_child(
         if child.connection_type.is_semi and child_idx == min(
             old_parent.correlated_children
         ):
+            if child.connection_type == ConnectionType.SEMI:
```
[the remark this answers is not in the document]
knassre-bodo: Another way to think about it: `.is_semi` means "does this connection type filter the parent on whether the child exists"

pydough/sqlglot/override_pushdown_predicates.py
```diff
@@ -0,0 +1,73 @@
...
+"""
+
+from sqlglot import exp
+from sqlglot.optimizer.pushdown_predicates import pushdown
+from sqlglot.optimizer.scope import build_scope
+
+
+def pushdown_predicates(expression, dialect=None):
+    """
+    Rewrite sqlglot AST to pushdown predicates in FROMS and JOINS
```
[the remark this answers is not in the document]
knassre-bodo: This docstring is, word-for-word, from sqlglot so I'm going to try avoiding changing it for the time being. I'm trying to contain the changes to JUST the part that gets replaced.

pydough/sqlglot/override_pushdown_predicates.py
```diff
@@ -0,0 +1,73 @@
...
+
+    Example:
+        >>> import sqlglot
+        >>> sql = "SELECT y.a AS a FROM (SELECT x.a AS a FROM x AS x) AS y WHERE y.a = 1"
+        >>> expression = sqlglot.parse_one(sql)
+        >>> pushdown_predicates(expression).sql()
+        'SELECT y.a AS a FROM (SELECT x.a AS a FROM x AS x WHERE x.a = 1) AS y WHERE TRUE'
+
+    Args:
+        expression (sqlglot.Expression): expression to optimize
```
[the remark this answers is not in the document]
knassre-bodo: Same as earlier

pydough/sqlglot/override_pushdown_predicates.py
```diff
@@ -0,0 +1,73 @@
...
+                    for k, (node, source) in selected_sources.items()
+                    if node.args.get("limit") is None
+                }
+
+                # a right join can only push down to itself and not the source FROM table
+                for k, (node, source) in selected_sources.items():
+                    parent = node.find_ancestor(exp.Join, exp.From)
+                    if isinstance(parent, exp.Join) and parent.side == "RIGHT":
+                        selected_sources = {k: (node, source)}
+                        break
```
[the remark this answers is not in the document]
knassre-bodo: I don't know, this is from the regular SQLGlot and I was trying to leave it alone except for the 3 lines I added.

tests/test_plan_refsols/correl_16.txt
knassre-bodo: I'm getting rid of this now (& fixing correl_16)

tests/test_relational_nodes_to_sqlglot.py
knassre-bodo: Formatter eliminated them bc they are no longer used in this file.

tests/test_sql_refsols/cumulative_stock_analysis_ansi.sql
knassre-bodo: I re-added the `canonicalize` pass from sqlglot which has multiple positive ripple effects, but one of its neutral side effects is that it inserts these cast calls.

tests/test_plan_refsols/correl_16.txt
knassre-bodo: I had to **further** modify `merge_subqueries` so it won't merge `INNER _QUERY-> OUTER_QUERY` if `INNER_QUERY` contains window functions.

pydough/conversion/hybrid_decorrelater.py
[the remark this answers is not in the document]
hadia206: Thanks for the detailed explanation :)

PR #372, opened 2025-06-17
Replace COUNT() with COUNT(*)

hadia206:
Generate SQL with `COUNT(*)` instead of `COUNT()` as `COUNT()` is not standard SQL. Writing `COUNT()` with no argument leads to a syntax error in most SQL databases

### DIFF
--- a/documentation/usage.md
+++ b/documentation/usage.md
@@ -401,7 +401,7 @@ FROM (
         WHERE name_3 = 'EUROPE'
     )
     LEFT JOIN (
-        SELECT nation_key, COUNT() AS agg_0
+        SELECT nation_key, COUNT(*) AS agg_0
         FROM (
             SELECT c_nationkey AS nation_key
             FROM main.CUSTOMER
--- a/pydough/sqlglot/transform_bindings/base_transform_bindings.py
+++ b/pydough/sqlglot/transform_bindings/base_transform_bindings.py
@@ -83,7 +83,6 @@ class BaseTransformBindings:
     ] = {
         pydop.SUM: sqlglot_expressions.Sum,
         pydop.AVG: sqlglot_expressions.Avg,
-        pydop.COUNT: sqlglot_expressions.Count,
         pydop.MIN: sqlglot_expressions.Min,
         pydop.MAX: sqlglot_expressions.Max,
         pydop.ANYTHING: sqlglot_expressions.AnyValue,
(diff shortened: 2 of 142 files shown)


pydough/sqlglot/transform_bindings/base_transform_bindings.py
```diff
@@ -1666,3 +1667,19 @@ def convert_std(
...
+        Returns:
+            The SQLGlot expression to calculate the count of the argument.
+        """
+        # If COUNT is called with no arguments, make it COUNT(*).
+        if len(args) == 0:
+            return sqlglot_expressions.Count(this=sqlglot_expressions.Star())
+        elif len(args) == 1:
+            return sqlglot_expressions.Count(this=args[0])
+        else:
+            raise ValueError(f"COUNT expects 0 or 1 argument, got {len(args)}")
```
john-sanchez31, line 1685: Looking at the `registered_expression_operators.py`, COUNT has `AllowAny()` as verifier. I was thinking. Wouldn't it be better if the length of args were restricted so you don't have to raise an error here.

hadia206: How would it be restricted? 
The raise error here is a form of saying it's restricted and signal that the code generated something wrong.

john-sanchez31: In the file `registered_expression_operators.py` COUNT is defined as: 
`
COUNT = ExpressionFunctionOperator(
    "COUNT", True, AllowAny(), ConstantType(NumericType())
)
`
Is it not necessary to change the `AllowAny()` ? Allowing just 0 or 1 arguments.

knassre-bodo: `COUNT` still requires exactly 1 argument when it is called by the user, but we don't check what the type is (since it could be a collection OR expression of any type). The difference is once we convert from QDAG to hybrid, we drop the argument if it was a collection (since `COUNT()` in hybrid/relational is the same thing as `COUNT(*)` in SQL).

hadia206: So I should fix that line in registered_expression_operators?
`COUNT = ExpressionFunctionOperator( "COUNT", True, AllowAny(), ConstantType(NumericType()) )`
from `AllowAny()` to `RequireArgRange(0,1)`

knassre-bodo: That works too

john-sanchez31 (approved):
Looks fine!

documentation/usage.md
```diff
@@ -401,7 +401,7 @@ FROM (
         WHERE name_3 = 'EUROPE'
     )
     LEFT JOIN (
-        SELECT nation_key, COUNT() AS agg_0
```
knassre-bodo, line 404: Good catch

knassre-bodo (approved):
LGTM

juankx-bodo (approved):
LGTM.

PR #377, opened 2025-06-20
[run SF] add some SF functions

hadia206:
These get re-written in PyDough, but for Snowflake they can just be converted to an ANONYMOUS function of the same name in Snowflake
- `STARTSWITH`, `ENDSWITH`, `CONTAINS`
- `LPAD`, `RPAD`
- `SIGN`
- `YEAR/QUARTER/MONTH/DAY/HOUR/MINUTE/SECOND`
- `DAYNAME`
- `SMALLEST` becomes `LEAST` 
- `LARGEST` becomes `GREATEST`
- `SLICE(string, start, stop, step)` becomes `SLICE(string, start, length)` 
     - It also removes `step` from list of arguments
     - Need to compute start/end if None
     - Computes `length` 
- `MEDIAN`
     - add the Snowflake dialect to the set of dialects that do NOT require to be re-written


pydough/sqlglot/sqlglot_relational_expression_visitor.py
```diff
@@ -295,6 +295,7 @@ def visit_literal_expression(self, literal_expression: LiteralExpression) -> Non
                     this=sqlglot_expressions.convert(dt.isoformat(sep=" ")),
                     to=sqlglot_expressions.DataType.build("TIMESTAMP"),
                 )
+        # TODO: Check Snowflake dialect for similar issues with date/time
```
knassre-bodo, line 298: This is just an ANSI thing, since each dialect (e.g. Snowflake) should already have a way of dealing with this.

pydough/sqlglot/transform_bindings/sf_transform_bindings.py
```diff
@@ -29,6 +29,42 @@ def convert_call_to_sqlglot(
         match operator:
             case pydop.SUM:
                 return self.convert_sum(args, types)
+            case pydop.STARTSWITH:
+                return sqlglot_expressions.Anonymous(
+                    this="STARTSWITH", expressions=args
+                )
```
knassre-bodo, line 35: Instead of overriding in `convert_call_to_sqlglot`, anything that already has its own `convert_xxx` method in the base class should get overridden here (the override will usually be a one-liner).

pydough/sqlglot/transform_bindings/sf_transform_bindings.py
```diff
@@ -50,3 +86,36 @@ def convert_sum(
             case _:
                 # For other types, use SUM directly
                 return sqlglot_expressions.Sum(this=arg[0])
+
+
+def convert_slice(
+    args: list[SQLGlotExpression], types: list[PyDoughType]
+) -> SQLGlotExpression:
```
knassre-bodo, line 93: Why is this being overidden here?

PR #389, opened 2025-07-08
Add error handling class that can be subclassed to customize error messages

### LINKED ISSUE #388: Add class object for error handling that can be bound to the session
Goal:
- Create a class with methods for every notable exception that can be raised within PyDough (at least the ones the LLM should care about), where the inputs are all of the possibly relevant information/objects, and the output is the exception object to be raised.
- Store an instance of this base class within the active session, so users (or the LLM team) can swap it out with a subclass.
- Have all of the relevant errors invoke the methods from the object stored in the active session to create the error that they raise.

The types of errors that should be affected:
- All unqualified errors, except for errors caused by patterns that shouldn't exist.
- All qualification errors, or errors raised within the QDAG module: type errors, accessing names that don't exist, singular vs plural errors, expression vs collection errors, etc., except for errors caused by patterns that shouldn't exist.
- A very small number of errors in hybrid/relational/sqlglot that correspond to malformed inputs as opposed to internal bugs or unsupported behavior

knassre-bodo:
Resolves #388. Also heavily refactors a lot of the error-related code, including moving all of the exception utilities/classes to a common module.


pydough/exploration/explain.py
```diff
@@ -130,15 +130,15 @@ def explain_property(property: PropertyMetadata, verbose: bool) -> str:
                         for cond_str in conditions:
                             lines.append(f"  {cond_str}")
                     case _:
-                        raise ValueError(
+                        raise NotImplementedError(
```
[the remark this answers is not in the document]
knassre-bodo: This is just an error I use in malformed cases that we should never reach. If we ever add new classes, they will trip these cases since we need to implement support for them. This should never be user/llm-facing.

pydough/jupyter_extensions/pydough_magic.py
```diff
@@ -33,7 +34,7 @@ def pydough(self, line="", cell="", local_ns=None):
         cell = self.shell.var_expand(cell)
         graph: GraphMetadata | None = pydough.active_session.metadata
         if graph is None:
-            raise Exception(
+            raise PyDoughSessionException(
```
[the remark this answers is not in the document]
knassre-bodo: Because we're not porting **_every_** error over, mostly just the ones that should be customized for the LLM.

pydough/pydough_operators/__init__.py
```diff
@@ -67,7 +67,7 @@
     "NOT",
     "PERCENTILE",
     "POPULATION_STD",
-    "POPULATION_VARIANCE",
+    "POPULATION_VAR",
```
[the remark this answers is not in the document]
knassre-bodo: Consistency with the naming.

pydough/sqlglot/transform_bindings/sqlglot_transform_utils.py
```diff
@@ -165,11 +167,15 @@ def truncation_string(self) -> str:
...
             case DateTimeUnit.QUARTER:
-                raise ValueError("Quarter unit does not have a truncation string.")
+                raise PyDoughSQLException(
+                    "Quarter unit does not have a truncation string."
+                )
             case DateTimeUnit.MONTH:
                 return "'%Y-%m-01 00:00:00'"
             case DateTimeUnit.WEEK:
-                raise ValueError("Week unit does not have a truncation string.")
+                raise PyDoughSQLException(
```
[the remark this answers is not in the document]
knassre-bodo: This should not be a user-facing error since we should never reach this case (we should never use `.truncation_string` for week).

tests/test_metadata_errors.py
```diff
@@ -765,7 +765,7 @@ def test_invalid_graphs(
         ),
         pytest.param(
             "parent.sub4",
-            "Malformed general join condition: 'is_prime(self.j1) != is_prime(self.j2)' (PyDough nodes is_prime is not callable. Did you mean to use a function?)",
+            "Malformed general join condition: 'is_prime(self.j1) != is_prime(self.j2)' (PyDough object is_prime is not callable. Did you mean: ISIN, LIKE, SUM, SLICE, STRIP, IFF, MIN, VAR, PREV, SIGN, SQRT, STRING, ABS, CEIL, FIND, HAS, HOUR, LPAD, RPAD, STD, YEAR, UPPER, DATETIME?)",
```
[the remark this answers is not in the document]
knassre-bodo: The point here is that it tried to call a function that didn't exist, so it suggested alternative function names that are similar.

pydough/errors/pydough_error_builder.py
```diff
@@ -0,0 +1,344 @@
...
+        Args:
+            `collection`: The collection in which the term is being accessed.
+            `expr`: The PyDoughQDAG expression
+
+        Returns:
+            An exception indicating the cardinality error.
+        """
+        raise PyDoughQDAGException(
+            f"Expected all terms in {collection.standalone_string} to be singular, but encountered a plural expression: {expr}"
+        )
```
hadia206, line 92: should this be `return PyDoughQDAGException` to match others?

pydough/errors/pydough_error_builder.py
```diff
@@ -0,0 +1,344 @@
...
+                rtol=0.1,
+                min_names=3,
+                insert_cost=0.5,
+                delete_cost=1.0,
+                substitution_cost=1.0,
+                capital_cost=0.1,
+            )
+        )
+
+    def down_streaming_conflict(
```
hadia206, line 57: nit: better to leave it as one word (downstream)
```suggestion
    def downstream_conflict(
```

pydough/errors/pydough_error_builder.py
```diff
@@ -0,0 +1,344 @@
...
+        """
+        from pydough.qdag import PyDoughExpressionQDAG
+
+        if isinstance(expr, PyDoughExpressionQDAG):
+            return PyDoughQDAGException(
+                f"Expected a collection, but received an expression: {expr}"
+            )
+        else:
+            return PyDoughQDAGException(
+                f"Expected a collection, but received {expr.__class__.__name__}:  {expr}"
```
hadia206, line 107: Can we explicitly handle the `None` case. I remember getting message like `Expected a collection, but received NoneType: None` which is confusing
Suggestion
```
if expr is None:
    return PyDoughQDAGException("Expected a collection, but received None")
```

pydough/errors/pydough_error_builder.py
```diff
@@ -0,0 +1,344 @@
...
+        """
+        from pydough.qdag import PyDoughCollectionQDAG
+
+        if isinstance(expr, PyDoughCollectionQDAG):
+            return PyDoughQDAGException(
+                f"Expected an expression, but received a collection: {expr}"
+            )
+        else:
+            return PyDoughQDAGException(
+                f"Expected an expression, but received {expr.__class__.__name__}:  {expr}"
```
hadia206, line 123: Same comment as above

pydough/errors/pydough_error_builder.py
```diff
@@ -0,0 +1,344 @@
...
+        Args:
+            `operator`: The operator that failed type verification.
+            `args`: The arguments passed to the operator.
+            `message`: The error message explaining the typing failure.
+
+        Returns:
+            An exception indicating the type verification failure.
+        """
+        arg_strings: list[str] = [str(arg) for arg in args]
+        raise PyDoughQDAGException(
```
hadia206, line 141: Again, for consistency this should be `return` instead of `raise`

pydough/errors/pydough_error_builder.py
```diff
@@ -0,0 +1,344 @@
...
+        Args:
+            `operator`: The operator that failed type inference.
+            `args`: The arguments passed to the operator.
+            `message`: The error message explaining the inference failure.
+
+        Returns:
+            An exception indicating the type inference failure.
+        """
+        arg_strings: list[str] = [str(arg) for arg in args]
+        raise PyDoughQDAGException(
```
hadia206, line 161: Same, for consistency this should be `return` instead of `raise`

pydough/errors/pydough_error_builder.py
```diff
@@ -0,0 +1,344 @@
...
+            An exception indicating the bad `columns` argument.
+        """
+        if isinstance(columns, list):
+            for column in columns:
+                if not isinstance(column, str):
+                    return PyDoughQDAGException(
+                        f"Expected `columns` argument to be a list of strings, found {column.__class__.__name__}"
+                    )
+            return PyDoughQDAGException(
+                "Expected `columns` argument to be a non-empty list"
```
hadia206, line 183: After iterating over columns why return an exception?
e.g. If columns = ["a", "b"] this still returns exception with "Expected non-empty list"
Shouldn't this be
```
if not columns:
        return PyDoughQDAGException("Expected `columns` argument to be a non-empty list")
```

pydough/errors/pydough_error_builder.py
```diff
@@ -0,0 +1,344 @@
...
+                if not isinstance(alias, str):
+                    return PyDoughQDAGException(
+                        f"Expected `columns` argument to be a dictionary where the keys are strings, found {alias.__class__.__name__}"
+                    )
+                if not isinstance(column, str):
+                    return PyDoughQDAGException(
+                        f"Expected `columns` argument to be a dictionary where the values are strings, found {column.__class__.__name__}"
+                    )
+            return PyDoughQDAGException(
+                "Expected `columns` argument to be a non-empty dictionary"
```
hadia206, line 196: Same. Add check for empty

pydough/errors/pydough_error_builder.py
hadia206: I believe this assumes that we have bad columns which is either empty or non-string. (plus being list or dict)
I still think adding explicit check for being empty to be clear and avoid misleading like what happened to me

pydough/errors/pydough_error_builder.py
```diff
@@ -0,0 +1,344 @@
...
+                substitution_cost=1.0,
+                capital_cost=0.1,
+            )
+
+            # Check if there are any suggestions to add
+            if len(suggestions) > 0:
+                suggestions_str: str = ", ".join(suggestions)
+                error_message += f" Did you mean: {suggestions_str}?"
+        else:
+            error_message += " Did you mean to use a function?"
```
hadia206, line 291: nit
```suggestion
            error_message += " Did you mean to access an attribute or method?""
```

pydough/errors/pydough_error_builder.py
```diff
@@ -0,0 +1,344 @@
...
+                    ancestor_idx = int(components[1])
+            # If an error was not found yet, figure out what is wrong with
+            # `name` or `idx`.
+            if msg is None:
+                if ancestor_name not in ancestral_names:
+                    msg = f"unrecognized ancestor {ancestor_name!r}"
+                elif ancestor_idx is None and ancestral_names.count(ancestor_name) > 1:
+                    msg = f"per-string {ancestor_name!r} is ambiguous in this context; use the form '{ancestor_name}:index' to disambiguate, where '{ancestor_name}:1' refers to the most recent ancestor"
+                else:
+                    msg = f"there are not {ancestor_idx} ancestors of the current context with name {ancestor_name!r}"
```
hadia206, line 338: This will run if `ancestor_idx is None` and `ancestral_names.count(ancestor_name)` is `<= 1`
If that's intentional, then `msg` should be updated as currently it'll have `there are not {ancestor_idx} ancestors` as that will be `there are not None ancestors`
If not intentional, then `else` should be `elif ancestor_idx is not None`

pydough/evaluation/evaluate_unqualified.py
```diff
@@ -80,25 +83,18 @@ def _load_column_selection(kwargs: dict[str, object]) -> list[tuple[str, str]] |
...
         for alias, column in columns_arg.items():
-            assert isinstance(alias, str), (
-                f"Expected alias name in `columns` argument to be a string, found {column.__class__.__name__}"
-            )
-            assert isinstance(column, str), (
-                f"Expected column name in `columns` argument to be a string, found {column.__class__.__name__}"
-            )
+            if not isinstance(column, str) and isinstance(alias, str):
+                raise pydough.active_session.error_builder.bad_columns(columns_arg)
             result.append((alias, column))
```
hadia206, line 93: This seems duplicated code to `bad_columns` iterate over columns, check type(s), and raise exception if anything is off.
Only difference is that one will append if all good.

hadia206: Suggestion from AI
```Python
def bad_columns(self, columns: object, reason: str | None = None) -> PyDoughException:
    """
    Creates an exception for when the `columns` to `to_sql` or `to_df` is not valid.
    Args:
        `columns`: The columns argument that caused the error.
        `reason`: Optional extra detail about why it's invalid.
    """
    base = f"Invalid `columns` argument of type {columns.__class__.__name__}"
    if reason:
        base += f": {reason}"
    return PyDoughQDAGException(base)
```

and here

```
if isinstance(columns_arg, list):
    if not columns_arg:
        raise pydough.active_session.error_builder.bad_columns(columns_arg, "list is empty")
    for column in columns_arg:
        if not isinstance(column, str):
            raise pydough.active_session.error_builder.bad_columns(columns_arg, f"found {type(column).__name__} in list")
        result.append((column, column))

elif isinstance(columns_arg, dict):
    if not columns_arg:
        raise pydough.active_session.error_builder.bad_columns(columns_arg, "dictionary is empty")
    for alias, column in columns_arg.items():
        if not isinstance(alias, str):
            raise pydough.active_session.error_builder.bad_columns(columns_arg, f"alias {alias!r} is not a string")
        if not isinstance(column, str):
            raise pydough.active_session.error_builder.bad_columns(columns_arg, f"value {column!r} is not a string")
        result.append((alias, column))

else:
    raise pydough.active_session.error_builder.bad_columns(columns_arg, "expected list or dict")
```

pydough/qdag/collections/augmenting_child_operator.py
```diff
@@ -88,7 +88,6 @@ def get_term(self, term_name: str) -> PyDoughQDAG:
             term = Reference(self.preceding_context, term_name, typ)
         return term
 
-    @cache
```
hadia206, line 91: Why removed?

pydough/unqualified/qualification.py
```diff
@@ -216,6 +208,59 @@ def qualify_binary_operation(
             operator, [qualified_lhs, qualified_rhs]
         )
 
+    def extract_window_per_args(
+        self,
+        per: str,
+        ancestral_names: list[str],
+        context: PyDoughCollectionQDAG,
+        window: UnqualifiedWindow,
+    ) -> tuple[str, int | None]:
```
hadia206, line 217: docstring

pydough/unqualified/unqualified_node.py
```diff
@@ -594,68 +588,21 @@ class UnqualifiedOperator(UnqualifiedNode):
...
+        operands: list[UnqualifiedNode] = [
+            self.coerce_to_unqualified(arg) for arg in args
+        ]
+        if isinstance(self._parcel[0], pydop.ExpressionWindowOperator):
+            return call_window_operator(self._parcel[0], operands, **kwargs)
+        elif isinstance(self._parcel[0], pydop.ExpressionFunctionOperator):
+            return call_function_operator(self._parcel[0], operands, **kwargs)
+        else:
+            raise NotImplementedError(
+                f"Unsupported operator type: {self._parcel[0].__class__.__name__}"
```
hadia206, line 604: should this `PyDoughUnqualifiedException`

tests/test_metadata_errors.py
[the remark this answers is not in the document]
hadia206: should there be an upper limit to suggestions? Top 3-5 is enough.

PR #390, opened 2025-07-09
Add projection pull-up optimization

knassre-bodo:
Rewrote several sections of the optimization passes, along with adding a projection pull-up pass which attempts to ensure function calls happen as late in the plan as possible (also transforms some aggregations to simplify/combine when possible). The list of all the changes:
- Added a `limit` feature to the root node so it can store a limit number (if provided); the projection merging pass can now suck a limit into the root using this feature, thus shrinking the plan and sometimes shrinking the generated SQL by requiring fewer CTEs
- Removed the non-convolution requirement for projection merging: now adjacent projects/roots will always be merged unless they both contain window functions.
- Allowed aggregation nodes to have keys that are NOT column references, since pullup can pull a non-column into the keys
- The pullup pass traverses from the bottom to the top, merging projections into one another, then attempting to merge a project into a parent aggregate/limit/join/filter, then trying to pull columns from the output o a limit/join/filter into a parent. The attempt to merge a project into the parent works unless a column is used in both the output columns and the condition/order keys/partition keys.
- Minor changes to how filter pushdown & certain utilities are written, to avoid certain bugs, and added the capability to run projection merging without join transpose
- Fixed a bug in the SQLGlot generation for `Filter` relational nodes by always generating the subquery
- End of the relational converter pipeline, after column bubbling, is now: (project pullup + column pruning)x2, filter pushdown again, projection merging again (w/o join transpose), bubbling again, and one last column pruning.
- During aggregation handling of projection pullup, performs several kinds of simplifications (added new tests for these):
  - Simplifies aggregation calls on a literal argument or on an aggregation key for a variety of functions. Notable patterns that are simplified:
    - `FUNC(x)` when `x` is a literal just becomes `x` for `MIN`, `MAX`, `ANYTHING`, `AVG`, `MEDIAN`, `QUANTILE`
    - `COUNT(x)` when `x` is a grouping key becomes `COUNT(*) * INTEGER(PRESENT(x))`
    - `NDISTINCT(x)` when `x` is a grouping key becomes `INTEGER(PRESENT(x))`
    - `FUNC(x)` when `x` is a grouping key becomes `x` for `MIN`, `MAX`, `ANYTHING`, `AVG`, `MEDIAN`, `QUANTILE` or `SUM`
    - `SUM(0)` becomes `0`
    - `SUM(NULL)` becomes `NULL`
    - `SUM(1)` becomes `COUNT(*)`
    - `SUM(n)` where `n` is a numeric literal becomes `COUNT(*) * n`
    - `COUNT(NULL)` becomes `0`
    - `COUNT(n)` where `n` is a non-null literal becomes `COUNT(*)`
    - `NDISTINCT(n)` where `n` is a literal becomes `1` if `n` is non-null, and `0` if it is null
  - Tries to merge adjacent aggregations when the top uses a subset of the aggregation keys from the bottom (or the same aggregation keys).
    - `COUNT(*)` on top becomes `ANYTHING(1)` if the keys are the same, or `NDISTINCT(x)` if there is one extra key `x` in the bottom agg not in the top agg.
    - `SUM(SUM(x))` becomes `SUM(x)` and `SUM(COUNT(x))` becomes `COUNT(x)`
    - `MIN(MIN(x))` and `MIN(ANYTHING(x))` become `MIN(x)`
    - `MAX(MAX(x))` and `MAX(ANYTHING(x))` become `MAX(x)`
    - `ANYTHING(ANYTHING(x))` become `ANYTHING(x)`

### DIFF
--- a/pydough/conversion/projection_pullup.py
+++ b/pydough/conversion/projection_pullup.py
@@ -64,7 +64,8 @@ def widen_columns(
     }
 
     # Pull all the columns from each input to the node into the node's output
-    # columns if they are not already in the node's output columns.
+    # columns if they are not already in the node's output columns. Make sure
+    # not to include no-op mappings.
     for input_idx in range(len(node.inputs)):
         input_alias: str | None = node.default_input_aliases[input_idx]
         input_node: RelationalNode = node.inputs[input_idx]
@@ -87,12 +88,13 @@ def widen_columns(
                 new_ref: ColumnReference = ColumnReference(new_name, expr.data_type)
                 node.columns[new_name] = ref_expr
                 existing_vals[expr] = ref_expr
-                substitutions[ref_expr] = new_ref
-            else:
+                if ref_expr != new_ref:
+                    substitutions[ref_expr] = new_ref
+            elif ref_expr != existing_vals[expr]:
                 substitutions[ref_expr] = existing_vals[expr]
 
-    # Return the substitution mapping, without any no-op substitutions
-    return {k: v for k, v in substitutions.items() if k != v}
+    # Return the substitution mapping
+    return substitutions
 
 
 def pull_non_columns(node: Join | Filter | Limit) -> RelationalNode:
@@ -181,7 +183,7 @@ def pull_project_helper(
     )
 
     # Iterate through the columns of the project to see which ones can be
-    # pulled up into the parent, adding them to a substitutions mapping that
+    # pulled up into the parent, dding them to a substitutions mapping that
     # will be used to apply the transformations.
     substitutions: dict[RelationalExpression, RelationalExpression] = {}
     for name, expr in project.columns.items():
@@ -205,7 +207,8 @@ def pull_project_into_join(node: Join, input_index: int) -> None:
     Args:
         `node`: The Join node to pull the Project columns into.
         `input_index`: The index of the input to the Join node that should have
-        its columns pulled up, if it is a project node.
+        its columns pulled up, if it is a project node. This is assumed to be
+        either 0 (for the LHS) or 1 (for the RHS).
     """
 
     # Skip if the input at the specified input is not a Project node.
@@ -480,7 +483,7 @@ def pull_project_into_aggregate(node: Aggregate) -> RelationalNode:
     possible. This transformation is done in-place.
 
     Args:
-        `node`: The Filter node to pull the Project columns into.
+        `node`: The Aggregate node to pull the Project columns into.
     """
     if not isinstance(node.input, Project):
         return node
--- a/pydough/conversion/merge_projects.py
+++ b/pydough/conversion/merge_projects.py
@@ -197,18 +197,17 @@ def merge_adjacent_projects(node: RelationalRoot | Project) -> RelationalNode:
             if isinstance(node, RelationalRoot):
                 # Replace all column references in the root's columns with
                 # the expressions from the child projection.
-                for idx, (name, expr) in enumerate(node.ordered_columns):
-                    new_expr = transpose_expression(expr, child_project.columns)
-                    node.columns[name] = new_expr
-                    node.ordered_columns[idx] = (name, new_expr)
+                node._ordered_columns = [
+                    (name, transpose_expression(expr, node.input.columns))
+                    for name, expr in node.ordered_columns
+                ]
+                node._columns = dict(node.ordered_columns)
                 # Do the same with the sort expressions.
                 for idx, sort_info in enumerate(node.orderings):
                     new_expr = transpose_expression(
                         sort_info.expr, child_project.columns
                     )
-                    node.orderings[idx] = ExpressionSortInfo(
-                        new_expr, sort_info.ascending, sort_info.nulls_first
-                    )
+                    node.orderings[idx].expr = new_expr
                 # Delete the child projection from the tree, replacing it
                 # with its input.
                 node._input = child_project.input
@@ -262,10 +261,11 @@ def merge_adjacent_projects(node: RelationalRoot | Project) -> RelationalNode:
             # If the orderings are the same, pull in the limit into the root.
             # Replace all column references in the root's columns with
             # the expressions from the child projection.
-            for idx, (name, expr) in enumerate(node.ordered_columns):
-                new_expr = transpose_expression(expr, node.input.columns)
-                node.columns[name] = new_expr
-                node.ordered_columns[idx] = (name, new_expr)
+            node._ordered_columns = [
+                (name, transpose_expression(expr, node.input.columns))
+                for name, expr in node.ordered_columns
+            ]
+            node._columns = dict(node.ordered_columns)
             node._orderings = new_orderings
             node._limit = node.input.limit
             # Delete the child projection from the tree, replacing it
--- a/pydough/conversion/relational_converter.py
+++ b/pydough/conversion/relational_converter.py
@@ -1417,7 +1417,7 @@ def optimize_relational_tree(
     Returns:
         The optimized relational root.
     """
-    # Step 1: push filters down as far as possible
+    # Step 1: push filters down as far as possible.
     root._input = push_filters(root.input, set())
 
     # Step 2: merge adjacent projections, unless it would result in excessive
@@ -1432,10 +1432,14 @@ def optimize_relational_tree(
     # operating on already unique data.
     root = remove_redundant_aggs(root)
 
-    # Step 5: re-run projection merging.
+    # Step 5: re-run projection merging since the removal of redundant
+    # aggregations may have created redundant projections that can be deleted.
     root = confirm_root(merge_projects(root))
 
-    # Step 6: re-run column pruning.
+    # Step 6: re-run column pruning after the various steps, which may have
+    # rendered more columns unused. This is done befre the next step to remove
+    # as many column names as possible so the column bubbling step can try to
+    # use nicer names without worrying about collisions.
     root = ColumnPruner().prune_unused_columns(root)
 
     # Step 7: bubble up names from the leaf nodes to further encourage simpler
@@ -1443,21 +1447,31 @@ def optimize_relational_tree(
     # possible.
     root = bubble_column_names(root)
 
-    # Step 8: run projection pullup followed by column pruning 2x.
+    # Step 8: the following pipeline twice:
+    #   A: projection pullup
+    #   B: filter pushdown
+    #   C: column pruning
+    # This is done because pullup will create more opportunities for filter
+    # pushdown, and the two together will create more opportunities for
+    # column pruning, the latter of which will unlock more opportunities for
+    # pullup and pushdown and so on.
     for _ in range(2):
         root = confirm_root(pullup_projections(root))
+        root._input = push_filters(root.input, set())
         root = ColumnPruner().prune_unused_columns(root)
 
-    # Step 9: re-run filter pushdown
-    root._input = push_filters(root.input, set())
-
-    # Step 10: re-run projection merging, without pushing into joins.
+    # Step 9: re-run projection merging, without pushing into joins. This
+    # will allow some redundant projections created by pullup to be removed
+    # entirely.
     root = confirm_root(merge_projects(root, push_into_joins=False))
 
-    # Step 11: re-run column bubbling
+    # Step 10: re-run column bubbling to further simplify the final names of
+    # columns in the output now that more columns have been pruned, and delete
+    # any new duplicate columns that were created during the pullup step.
     root = bubble_column_names(root)
 
-    # Step 12: re-run column pruning.
+    # Step 11: re-run column pruning one last time to remove any columns that
+    # are no longer used after the final round of transformations.
     root = ColumnPruner().prune_unused_columns(root)
 
     return root


pydough/conversion/column_bubbler.py
```diff
@@ -188,16 +188,19 @@ def run_column_bubbling(
             # For aggregate, do the same as projection but run separately for
             # keys and aggregations.
             new_input, input_mapping = run_column_bubbling(node.input, corr_remap)
-            new_keys: dict[str, ColumnReference] = {}
+            new_keys: dict[str, RelationalExpression] = {}
```
knassre-bodo, line 191: Changes to this file and `agg_split.py` are because of the changes to the definition of `aggregate.py`

pydough/relational/relational_nodes/limit.py
```diff
@@ -66,7 +66,7 @@ def to_string(self, compact: bool = False) -> str:
         orderings: list[str] = [
             ordering.to_string(compact) for ordering in self.orderings
         ]
-        return f"LIMIT(limit={self.limit}, columns={self.make_column_string(self.columns, compact)}, orderings=[{', '.join(orderings)}])"
+        return f"LIMIT(limit={self.limit.to_string(compact)}, columns={self.make_column_string(self.columns, compact)}, orderings=[{', '.join(orderings)}])"
```
knassre-bodo, line 69: This makes the stringification of `Limit` nodes more consistent with other nodes.

pydough/relational/relational_nodes/relational_root.py
```diff
@@ -51,6 +53,13 @@ def orderings(self) -> list[ExpressionSortInfo]:
         """
         return self._orderings
 
+    @property
+    def limit(self) -> RelationalExpression | None:
+        """
+        The limit on the number of rows in the final output, if any.
+        """
+        return self._limit
```
knassre-bodo, line 61: This allows pulling a limit into the root, which can simplify the SQLGlot generation & reduce the number of CTEs.

pydough/sqlglot/sqlglot_relational_visitor.py
```diff
@@ -457,23 +457,7 @@ def visit_filter(self, filter: Filter) -> None:
...
-                or "limit" in input_expr.args
-            ):
-                # Check if we already have a where clause or limit. We
-                # cannot merge these yet.
-                # TODO: (gh #151) Consider allowing combining where if
-                # limit isn't present?
-                query = self._build_subquery(input_expr, exprs)
-            else:
-                # Try merge the column sections
-                query = self._merge_selects(exprs, input_expr, find_identifiers(cond))
```
knassre-bodo, line 476: It turns out this is sometimes bugged when window functions are involved

pydough/relational/relational_nodes/aggregate.py
```diff
@@ -78,11 +77,8 @@ def node_copy(
         keys = {}
         aggregations = {}
         for key, val in columns.items():
-            if isinstance(val, ColumnReference):
-                keys[key] = val
-            else:
-                assert isinstance(val, CallExpression), (
-                    "All columns must be references or functions"
-                )
```
knassre-bodo, line 86: This is due to the fact that the keys no longer MUST be column references

tests/test_plan_refsols/agg_orders_by_year_month_just_europe.txt
```diff
@@ -1,12 +1,10 @@
 ROOT(columns=[('year', year), ('month', month), ('num_european_orders', DEFAULT_TO(n_rows, 0:numeric))], orderings=[])
  JOIN(condition=t0.month == t1.month & t0.year == t1.year, type=LEFT, cardinality=SINGULAR_ACCESS, columns={'month': t0.month, 'n_rows': t1.n_rows, 'year': t0.year})
-  AGGREGATE(keys={'month': month, 'year': year}, aggregations={})
-   PROJECT(columns={'month': MONTH(o_orderdate), 'year': YEAR(o_orderdate)})
-    SCAN(table=tpch.ORDERS, columns={'o_orderdate': o_orderdate})
-  AGGREGATE(keys={'month': month, 'year': year}, aggregations={'n_rows': COUNT()})
-   JOIN(condition=t0.o_custkey == t1.c_custkey, type=INNER, cardinality=SINGULAR_FILTER, columns={'month': t0.month, 'year': t0.year})
-    PROJECT(columns={'month': MONTH(o_orderdate), 'o_custkey': o_custkey, 'year': YEAR(o_orderdate)})
```
knassre-bodo, line 8: This project got sucked up into the join, then spat out as a parent project, then sucked into the parent aggregate.

tests/test_plan_refsols/aggregate_anti.txt
```diff
@@ -1,5 +1,5 @@
-ROOT(columns=[('name', s_name), ('num_10parts', DEFAULT_TO(NULL_2, 0:numeric)), ('avg_price_of_10parts', NULL_2), ('sum_price_of_10parts', DEFAULT_TO(NULL_2, 0:numeric))], orderings=[])
- JOIN(condition=t0.s_suppkey == t1.ps_suppkey, type=ANTI, cardinality=SINGULAR_FILTER, columns={'NULL_2': None:unknown, 's_name': t0.s_name})
```
knassre-bodo, line 2: Here, the null literal was ejected out of the join and sucked into the parent root.

tests/test_plan_refsols/aggregate_mixed_levels_simple.txt
```diff
@@ -1,8 +1,7 @@
 ROOT(columns=[('order_key', o_orderkey), ('max_ratio', max_ratio)], orderings=[])
  JOIN(condition=t0.o_orderkey == t1.l_orderkey, type=INNER, cardinality=SINGULAR_ACCESS, columns={'max_ratio': t1.max_ratio, 'o_orderkey': t0.o_orderkey})
   SCAN(table=tpch.ORDERS, columns={'o_orderkey': o_orderkey})
-  AGGREGATE(keys={'l_orderkey': l_orderkey}, aggregations={'max_ratio': MAX(ratio)})
-   PROJECT(columns={'l_orderkey': l_orderkey, 'ratio': l_quantity / ps_availqty})
```
knassre-bodo, line 5: Here, the calculation `l_quantity / ps_availqty` was sucked into the `MAX` call

tests/test_plan_refsols/common_prefix_ad.txt
```diff
@@ -1,16 +1,15 @@
...
+  JOIN(condition=t0.ps_partkey == t1.l_partkey & t0.ps_suppkey == t1.l_suppkey, type=INNER, cardinality=SINGULAR_FILTER, columns={'p_name': t0.p_name, 'ps_availqty': t0.ps_availqty, 'ps_suppkey': t0.ps_suppkey, 'sum_l_quantity': t1.sum_l_quantity})
    FILTER(condition=RANKING(args=[], partition=[ps_suppkey], order=[(ps_availqty):desc_first, (p_name):asc_last], allow_ties=False) == 1:numeric, columns={'p_name': p_name, 'ps_availqty': ps_availqty, 'ps_partkey': ps_partkey, 'ps_suppkey': ps_suppkey})
     JOIN(condition=t0.ps_partkey == t1.p_partkey, type=INNER, cardinality=SINGULAR_FILTER, columns={'p_name': t1.p_name, 'ps_availqty': t0.ps_availqty, 'ps_partkey': t0.ps_partkey, 'ps_suppkey': t0.ps_suppkey})
      SCAN(table=tpch.PARTSUPP, columns={'ps_availqty': ps_availqty, 'ps_partkey': ps_partkey, 'ps_suppkey': ps_suppkey})
      FILTER(condition=p_container == 'WRAP CASE':string, columns={'p_name': p_name, 'p_partkey': p_partkey})
       SCAN(table=tpch.PART, columns={'p_container': p_container, 'p_name': p_name, 'p_partkey': p_partkey})
-   PROJECT(columns={'l_partkey': l_partkey, 'l_suppkey': l_suppkey, 'qty_shipped': DEFAULT_TO(sum_l_quantity, 0:numeric)})
```
knassre-bodo, line 13: This `DEFAULT_TO` call was pulled up through 2 joins into the final root.

tests/test_plan_refsols/correl_14.txt
```diff
@@ -1,18 +1,15 @@
 ROOT(columns=[('n', n)], orderings=[])
- AGGREGATE(keys={}, aggregations={'n': COUNT()})
-  AGGREGATE(keys={'s_suppkey': s_suppkey}, aggregations={})
```
knassre-bodo, line 3: This duo was merged into `AGGREGATE(keys={}, aggregations={'n': NDISTINCT(s_suppkey)})`

tests/test_plan_refsols/correl_18.txt
```diff
@@ -1,12 +1,10 @@
 ROOT(columns=[('n', DEFAULT_TO(sum_n_above_avg, 0:numeric))], orderings=[])
- AGGREGATE(keys={}, aggregations={'sum_n_above_avg': SUM(n_above_avg)})
-  AGGREGATE(keys={'o_custkey': o_custkey, 'o_orderdate': o_orderdate}, aggregations={'n_above_avg': COUNT()})
```
knassre-bodo, line 3: These two were merged into `AGGREGATE(keys={}, aggregations={'sum_n_above_avg': COUNT()})`

tests/test_plan_refsols/count_at_most_100_suppliers_per_nation.txt
```diff
@@ -2,5 +2,5 @@ ROOT(columns=[('name', n_name), ('n_top_suppliers', DEFAULT_TO(count_s_suppkey,
  JOIN(condition=t0.n_nationkey == t1.s_nationkey, type=LEFT, cardinality=SINGULAR_ACCESS, columns={'count_s_suppkey': t1.count_s_suppkey, 'n_name': t0.n_name})
   SCAN(table=tpch.NATION, columns={'n_name': n_name, 'n_nationkey': n_nationkey})
   AGGREGATE(keys={'s_nationkey': s_nationkey}, aggregations={'count_s_suppkey': COUNT(s_suppkey)})
-   LIMIT(limit=Literal(value=100, type=NumericType()), columns={'s_nationkey': s_nationkey, 's_suppkey': s_suppkey}, orderings=[(s_acctbal):asc_last])
+   LIMIT(limit=100:numeric, columns={'s_nationkey': s_nationkey, 's_suppkey': s_suppkey}, orderings=[(s_acctbal):asc_last])
```
knassre-bodo, line 5: This is just t he change to make the stringification more consistent.

tests/test_plan_refsols/multi_partition_access_2.txt
```diff
@@ -9,8 +9,7 @@ ROOT(columns=[('transaction_id', sbTxId), ('name', sbCustName), ('symbol', sbTic
         AGGREGATE(keys={'sbTxCustId': sbTxCustId}, aggregations={})
          JOIN(condition=t0.sbTxCustId == t1.sbTxCustId & t0.sbTxTickerId == t1.sbTxTickerId, type=INNER, cardinality=SINGULAR_ACCESS, columns={'sbTxCustId': t0.sbTxCustId})
           AGGREGATE(keys={'sbTxCustId': sbTxCustId, 'sbTxTickerId': sbTxTickerId}, aggregations={})
-           AGGREGATE(keys={'sbTxCustId': sbTxCustId, 'sbTxTickerId': sbTxTickerId, 'sbTxType': sbTxType}, aggregations={})
```
knassre-bodo, line 12: These two were merged into just the top one

tests/test_plan_refsols/multi_partition_access_6.txt
```diff
@@ -2,60 +2,50 @@ ROOT(columns=[('transaction_id', sbTxId)], orderings=[(sbTxId):asc_first])
  FILTER(condition=n_ticker_type_trans == 1:numeric | n_cust_type_trans == 1:numeric, columns={'sbTxId': sbTxId})
   JOIN(condition=t0.sbTxCustId == t1.sbTxCustId & t0.sbTxType == t1.sbTxType, type=INNER, cardinality=PLURAL_FILTER, columns={'n_cust_type_trans': t0.n_cust_type_trans, 'n_ticker_type_trans': t1.n_ticker_type_trans, 'sbTxId': t1.sbTxId})
    JOIN(condition=t0.sbTxCustId == t1.sbTxCustId, type=INNER, cardinality=PLURAL_FILTER, columns={'n_cust_type_trans': t1.n_cust_type_trans, 'sbTxCustId': t1.sbTxCustId, 'sbTxType': t1.sbTxType})
-    FILTER(condition=n_cust_trans > 1:numeric, columns={'sbTxCustId': sbTxCustId})
-     PROJECT(columns={'n_cust_trans': DEFAULT_TO(sum_n_cust_type_trans, 0:numeric), 'sbTxCustId': sbTxCustId})
-      AGGREGATE(keys={'sbTxCustId': sbTxCustId}, aggregations={'sum_n_cust_type_trans': SUM(n_cust_type_trans)})
-       AGGREGATE(keys={'sbTxCustId': sbTxCustId, 'sbTxType': sbTxType}, aggregations={'n_cust_type_trans': COUNT()})
```
knassre-bodo, line 8: These two were merged into `GGREGATE(keys={'sbTxCustId': sbTxCustId}, aggregations={'sum_n_cust_type_trans': COUNT()})`

tests/test_sql_refsols/tpch_q20_sqlite.sql
```diff
@@ -1,6 +1,6 @@
 WITH _s3 AS (
   SELECT
-    COALESCE(SUM(l_quantity), 0) AS agg_0,
+    SUM(l_quantity) AS sum_l_quantity,
```
knassre-bodo, line 3: This `COALESCE` call got moved later (line 39)

tests/test_sql_refsols/tpch_q20_sqlite.sql
```diff
@@ -9,22 +9,22 @@ WITH _s3 AS (
...
+), _t2 AS (
   SELECT
     COUNT(*) AS n_rows,
     partsupp.ps_suppkey
   FROM tpch.partsupp AS partsupp
   JOIN _s5 AS _s5
     ON _s5.p_partkey = partsupp.ps_partkey
     AND partsupp.ps_availqty > (
-      0.5 * COALESCE(_s5.agg_0, 0)
+      0.5 * COALESCE(COALESCE(_s5.sum_l_quantity, 0), 0)
```
knassre-bodo, line 27: It got moved here, which presents a future opportunity for simplification.

tests/test_sql_refsols/tpch_q15_sqlite.sql
```diff
@@ -9,29 +9,26 @@ WITH _t5 AS (
     l_shipdate < '1996-04-01' AND l_shipdate >= '1996-01-01'
 ), _s1 AS (
   SELECT
-    COALESCE(SUM(l_extendedprice * (
```
knassre-bodo, line 12: This coalesce got moved to line 21

tests/test_sql_refsols/time_threshold_reached_sqlite.sql
```diff
@@ -3,15 +3,14 @@ WITH _t3 AS (
...
-    DATE(sbtxdatetime, 'start of day') AS txn_day
+    sbtxdatetime
   FROM main.sbtransaction
   WHERE
     CAST(STRFTIME('%Y', sbtxdatetime) AS INTEGER) = 2023
 ), _t AS (
   SELECT
     sbtxdatetime,
-    ROW_NUMBER() OVER (PARTITION BY txn_day ORDER BY pct_of_day) AS _w
+    ROW_NUMBER() OVER (PARTITION BY DATE(sbtxdatetime, 'start of day') ORDER BY pct_of_day) AS _w
```
knassre-bodo, line 13: Minor case of better column pruning as a result of pullup. Instead of including `DATE(sbtxdatetime, 'start of day')` as a column then using it later, just uses `DATE(sbtxdatetime, 'start of day')` directly.

tests/test_sql_refsols/func_rank_with_filters_a_ansi.sql
```diff
@@ -1,16 +1,14 @@
...
 )
 SELECT
   a,
   b,
   r
-FROM _t0
+FROM _t1
 WHERE
-  r >= 3
+  b = 0 AND r >= 3
```
juankx-bodo, line 14: Do we want this? 

It seems to me that both querys are not equivalent. We were applying `WHERE b = 0` filter before the `RANK()` function, so `RANK()` only ranked rows `where b = 0` and then it filtered for `r >= 3`. Now, `RANK()` is computed over all rows in the table and then we apply the filter. The value of `r` will be different.

Also, in case we were not using the value of `r`, it seems to me that it would be better to apply the `WHERE b = 0` filter before because there will be less data to rank and we will rank only the relevant subset.

knassre-bodo: The old version was likely created at some point because of other changes to the SQLGlot generation, and we never noticed anything was wrong bc this isn't an e2e test (just generates SQL).

I fixed this because some of the projection pullup changes caused this same bug to be triggered by some other e2e tests.

tests/test_sql_refsols/defog_broker_adv5_ansi.sql
```diff
@@ -47,10 +48,12 @@ WITH _s0 AS (
...
+  sum_sum_sbdpclose / sum_count_sbdpclose AS avg_close,
   max_high,
   min_low,
   (
-    avg_close - LAG(avg_close, 1) OVER (PARTITION BY sbtickersymbol ORDER BY month NULLS LAST)
-  ) / LAG(avg_close, 1) OVER (PARTITION BY sbtickersymbol ORDER BY month NULLS LAST) AS momc
+    (
+      sum_sum_sbdpclose / sum_count_sbdpclose
+    ) - LAG(sum_sum_sbdpclose / sum_count_sbdpclose, 1) OVER (PARTITION BY sbtickersymbol ORDER BY month NULLS LAST)
+  ) / LAG(sum_sum_sbdpclose / sum_count_sbdpclose, 1) OVER (PARTITION BY sbtickersymbol ORDER BY month NULLS LAST) AS momc
```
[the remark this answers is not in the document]
knassre-bodo: Yes, this is one of the occasional minor decreases to readability.

john-sanchez31 (approved):
Looks good! Just a little doubt below

tests/test_sql_refsols/func_rank_with_filters_a_ansi.sql
juankx-bodo: Could be have an e2e test case that exploits this same issue? Maybe another one that do the opposite, generate an SQL like the previous one.

knassre-bodo: We do, elsewhere. They are named something similar to this test.

juankx-bodo:
When I began this review, I was concerned about specific SQL query rewrites such as:
- `COUNT(x)` where `x` is a grouping key transforming into `COUNT(*) * INTEGER(PRESENT(x))`
- `NDISTINCT(x)` where `x` is a grouping key transforming into `INTEGER(PRESENT(x))`
- `COUNT(n)` where n is a non-null literal transforming into `COUNT(*)`
- `COUNT(*)` on a top aggregation becoming `ANYTHING(1)` if keys are identical, or `NDISTINCT(x)` if an additional key `x` exists in the bottom aggregation but not the top.

Do we currently have tests that cover these transformations? I still believe it might be more effective to allow the database engine to handle these specific optimization patterns.

tests/test_sql_refsols/technograph_battery_failure_rates_anomalies_ansi.sql
```diff
@@ -11,7 +11,7 @@ WITH _s7 AS (
 SELECT
   countries.co_name AS country_name,
   products.pr_name AS product_name,
-  ROUND(COALESCE(SUM(COALESCE(_s7.n_rows, 0)), 0) / COUNT(*), 2) AS ir
+  ROUND(COALESCE(COALESCE(SUM(_s7.n_rows), 0), 0) / COUNT(*), 2) AS ir
```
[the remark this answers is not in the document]
knassre-bodo: The `SUM(COALESCE(x, 0))` was optimized out to `COALESCE(SUM(x), 0)`, but the original expression was already inside of **another** `COALESCE`. Down the line, we can optimize by adding a simplification pass that removes the 2nd `COALESCE`.

tests/test_sql_refsols/technograph_year_cumulative_incident_rate_goldcopperstar_ansi.sql
```diff
@@ -19,74 +19,65 @@ WITH _s14 AS (
...
-  years_since_release,
-  cum_ir,
-  pct_bought_change,
-  pct_incident_change,
-  n_devices AS bought,
-  n_incidents AS incidents
-FROM _t0
+  _s15.year - EXTRACT(YEAR FROM CAST(_s14.release_date AS DATETIME)) AS years_since_release,
+  ROUND(
+    SUM(COALESCE(_s15.sum_expr_4, 0)) OVER (ORDER BY _s15.year NULLS LAST ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) / SUM(COALESCE(_s15.sum_n_rows, 0)) OVER (ORDER BY _s15.year NULLS LAST ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW),
```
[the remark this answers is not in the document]
knassre-bodo: That was for the aggregation version; this is a window function (which wasn't affected)

pydough/conversion/column_bubbler.py
```diff
@@ -333,4 +336,4 @@ def bubble_column_names(root: RelationalRoot) -> RelationalRoot:
                     ordering.nulls_first,
                 )
             )
-    return RelationalRoot(new_input, new_ordered_columns, new_orderings)
+    return RelationalRoot(new_input, new_ordered_columns, new_orderings, root.limit)
```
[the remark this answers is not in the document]
knassre-bodo: Yeah might as well. Down the line we can try to pull some of the logic into common helpers (like I did with projection pullup) in a refactoring PR.

pydough/conversion/projection_pullup.py
```diff
@@ -0,0 +1,715 @@
...
+    # A boolean to indicate if any columns were pulled besides no-ops. If this
+    # never becomes true, then we skip the rewrite and return the original.
+    needs_pull: bool = False
+
+    # Iterate through the columns of the node and check if they are column
+    # references or not. If they are not, then we need to pull them up into
+    # the parent projection.
+    for name, expr in node.columns.items():
+        if isinstance(expr, ColumnReference):
+            new_project_columns[name] = ColumnReference(name, expr.data_type)
```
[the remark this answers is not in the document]
knassre-bodo: Because it may not be the **same** column reference (there may be renaming going on where `expr` points to a different column further down).

pydough/conversion/projection_pullup.py
```diff
@@ -0,0 +1,715 @@
...
+
+def pull_project_into_aggregate(node: Aggregate) -> RelationalNode:
+    """
+    Attempts to pull columns from a Project node that is an input to an
+    Aggregate into the inputs of the aggregation calls of the Aggregate, and
+    into the grouping keys. Additionally, simplifies the aggregation calls when
+    possible. This transformation is done in-place.
+
+    Args:
+        `node`: The Filter node to pull the Project columns into.
```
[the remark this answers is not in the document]
knassre-bodo: Good catch. There was a lot of copy-paste going on when I split up the omni-function into 4-5 helpers :P

pydough/conversion/projection_pullup.py
```diff
@@ -0,0 +1,715 @@
...
+    This transformation is done in-place.
+
+    Args:
+        `node`: The Join node to pull the Project columns into.
+        `input_index`: The index of the input to the Join node that should have
+        its columns pulled up, if it is a project node.
+    """
+
+    # Skip if the input at the specified input is not a Project node.
+    if not isinstance(node.inputs[input_index], Project):
```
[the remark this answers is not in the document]
knassre-bodo: I think it's a reasonable assumption.

pydough/conversion/projection_pullup.py
```diff
@@ -0,0 +1,715 @@
...
+    # which columns from the project can be pulled up into the join's output
+    # columns or condition, and modifies the project node in-place to ensure
+    # every column in the project's inputs is available to the current node.
+    substitutions: dict[RelationalExpression, RelationalExpression] = (
+        pull_project_helper(project, node.default_input_aliases[input_index])
+    )
+
+    # Apply the substitutions to the join's condition and output columns.
+    node._condition = apply_substitution(node.condition, substitutions, {})
+    node._columns = {
```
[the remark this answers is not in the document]
knassre-bodo: These have getters, but don't currently have setters. I'd prefer to fix that in a larger refactor, so for now I'll leave alone but take a note.

pydough/relational/rel_util.py
```diff
@@ -751,3 +765,51 @@ def apply_substitution(
...
+    Adds an input name to all column references inside the given expression.
+
+    Args:
+        `expr`: The expression to add the input name to its contents.
+        `input_name`: The input name to add.
+
+    Returns:
+        The expression with the input name added to all contents, if
+        applicable.
+    """
```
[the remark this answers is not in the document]
knassre-bodo: No, since input names can always be either a string or None. If it is None, it replaces whatever string is currently there with None. If it is a string, it replaces whatever is currently there (string or None) with the provided string.

pydough/conversion/filter_pushdown.py
```diff
@@ -48,17 +48,21 @@ def push_filters(
...
     remaining_filters: set[RelationalExpression]
     pushable_filters: set[RelationalExpression]
+    new_input: RelationalNode
     match node:
         case Filter():
             # Add all of the conditions from the filters pushed down this far
             # with the filters from the current node. If there is a window
             # function, materialize all of them at this point, otherwise push
             # all of them further.
+            filters = {transpose_expression(expr, node.columns) for expr in filters}
```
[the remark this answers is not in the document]
knassre-bodo: Because, now that we are also doing filter pushdown later in the pipeline, it is possible that `node.columns` includes expressions like function calls or column renamings, so we need `transpose_expressions` to account for this and convert `filters` (which is expressed in terms of the **outputs** to the current node) to be expressed in terms of the **inputs** to the current node.

PR #395, opened 2025-07-16
Add PageRank testing suite

knassre-bodo:
Adds new set of tests that perform PageRank calculations in PyDough on small sample graphs. Fixes several minor bugs found in extreme edge cases that were uncovered by these tests. The graphs were generates & had their reference PageRank solutions derived via this website: https://pagerank-visualizer.netlify.app/. Also adjusted how the common `PyDoughPandasTest` dataclass is used in three ways:
- Can pass in an optional list of arguments to be passed into the decorated function when called via `*args` (all of the signatures are now `Callable[..., UnqualifiedNode]` instead of `Callable[[], UnqualifiedNode]`).
- Can pass in an optional kwarg `skip_relational` (default False): if True skips using that test when running relational plan tests (useful since most of the pagerank queries are duplicate relational plans with different datasets; the plan only changes when the # of iterations changes).
- Can pass in an optional kwarg `skip_sql` (default False): if True skips using that test when running SQL query text tests (same motivation as `skip_relational`).


pydough/conversion/relational_converter.py
```diff
@@ -1023,7 +1023,7 @@ def translate_calculate(
         # it relative to the input context.
         for name in node.new_expressions:
             name = node.renamings.get(name, name)
-            hybrid_expr: HybridExpr = node.terms[name]
+            hybrid_expr: HybridExpr = node.new_expressions[name]
```
knassre-bodo, line 1026: Bug fix

pydough/conversion/relational_converter.py
```diff
@@ -1416,6 +1416,10 @@ def optimize_relational_tree(
     Returns:
         The optimized relational root.
     """
+
+    # Step 0: prune unused columns.
+    root = ColumnPruner().prune_unused_columns(root)
+
```
knassre-bodo, line 1422: Adding this first makes the rest of the optimization slightly faster for **very** big plans (causes some minor renaming changes to a bunch of existing tests).

pydough/unqualified/qualification.py
```diff
@@ -903,6 +903,7 @@ def split_partition_ancestry(
                 | UnqualifiedOrderBy()
                 | UnqualifiedSingular()
                 | UnqualifiedPartition()
+                | UnqualifiedBest()
```
knassre-bodo, line 906: Bug fix

tests/test_pydough_functions/simple_pydough_functions.py
```diff
@@ -3030,3 +3030,67 @@ def quantile_function_test_4():
         orders_99_percent=QUANTILE(selected_orders.total_price, 0.99),
         orders_max=QUANTILE(selected_orders.total_price, 1.0),
     )
+
+
+def pagerank(n_iters):
```
knassre-bodo, line 3035: Should I move this to a different file?

tests/test_pipeline_pagerank.py
```diff
@@ -0,0 +1,324 @@
...
+                "PAGERANK_B",
+                lambda: pd.DataFrame(
+                    {
+                        "key": [1, 2, 3, 4, 5],
+                        "page_rank": [0.115, 0.455, 0.2, 0.03, 0.2],
+                    }
+                ),
+                "pagerank_b1",
+                skip_relational=True,
+                skip_sql=True,
```
knassre-bodo, line 113: These are skipped so we have exactly one relational/sql test for each `n_iters` value from 0-6, but always have the e2e tests.

tests/test_plan_refsols/common_prefix_a.txt
```diff
@@ -1,9 +1,9 @@
...
  JOIN(condition=t0.r_regionkey == t1.n_regionkey, type=INNER, cardinality=SINGULAR_ACCESS, columns={'n_customers': t1.n_customers, 'n_nations': t1.n_nations, 'r_name': t0.r_name})
   SCAN(table=tpch.REGION, columns={'r_name': r_name, 'r_regionkey': r_regionkey})
-  AGGREGATE(keys={'n_regionkey': n_regionkey}, aggregations={'n_customers': SUM(n_rows), 'n_nations': SUM(n_nations_0)})
-   JOIN(condition=t0.n_nationkey == t1.c_nationkey, type=INNER, cardinality=SINGULAR_ACCESS, columns={'n_nations_0': t0.n_nations_0, 'n_regionkey': t0.n_regionkey, 'n_rows': t1.n_rows})
-    PROJECT(columns={'n_nationkey': n_nationkey, 'n_nations_0': 1:numeric, 'n_regionkey': n_regionkey})
+  AGGREGATE(keys={'n_regionkey': n_regionkey}, aggregations={'n_customers': SUM(n_rows), 'n_nations': SUM(n_nations)})
+   JOIN(condition=t0.n_nationkey == t1.c_nationkey, type=INNER, cardinality=SINGULAR_ACCESS, columns={'n_nations': t0.n_nations, 'n_regionkey': t0.n_regionkey, 'n_rows': t1.n_rows})
+    PROJECT(columns={'n_nationkey': n_nationkey, 'n_nations': 1:numeric, 'n_regionkey': n_regionkey})
```
knassre-bodo, line 6: Changes caused by the extra column pruning earlier in the optimizer.

tests/gen_data/init_pagerank.sql
```diff
@@ -0,0 +1,16 @@
+-- Custom SQL schema to initialize a custom PageRank database with tables for
```
juankx-bodo, line 1: Other init_X.sql files (except for tests/gen_data/init_technograph.sql) initialize the data directly within their SQL scripts. Technograph, does this in gen_technograph.py. Given this, it makes me wonder: Why are we initializing this test data within tests/conftest.py? An advantage to handling this in SQL files directly is that it simplifies the process of setting up that databases for manual testing.

Not sure how relevant is this for pagerank but it is just an observation that we should try to keep a similar way to initialize the test environments.

juankx-bodo: Considering PageRank involves eight separate data sets, the SQL-based initialization might not be the best fit here. However, we may think about doing it like technograph and, if it is possible, have an easy way to manually create these databases for manual testing.

tests/conftest.py
```diff
@@ -425,3 +425,167 @@ def sqlite_technograph_connection() -> DatabaseContext:
...
+    """
+    # Setup the directory to be the main PyDough directory.
+    base_dir: str = os.path.dirname(os.path.dirname(__file__))
+
+    # The configurations for the pagerank databases. Each tuple contains:
+    # - The name of the database.
+    # - The number of nodes n in the graph.
+    # - The edges in the graph as a list of tuples (src, dst), assuming the
+    #   nodes are numbered from 1 to n.
+    pagerank_configs = [
```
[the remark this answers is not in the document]
knassre-bodo: Might as well.

tests/test_pydough_functions/simple_pydough_functions.py
knassre-bodo: Fair point. I also was considering moving it directly into the same file as the tests & using `pydough.from_string`. Thoughts? (cc @juankx-bodo)

knassre-bodo: > Why do you want to move this function to a new file?

Originally considered it so that it would be separate from the extremely-bloated (& generic) `simple_pydough_functions.py`.

The idea is if we use `pydough.from_string` we can skip the function alltogether and just put the string inside the testing params (you'll see some examples in some upcoming PRs). I'll leave it be for now since this one is kinda complicated.

PR #396, opened 2025-07-19
Add relational expression simplification with predicate inference

knassre-bodo:
Adds a simplification pass in the relational optimizer that simplifies relational expressions using predicate inference & constant folding. Built via a relational visitor which modifies the relational nodes in place by invoking a relational shuttle that transforms the expressions, both of which propagate information via `PredicateSet` objects which contain information about whether an expression matches certain criteria (positive, non-negative, non-null).

Also updates the testing infrastructure in several ways:
- `PyDoughPandasTest` can now take in a string instead of a function (evaluated via `pydough.from_string`).
- The `args` argument to `PyDoughPandasTest` is replaced with `kwargs`, passed in as keyword arguments to functions or as an environment to `pydough.from_string`.


pydough/conversion/relational_converter.py
```diff
@@ -1531,6 +1540,9 @@ def convert_ast_to_relational(
     raw_result: RelationalRoot = postprocess_root(node, columns, hybrid, output)
 
     # Invoke the optimization procedures on the result to clean up the tree.
-    optimized_result: RelationalRoot = optimize_relational_tree(raw_result, configs)
+    additional_shuttles: list[RelationalExpressionShuttle] = []
```
knassre-bodo, line 1543: This list will be used down the lines for when additional configurations are set that allow more specific/advanced transformations to happen during simplification.

tests/test_pipeline_defog_custom.py
```diff
@@ -1587,6 +1610,189 @@ def get_day_of_week(
...
+        pytest.param(
+            PyDoughPandasTest(
+                "result = Broker.CALCULATE("
+                " s00 = DEFAULT_TO(None, 0) == 0,"  # -> True
+                " s01 = DEFAULT_TO(None, 0) != 0,"  # -> False
+                " s02 = DEFAULT_TO(None, 0) >= 0,"  # -> True
+                " s03 = DEFAULT_TO(None, 0) > 0,"  # -> False
+                " s04 = DEFAULT_TO(None, 0) <= 0,"  # -> True
+                " s05 = DEFAULT_TO(None, 0) < 0,"  # -> False
+                " s06 = DEFAULT_TO(None, 0) == None,"  # -> None
```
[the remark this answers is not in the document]
knassre-bodo: `x == NULL` is always NULL, the same goes for other operators. You can test it in SQL. Some dialects have "null safe" variants (for example, in Snowflake you can do `EQUAL_NULL(1, NULL)` which is False, or `EQUAL_NULL(NULL, NULL)` which is True)

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+    def reset(self) -> None:
+        self.stack = []
+
+    def visit_literal_expression(
+        self, literal_expression: LiteralExpression
+    ) -> RelationalExpression:
+        output_predicates: PredicateSet = PredicateSet()
+        if literal_expression.value is not None:
+            output_predicates.not_null = True
+            if isinstance(literal_expression.value, (int, float)):
```
hadia206, line 230: what about booleans? 0/1

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+                if literal_expression.value >= 0:
+                    output_predicates.not_negative = True
+                    if literal_expression.value > 0:
+                        output_predicates.positive = True
```
hadia206, line 234: Should we have a default or fallback for other types? (string, boolean, dates, ...)

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+
+    @no_group_aggregate.setter
+    def no_group_aggregate(self, value: bool) -> None:
+        """
+        Sets whether the shuttle is handling a no-group-aggregate.
+        """
+        self._no_group_aggregate = value
+
+    def reset(self) -> None:
+        self.stack = []
```
hadia206, line 222: should it also reset `_input_predicates` and `_no_group_aggregate`?

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+    def input_predicates(self, value: dict[RelationalExpression, PredicateSet]) -> None:
+        """
+        Sets the input predicates for the shuttle.
+        """
+        self._input_predicates = value
+
+    @property
+    def no_group_aggregate(self) -> bool:
+        """
+        Returns whether the shuttle currently a handling no-group-aggregate.
```
hadia206, line 210:
```suggestion
        Returns whether the shuttle is currently handling a no-group-aggregate.
```

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+                output_predicates.not_null = True
+
+        match expr.op:
+            case pydop.COUNT | pydop.NDISTINCT:
+                # COUNT(n), COUNT(*), and NDISTINCT(n) are guaranteed to be
+                # non-null and non-negative.
+                output_predicates.not_null = True
+                output_predicates.not_negative = True
+
+                # The output if COUNT(*) is positive if unless doing a
```
hadia206, line 320:
```suggestion
                # The output of COUNT(*) is positive unless doing a
```

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+                # no-groupby aggregation. Same goes for calling COUNT or
+                # NDISTINCT ona non-null column.
```
hadia206, line 322:
```suggestion
                # NDISTINCT on a non-null column.
```

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+            The simplified expression with the predicates updated based on the
+            simplification rules. The predicates for the output are placed on
+            the stack.
+        """
+        output_expr: RelationalExpression = expr
+        output_predicates: PredicateSet = PredicateSet()
+        union_set: PredicateSet = PredicateSet.union(arg_predicates)
+        intersect_set: PredicateSet = PredicateSet.intersect(arg_predicates)
+
+        # If the call has null propagating rules, all of hte arguments are
```
hadia206, line 307:
```suggestion
        # If the call has null propagating rules, all of the arguments are
```

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+                # COUNT(n), COUNT(*), and NDISTINCT(n) are guaranteed to be
+                # non-null and non-negative.
+                output_predicates.not_null = True
+                output_predicates.not_negative = True
+
+                # The output if COUNT(*) is positive if unless doing a
+                # no-groupby aggregation. Same goes for calling COUNT or
+                # NDISTINCT ona non-null column.
+                if not no_group_aggregate:
+                    if len(expr.inputs) == 0 or arg_predicates[0].not_null:
```
hadia206, line 324: accessing `arg_predicates[0]`, is it guaranteed that it's not empty list?

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+                elif (
+                    expr.op == pydop.COUNT
+                    and len(expr.inputs) == 1
+                    and arg_predicates[0].not_null
+                ):
+                    if not no_group_aggregate:
+                        output_predicates.positive = True
+                    output_expr = CallExpression(pydop.COUNT, expr.data_type, [])
+
+            # All of these operators are non-null aor non-negative if their
```
hadia206, line 338:
```suggestion
            # All of these operators are non-null or non-negative if their
```

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+            # first argument is.
```
hadia206, line 339: is what? non-null or null?

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+                pydop.SUM
+                | pydop.AVG
+                | pydop.MIN
+                | pydop.MAX
+                | pydop.ANYTHING
+                | pydop.MEDIAN
+                | pydop.QUANTILE
+            ):
+                output_predicates |= arg_predicates[0] & PredicateSet(
+                    not_null=True, not_negative=True
```
hadia206, line 350: what will happen for the `positive` flag in this case?

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+            # folded if the inputs are string literals. The boolean-returning
+            # operators are always non-negative.
+            case pydop.LOWER:
+                if isinstance(expr.inputs[0], LiteralExpression) and isinstance(
+                    expr.inputs[0].value, str
+                ):
+                    output_expr = LiteralExpression(
+                        expr.inputs[0].value.lower(), expr.data_type
+                    )
+            case pydop.UPPER:
```
hadia206, line 442: Add a comment clarifying why no predicates are set

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+
+            # X & Y is False if any of the arguments are False-y literals, and True
+            # if all of the arguments are Truth-y literals.
+            case pydop.BAN:
+                if any(
+                    isinstance(arg, LiteralExpression) and arg.value in [0, False, None]
+                    for arg in expr.inputs
+                ):
+                    output_expr = LiteralExpression(False, expr.data_type)
+                if all(
```
hadia206, line 567: why not `elif all(...`?

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+            # X | Y is True if any of the arguments are Truth-y literals, and False
+            # if all of the arguments are False-y literals.
+            case pydop.BOR:
+                if any(
+                    isinstance(arg, LiteralExpression)
+                    and arg.value not in [0, False, None]
+                    for arg in expr.inputs
+                ):
+                    output_expr = LiteralExpression(True, expr.data_type)
+                if all(
```
hadia206, line 584: Same why not `elif all(...`?

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+                if (
+                    isinstance(expr.inputs[0], LiteralExpression)
+                    and expr.inputs[0].value is not None
+                ):
+                    output_expr = LiteralExpression(
+                        not bool(expr.inputs[0].value), expr.data_type
+                    )
+                    output_predicates.positive = not bool(expr.inputs[0].value)
+                output_predicates.not_negative = True
+                pass
```
hadia206, line 603: Is this `pass` needed?

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+                            case (x, y, pydop.GRT) if isinstance(
+                                x, (int, float, str, bool)
+                            ) and isinstance(y, (int, float, str, bool)):
+                                output_expr = LiteralExpression(x > y, expr.data_type)  # type: ignore
+                            case (x, y, pydop.GEQ) if isinstance(
+                                x, (int, float, str, bool)
+                            ) and isinstance(y, (int, float, str, bool)):
+                                output_expr = LiteralExpression(x >= y, expr.data_type)  # type: ignore
+
+                    case _:
```
hadia206, line 668: Let's add a comment, "no simplification is applied" or something similar

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+
+            # RELCOUNT is always non-negative, but it is only non-null if there
+            # is no frame, and positive if there is no frame and the first
+            # argument is non-null.
+            case pydop.RELCOUNT:
+                if no_frame:
+                    output_predicates.not_null = True
+                    if arg_predicates[0].not_null:
+                        output_predicates.positive = True
+                output_predicates.not_negative = True
```
hadia206, line 791: should it have default case for unsupported cases?
```
case _:
            pass
```

tests/test_pipeline_defog_custom.py
```diff
@@ -1587,6 +1610,189 @@ def get_day_of_week(
...
+                        "s06": [1],
+                        "s07": [0],
+                        "s08": [0],
+                        "s09": [0],
+                    }
+                ),
+                "simplification_3",
+            ),
+            id="simplification_3",
+        ),
```
hadia206, line 1795: Other tests to include
- Tests for ADD, MUL, DIV simplification outside ABS and DEFAULT_TO
- Test multiplication by negative or zero explicitly to confirm positive/not-negative predicates.
- Tests for COUNT(*), COUNT(non-null column), and their positive/non-null predicates  outside ABS and DEFAULT_TO
- Logical operators with mixed literal and non-literal inputs
- XOR (BXR) and LIKE operators
- Tests for Window call operators with and without frames
- Tests for aggregate functions
- Tests where expressions are nested multiple levels to verify recursive simplification works correctly. Example: `ABS(ADD(COUNT(customers), DEFAULT_TO(None, 5))).`
- Empty inputs, zero-length strings in string functions.
- `DEFAULT_TO` with more than two arguments.
- Boolean operators with `None` or unexpected values.

hadia206 (commented):
Overall code seems to do what the PR description is saying.
Please see my comments below, I'm mainly looking for some more tests.

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+            positive=self.positive and not other.positive,
+        )
+
+    @staticmethod
+    def union(predicates: list["PredicateSet"]) -> "PredicateSet":
+        """
+        Computes the union of a list of predicate sets.
+        """
+        result: PredicateSet = PredicateSet()
+        for pred in predicates[1:]:
```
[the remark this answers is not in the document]
knassre-bodo: Whoops, good catch thanks.

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+    @staticmethod
+    def intersect(predicates: list["PredicateSet"]) -> "PredicateSet":
+        """
+        Computes the intersection of a list of predicate sets.
+        """
+        result: PredicateSet = PredicateSet()
+        if len(predicates) == 0:
+            return result
+        else:
+            result |= predicates[0]
```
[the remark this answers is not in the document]
knassre-bodo: Because I don't want the `intersect` operator to actually mutate any of the input sets (since everything is mutable). If I do `result |= predicates[0]` then in the next section do `result &= pred`, this will mutate both result and `predicates[0]`, which is a problem if `predicates[0]` is being used elsewhere.

pydough/conversion/relational_simplification.py
```diff
@@ -0,0 +1,994 @@
...
+        return result
+
+    @staticmethod
+    def intersect(predicates: list["PredicateSet"]) -> "PredicateSet":
+        """
+        Computes the intersection of a list of predicate sets.
+        """
+        result: PredicateSet = PredicateSet()
+        if len(predicates) == 0:
+            return result
```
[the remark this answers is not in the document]
knassre-bodo: This is saying if we call it on an empty list of sets, the output assumption is that we know nothing.

pydough/conversion/relational_simplification.py
knassre-bodo: Good point

pydough/conversion/relational_simplification.py
knassre-bodo: We start with the default predicate set (we know nothing to be true) and if the value happens to be in these cases then we add predicates to it. The default is we don't enter these cases so we just return the empty predicate set.

pydough/conversion/relational_simplification.py
knassre-bodo: Yess because we assume `len(expr.inputs)` is the same as `len(arg_predicates)`.

pydough/conversion/relational_simplification.py
knassre-bodo: non-null or non-negative.
E.g. if all of their arguments are non-null -> output is non-null
& if all of their arguments are non-negative -> output is non-negative

pydough/conversion/relational_simplification.py
knassre-bodo: The default is False. In this case, by doing `arg_predicates[0] & PredicateSet(...)` I'm saying only keep the predicates from `arg_predicates` that are also set in the other set (not_null & not_negative) -> treating like a mask

pydough/conversion/relational_simplification.py
knassre-bodo: Good point

tests/test_pipeline_defog_custom.py
knassre-bodo: > Tests for ADD, MUL, DIV simplification outside ABS and DEFAULT_TO

Sure thing

> Tests for COUNT(*), COUNT(non-null column), and their positive/non-null predicates outside ABS and DEFAULT_TO

There isn't really a way to test their positive/non-null predicates w/o `ABS` / `DEFAULT_TO`. The point of those two is that they only get simplified correctly if the predicates to their arguments are correct.

> Logical operators with mixed literal and non-literal inputs

We have some of these already

> Tests where expressions are nested multiple levels to verify recursive simplification works correctly

We have some of those already: `MONOTONIC(1, LENGTH('foo'), COUNT(customers))`, `ABS(DEFAULT_TO(AVG(ABS(DEFAULT_TO(LENGTH(customers.name), 0))), 0))`

knassre-bodo: I think I got most of the ones you suggested

pydough/conversion/relational_simplification.py
hadia206: Let's clarify the comment then

hadia206 (approved):
Thanks Kian

john-sanchez31 (approved):
Looks fine

PR #398, opened 2025-07-24
Fixing bug with cross of two partition nodes with correlates

knassre-bodo:
A bug was brought to our attention with code in this pattern:

```py
def double_cross():
    # For each of the first 10 weeks of orders, identify the ratio between the
    # cumulative number of lines returned via train versus the cumulative numbe
    # of urgent orders with status F. Also return the numbers of orders/lines in
    # each week.
    global_info = TPCH.CALCULATE(min_date=MIN(orders.order_date))
    order_week_info = (
        global_info.orders.CALCULATE(ord_wk=DATEDIFF("week", min_date, order_date))
        .WHERE((ord_wk < 10) & (order_status == "F") & (order_priority == "1-URGENT"))
        .PARTITION(name="weeks", by=ord_wk)
        .CALCULATE(ord_wk, n_orders=COUNT(orders))
    )
    line_week_info = (
        global_info.lines.CALCULATE(line_wk=DATEDIFF("week", min_date, receipt_date))
        .WHERE(
            (line_wk < 10)
            & (return_flag == "R")
            & (YEAR(receipt_date) == 1992)
            & (ship_mode == "RAIL")
        )
        .PARTITION(name="weeks", by=line_wk)
        .CALCULATE(line_wk, n_lines=COUNT(lines))
    )
    return (
        order_week_info.CROSS(line_week_info)
        .WHERE(ord_wk == line_wk)
        .CALCULATE(
            wk=ord_wk,
            n_lines=n_lines,
            n_orders=n_orders,
            lpo=ROUND(
                RELSUM(n_lines, by=line_wk.ASC(), cumulative=True)
                / RELSUM(n_orders, by=ord_wk.ASC(), cumulative=True),
                4,
            ),
        )
        .ORDER_BY(wk.ASC())
    )
```

The offending logic was due to a confluence of factors in how partition is handled by qualification & hybrid conversion (specifically, the creation of correlated references) which were previously hidden bugs until the use of `CROSS`. Various patches were implemented to fix these bugs, verified by adding the test above to the suite.

### DIFF
--- a/pydough/unqualified/qualification.py
+++ b/pydough/unqualified/qualification.py
@@ -1016,7 +1016,7 @@ class Qualifier:
             unqualified_parent, None
         )
         qualified_parent: PyDoughCollectionQDAG = self.qualify_collection(
-            unqualified_parent, context, False, is_cross
+            unqualified_parent, context, is_child, is_cross
         )
         qualified_child: PyDoughCollectionQDAG = self.qualify_collection(
             unqualified_child, qualified_parent, True, is_cross


pydough/unqualified/qualification.py
```diff
@@ -1016,7 +1016,7 @@ def qualify_partition(
             unqualified_parent, None
         )
         qualified_parent: PyDoughCollectionQDAG = self.qualify_collection(
-            unqualified_parent, context, True, is_cross
+            unqualified_parent, context, is_child, is_cross
```
knassre-bodo, line 1019: This is the main & most important change: when qualifying the parent of the partition, we don't consider it a child **unless** the entire partition is a child of something. This is important for cases like `xxx.CROSS(a.b.c.PARTITION(...))` because we don't want `a` to be considered a child in such a scenario (it isn't a child of `xxx`, it is a **descendant** of `xxx`)

pydough/conversion/hybrid_translator.py
```diff
@@ -866,6 +866,7 @@ def make_hybrid_correl_expr(
         back_expr: BackReferenceExpression,
         collection: PyDoughCollectionQDAG,
         steps_taken_so_far: int,
+        down_shift: int,
```
knassre-bodo, line 869: This was a supporting change after the fix to qualification.py, required to get the number of levels correct when doing `CORREL(BACK(n).x)` in the `double_cross` example.

pydough/unqualified/qualification.py
hadia206: Ugh. That make sense.

hadia206 (approved):
Thanks Kian

Issue #384, opened 2025-07-02
Add PyDough support for PostgreSQL dialect


knassre-bodo:
What is required:
- Adding new Database enums & connector support.
- Updating any bindings (if needed) to work for MySQL/SingleStore functions.
- Ensuring there are comprehensive SQL generation unit tests for the dialect that cover the functions.
- Adding test infrastructure that can spin up / tear down a PostgreSQL server (both locally for pytest & in CI), along with the same conditional running as described in #296 
- User documentation for how to use PyDough with MySQL/SingleStore.

Note: the SQLGlot MySQL dialect should work for generating SingleStore queries.

PR #400, opened 2025-07-29
Add support for PostgreSQL dialect

juankx-bodo:
This PR is for [Add PyDough support for PostgreSQL dialect #384](https://github.com/bodo-ai/PyDough/issues/384)


pydough/sqlglot/transform_bindings/postgres_transform_bindings.py
```diff
@@ -108,9 +108,19 @@ def convert_get_part(
         result: SQLGlotExpression = sqlglot_expressions.SplitPart(
             this=string_expr,  # string expression
             delimiter=delimiter_expr,  # delimiter
-            part_index=index_expr,  # position (1-based)
+            part_index=sqlglot_expressions.Case(  # position (1-based)
+                ifs=[
+                    sqlglot_expressions.If(
+                        this=sqlglot_expressions.EQ(
+                            this=index_expr,
+                            expression=sqlglot_expressions.Literal.number(0),
```
juankx-bodo, line 116: What happens if the index is negative?

demos/notebooks/PG_TPCH.ipynb
```diff
@@ -2,86 +2,217 @@
...
+    "> You can connect to your **own Postgres database** using your credentials — for example, if you have **PgAdmin** or another local/remote server running.\n",
+    ">\n",
+    "> ---\n",
+    ">\n",
+    "> ### 2️⃣ Docker Image (TPC-H Database)\n",
+    ">\n",
+    "> You can also test with our **pre-built Postgres TPC-H database** available on **Docker Hub**.\n",
+    ">\n",
+    "> #### 📋 Requirements\n",
+    "> - Make sure you have **Docker Desktop** installed and running.\n",
```
juankx-bodo, line 23: You don't really need Docker Desktop. All you need is Docker or Podman to be able to run the container using that docker image.

pydough/sqlglot/transform_bindings/base_transform_bindings.py
```diff
@@ -1400,11 +1400,23 @@ def apply_datetime_offset(
...
+            )
+        elif amt < 0:
+            amt *= -1
+            new_expr = sqlglot_expressions.DateSub(
+                this=base,
+                expression=sqlglot_expressions.convert(amt),
+                unit=sqlglot_expressions.Var(this=unit.value),
+            )
+        else:
+            new_expr = base
```
juankx-bodo, line 1418: Is base in the same format? timestamp vs unit inteval?

pydough/sqlglot/transform_bindings/postgres_transform_bindings.py
```diff
@@ -0,0 +1,608 @@
...
+        return sqlglot_expressions.Cast(
+            this=args[0], to=sqlglot_expressions.DataType.build("INTEGER")
+        )
+
+    def convert_lpad(
+        self, args: list[SQLGlotExpression], types: list[PyDoughType]
+    ) -> SQLGlotExpression:
+        result: SQLGlotExpression = sqlglot_expressions.Pad(
+            this=sqlglot_expressions.Cast(
+                this=args[0], to=sqlglot_expressions.DataType.build("TEXT")
```
juankx-bodo, line 589: Any special reason why to use `TEXT` instead of `VARCHAR`?

tests/conftest.py
```diff
@@ -920,6 +921,165 @@ def mysql_params_tpch_db_context(
...
+    - POSTGRES_DB: The name of the TPCH database in Postgres.
+"""
+
+
+@pytest.fixture(scope="session")
+def postgres_docker_setup() -> None:
+    """Set up the Postgres Docker container for testing."""
+    try:
+        if not is_ci():
+            if container_exists(POSTGRES_DOCKER_CONTAINER):
```
juankx-bodo, line 963: Maybe we may try to test if the connection to the engine is available before doing all this docker setup. Service could be already available.
Also, we may try to get all the connection variables (`POSTGRES_USER`, `POSTGRES_USER`, `POSTGRES_HOST`, `POSTGRES_PORT` & `POSTGRES_DB `) from env with a default value: `os.getenv("VAR_NAME", "default_value")`. We don't really need user provided `POSTGRES_USER` and `POSTGRES_PASSWORD` if we setup the container.

Something like:
- get all env vars with a default value in case they are not provided
- try to connect
- if connection is not available then try to setup the container

We need to terminate the container only if we started it. No need to be running/removing the container for each test, service may be already available.

tests/conftest.py
```diff
@@ -920,6 +921,165 @@ def mysql_params_tpch_db_context(
...
+            else:
+                subprocess.run(
+                    [
+                        "docker",
+                        "run",
+                        "-d",
+                        "--name",
+                        POSTGRES_DOCKER_CONTAINER,
+                        "-e",
+                        f"POSTGRES_PASSWORD={os.getenv('POSTGRES_PASSWORD')}",
```
juankx-bodo, line 977: No `POSTGRES_USER` is provided to the container but we use it from env in the connection.

tests/conftest.py
```diff
@@ -742,6 +752,23 @@ def container_is_running(name: str) -> bool:
...
+- `MYSQL_PASSWORD`: The password for MySQL.
+"""
+
+
+@pytest.fixture(scope="session")
+def require_mysql_env() -> None:
+    """
+    Checks whether all required MySQL environment variables are set.
+    """
+    if not all(os.getenv(var) is not None for var in MYSQL_ENVS):
```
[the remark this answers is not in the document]
juankx-bodo: No `MYSQL_USERNAME` env var is provided to the container. 
What happens with the container removal if it is CI? We should remove the container only if we created it.
Maybe we may try to test if the connection to the engine is available before doing all this docker setup. Service could be already available. See the full comment in the postgres version of this function.

tests/conftest.py
```diff
@@ -920,6 +921,165 @@ def mysql_params_tpch_db_context(
...
+    """
+    This fixture is used to connect to the Postgres TPCH database using
+    a connection object.
+    Returns a DatabaseContext for the Postgres TPCH database.
+    """
+    import psycopg2
+
+    postgres_user: str | None = os.getenv("POSTGRES_USER")
+    postgres_password: str | None = os.getenv("POSTGRES_PASSWORD")
+    postgres_db: str = POSTGRES_DB
```
juankx-bodo, line 1027: We may try to read this from env and use these defaults values in case other values are not provided. This approach works if we accept that an available service may already exist.

tests/conftest.py
```diff
@@ -920,6 +921,165 @@ def mysql_params_tpch_db_context(
...
+        dbname=postgres_db,
+        user=postgres_user,
+        password=postgres_password,
+        host=postgres_host,
+        port=postgres_port,
+    )
+    connection.autocommit = True  # It avoids getting stuck when DROP/CREATE
+    # Loads the defog data into the Postgres engine.
+    base_dir: str = os.path.dirname(os.path.dirname(__file__))
+    path: str = os.path.join(base_dir, "tests/gen_data/init_defog_postgres.sql")
```
juankx-bodo, line 1040: Why is this required? Why don't put this file in `/docker-entrypoint-initdb.d/`. We already have defog scripts there, in case we want to use this version then we can put this file directly in the image to replace the others.

tests/conftest.py
```diff
@@ -920,6 +921,165 @@ def mysql_params_tpch_db_context(
...
+    cursor.close()
+
+    return load_database_context(
+        "postgres",
+        connection=connection,
+    )
+
+
+@pytest.fixture
+def postgres_params_tpch_db_context(
```
juankx-bodo, line 1058: Do we really need 2 different fixtures (`postgres_conn_db_context` and `postgres_params_tpch_db_context`)?
Defog databases are created in their own schema. All schemas may be available in the same pydough_tests database.

pydough/sqlglot/override_canonicalize.py
```diff
@@ -0,0 +1,67 @@
...
+        return expression
+
+    return exp.replace_tree(expression, _canonicalize)
+
+
+def fix_expression_type(node: exp.Expression, dialect: Dialect) -> exp.Expression:
+    if dialect.__class__.__name__ != "Postgres":
+        return node
+
+    if isinstance(node, exp.Sub) and (
```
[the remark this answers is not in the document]
john-sanchez31: It's not using `AGE` at all. So the fix was not needed

tests/gen_data/init_defog_postgres.sql
```diff
@@ -0,0 +1,1060 @@
+-- Create schemas
```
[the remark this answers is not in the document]
john-sanchez31: I changed the image and those files aren't there, to be consistent I managing the defog init files the same way I managed MySQL.

tests/test_pipeline_postgres.py
```diff
@@ -0,0 +1,557 @@
...
+    """
+    return request.param
+
+
+@pytest.fixture
+def defog_postgres_test_data(
+    defog_custom_pipeline_test_data: PyDoughPandasTest,  # noqa: F811
+) -> PyDoughPandasTest:
+    """
+    Modify reference solution data for some Defog queries.
```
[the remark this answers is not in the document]
john-sanchez31: The `GETPART` function has a bug (in the base implementation) that needs to be fixed. That's why is changed here to match the correct expected answer for the dialects. For `week_offset` I think there was a bug related to Snowflake kind of similar so this change fixes those (not sure of the last one).

pydough/sqlglot/transform_bindings/postgres_transform_bindings.py
john-sanchez31: I think is the required argument type for this function in Postgres, besides that there is not a particularly reason

tests/conftest.py
john-sanchez31: This fixture is run for every session or command not for every individual test. The way it works, before this `require_postgres_env` make sure that the ENV is set. For ci this part doesn't run and ask if the container is already running or stopped (in this case this fixture starts it). If there is not container it's created. Finally test for connection to Postgres for every case.

tests/conftest.py
john-sanchez31: Don't sure what do you mean. `postgres_conn_db_context` uses a connection and `postgres_params_tpch_db_context` uses the params to make the connection. The reason is to test both ways to connect

tests/conftest.py
juankx-bodo: Yes, the fixture run for every session, but I mean, do we really need to create/delete the container for each test session? If you are testing with mySQL or postgreSQL you may want to have the container running for the whole time. 

This seems to me like start/stop a database server for each test session. Worst than that, starting a clean brand new server and creating everything for each session.

john-sanchez31: It doesn't do that way, creates the container only when not available, removes the container when the connection fails otherwise, the container will be reuse every time.

PR #401, opened 2025-08-04
Add PyDough metadata for masked table columns

knassre-bodo:
Adds a new metadata type `masked table column`, a subclass of `table column` representing data that is stored in the underlying data in some kind of encrypted manner but where the metadata stores additional information such as how to decrypt the data. These new metadata column types have the following additional properties:
- `unprotect protocol`: a Python format string containing SQL text that, when the value is injected, indicates how to decrypt the data.
- `protect protocol`: a Python format string containing SQL text that, when a value is injected, indicates how to encrypt the value in the same manner as the column was encrypted, so the unprotect protocol will reverse it.
- `server masked`: optional boolean (default False) which, if True, indicates that information about the encryption/decryption scheme is available in a server that can be queried to rewrite/optimize predicates to avoid unmasking.

For example, suppose a string column `name` is "encrypted" by uppercasing it, and "decrypted" by lowercasing it, and does not have a server. This would be the following metadata:
```json
{
  "name": "name",
  "type": "masked table column",
  "column name": "c_name",
  "data type": "string",
  "unprotect protocol": "UPPER({})",
  "protect protocol": "LOWER({})",
  "server masked": false
}
```

So for instance, when reading the data from the table, instead of placing `c_name` in the SELECT clause we would place `LOWER(c_name)` in the SELECT clause so it is "decrypted" (since the data int he table was uppercase). Conversely, if we wish to do a filter on `name == "john smith"`, instead of doing `LOWER(c_name) == "john smith"` we could use the protect protocol to rewrite as `c_name == UPPER("john smith")`.

### DIFF
--- a/pydough/metadata/properties/masked_table_column_metadata.py
+++ b/pydough/metadata/properties/masked_table_column_metadata.py
@@ -1,0 +1,194 @@
+"""
+Definition of the class for PyDough metadata for properties that access a
+column of a table from a relational system that has been masked by a certain
+protocol, but has an un-masking protocol as well.
+"""
+
+__all__ = ["MaskedTableColumnMetadata"]
+
+
+from pydough.metadata.collections import CollectionMetadata
+from pydough.metadata.errors import (
+    NoExtraKeys,
+    PyDoughMetadataException,
+    extract_array,
+    extract_bool,
+    extract_object,
+    extract_string,
+)
+from pydough.types import PyDoughType, parse_type_from_string
+from pydough.types.errors import PyDoughTypeException
+
+from .table_column_metadata import TableColumnMetadata
+
+
+class MaskedTableColumnMetadata(TableColumnMetadata):
+    """
+    Concrete metadata implementation for a PyDough property representing a
+    column of data from a relational table that has been masked by a certain
+    protocol, but has an un-masking protocol as well.
+    """
+
+    # Set of names of fields that can be included in the JSON object
+    # describing a table column property.
+    allowed_fields: set[str] = TableColumnMetadata.allowed_fields | {
+        "protect protocol",
+        "unprotect protocol",
+        "server masked",
+    }
+
+    def __init__(
+        self,
+        name: str,
+        collection: CollectionMetadata,
+        data_type: PyDoughType,
+        column_name: str,
+        unprotect_protocol: str,
+        protect_protocol: str,
+        server_masked: bool,
+        sample_values: list | None = None,
+        description: str | None = None,
+        synonyms: list[str] | None = None,
+        extra_semantic_info: dict | None = None,
+    ):
+        super().__init__(
+            name,
+            collection,
+            data_type,
+            column_name,
+            sample_values,
+            description,
+            synonyms,
+            extra_semantic_info,
+        )
+        self._unprotect_protocol: str = unprotect_protocol
+        self._protect_protocol: str = protect_protocol
+        self._server_masked: bool = server_masked
+
+    @property
+    def unprotect_protocol(self) -> str:
+        """
+        Returns the format string used to un-mask the data in this
+        table column property. The format string should create SQL text such
+        that when `self.unprotect_protocol.format(value)` is called, where
+        `value` is the SQL text string corresponding to a value that is masked,
+        it returns the SQL text string that will un-mask the value.
+        """
+        return self._unprotect_protocol
+
+    @property
+    def protect_protocol(self) -> str:
+        """
+        Returns the format string used to mask the data in this table column
+        property. The format string should create SQL text such that when
+        `self.protect_protocol.format(value)` is called, where `value` is the
+        SQL text string corresponding to a value that is to be masked, it
+        returns the SQL text string that will mask the value in a way that can
+        be reversed by `self.unprotect_protocol`.
+        """
+        return self._protect_protocol
+
+    @property
+    def server_masked(self) -> bool:
+        """
+        Returns whether the data in this table column property is masked in
+        a manner that corresponds to a server that can be sent predicate queries
+        in order to infer smarter predicates that do not require un-masking the
+        data.
+        """
+        return self._server_masked
+
+    @property
+    def column_name(self) -> str:
+        return self._column_name
+
+    @staticmethod
+    def create_error_name(name: str, collection_error_name: str):
+        return f"masked table column property {name!r} of {collection_error_name}"
+
+    @property
+    def components(self) -> list:
+        comp: list = super().components
+        comp.append(self.column_name)
+        comp.append(self.unprotect_protocol)
+        comp.append(self.protect_protocol)
+        comp.append(self.server_masked)
+        return comp
+
+    @staticmethod
+    def parse_from_json(
+        collection: CollectionMetadata, property_name: str, property_json: dict
+    ) -> None:
+        """
+        Procedure dispatched from PropertyMetadata.parse_from_json to handle
+        the parsing for table column properties.
+
+        Args:
+            `collection`: the metadata for the PyDough collection that the
+            property would be inserted into.
+            `property_name`: the name of the property that would be inserted.
+            `property_json`: the JSON object that would be parsed to create
+            the new table column property.
+
+        Raises:
+            `PyDoughMetadataException`: if the JSON for the property is
+            malformed.
+        """
+        error_name: str = MaskedTableColumnMetadata.create_error_name(
+            property_name, collection.error_name
+        )
+        # Extract the `data_type` and `column_name` fields from the JSON object
+        type_string: str = extract_string(property_json, "data type", error_name)
+        try:
+            data_type: PyDoughType = parse_type_from_string(type_string)
+        except PyDoughTypeException as e:
+            raise PyDoughMetadataException(*e.args)
+        column_name: str = extract_string(property_json, "column name", error_name)
+
+        # Extract the `unprotect protocol`, `protect protocol`, and
+        # `server masked` fields from the JSON object.
+        unprotect_protocol: str = extract_string(
+            property_json, "unprotect protocol", error_name
+        )
+        protect_protocol: str = extract_string(
+            property_json, "protect protocol", error_name
+        )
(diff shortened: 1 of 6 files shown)


documentation/metadata.md
```diff
@@ -145,6 +146,32 @@ Example of the structure of the metadata for a table column property:
 }
 ```
 
+<!-- TOC --><a name="property-type-masked-table-column"></a>
+### Property Type: Masked Table Column
+
+A property with this type is the same as a regular table column, except the data in the underlying table has been masked by some sort of encryption protocol. The metadata includes information required to encrypt values in SQL with the same masking protocol, or how to generate SQL to unmask data that has been masked. Properties of this type have a type string of "masked table column" and have the following additional key-value pairs in their metadata JSON object, in addition to all of the properties from [table column](#property-type-table-column):
```
hadia206, line 152:
```suggestion
A property with this type is the same as a regular table column, except the data in the underlying table has been masked using an encryption protocol. The metadata includes the information required to either encrypt values in SQL using the same masking protocol, or unmask previously masked data in SQL queries. 

Properties of this type use the type string "masked table column" and include all the properties from [table column](#property-type-table-column), plus the following additional key-value pairs in their metadata JSON object:
```

documentation/metadata.md
```diff
@@ -145,6 +146,32 @@ Example of the structure of the metadata for a table column property:
...
+
+- `unprotect protocol` (required): a Python format string indicating the SQL text that is used to unmask the data after reading it from the underlying table. The format string should expect a single value to get injected (e.g. `"SUBSTRING({0}, -1) || SUBSTRING({0}, 1, LENGTH({0}) - 1)".format("c_name")` can be used to create the SQL text `SUBSTRING(c_name, -1) || SUBSTRING(c_name, 1, LENGTH(c_name) - 1)`).
```
hadia206, line 154:
```suggestion
- `unprotect protocol` (required): a Python format string representing the SQL text that is used to unmask the data after reading it from the underlying table. The format string should expect a single placeholder value (e.g. `"SUBSTRING({0}, -1) || SUBSTRING({0}, 1, LENGTH({0}) - 1)".format("c_name")` will generate the SQL text `SUBSTRING(c_name, -1) || SUBSTRING(c_name, 1, LENGTH(c_name) - 1)`).
```

documentation/metadata.md
```diff
@@ -145,6 +146,32 @@ Example of the structure of the metadata for a table column property:
...
+- `protect protocol` (required): a Python format string, in the same format as `unprotect protocol`, indicating the scheme used to mask the data in the first place. This can be used to create values that are encrypted in the same scheme as the data, thus allowing comparisons between masked data.
```
hadia206, line 155:
```suggestion
- `protect protocol` (required): a Python format string, in the same format as `unprotect protocol`, used to describe how the data was originally masked. This can be used to generate masked values consistent with the encryption scheme, allowing operations such as comparisons between masked data.
```

documentation/metadata.md
```diff
@@ -145,6 +146,32 @@ Example of the structure of the metadata for a table column property:
...
+- `server masked` (optional): a boolean indicating whether the column was encrypted in such at a server, which has been attached to PyDough, can be queried in order to rewrite predicates & expressions involving the column to avoid unmasking the data.
```
hadia206, line 156:
```suggestion
- `server masked` (optional): a boolean flag indicating whether the column was masked on a server that is attached to PyDough. If `true`, PyDough can use it to optimize queries by rewriting predicates and expressions to avoid unmasking the data.
```

pydough/metadata/properties/masked_table_column_metadata.py
```diff
@@ -0,0 +1,189 @@
...
+        """
+        Returns whether the data in this table column property is masked in
+        a manner that corresponds to a server that can be sent predicate queries
+        in order to infer smarter predicates that do not require un-masking the
+        data.
+        """
+        return self._server_masked
+
+    @staticmethod
+    def create_error_name(name: str, collection_error_name: str):
```
hadia206, line 102: return type hint `-> str`

pydough/metadata/properties/masked_table_column_metadata.py
```diff
@@ -0,0 +1,189 @@
...
+        return f"masked table column property {name!r} of {collection_error_name}"
+
+    @property
+    def components(self) -> list:
```
hadia206, line 106: nit: `list[Any]`

pydough/metadata/properties/masked_table_column_metadata.py
```diff
@@ -0,0 +1,189 @@
+"""
+Definition of the class for PyDough metadata for properties that access a
+column of a table from a relational system that has been masked by a certain
+protocol, but has an un-masking protocol as well.
```
hadia206, line 4:
```suggestion
protocol, and includes an associated unmasking protocol to reverse the operation.
```

pydough/metadata/properties/masked_table_column_metadata.py
```diff
@@ -0,0 +1,189 @@
...
+from pydough.types import PyDoughType, parse_type_from_string
+from pydough.types.errors import PyDoughTypeException
+
+from .table_column_metadata import TableColumnMetadata
+
+
+class MaskedTableColumnMetadata(TableColumnMetadata):
+    """
+    Concrete metadata implementation for a PyDough property representing a
+    column of data from a relational table that has been masked by a certain
```
hadia206, line 28:
```suggestion
    column of data in a relational table that where the data is masked by a specific
```

pydough/metadata/properties/masked_table_column_metadata.py
```diff
@@ -0,0 +1,189 @@
...
+    protocol, but has an un-masking protocol as well.
```
hadia206, line 29:
```suggestion
    protocol, and can be unmasked using a defined unmasking protocol.
```

pydough/metadata/properties/masked_table_column_metadata.py
```diff
@@ -0,0 +1,189 @@
...
+        self._server_masked: bool = server_masked
+
+    @property
+    def unprotect_protocol(self) -> str:
+        """
+        Returns the format string used to un-mask the data in this
+        table column property. The format string should create SQL text such
+        that when `self.unprotect_protocol.format(value)` is called, where
+        `value` is the SQL text string corresponding to a value that is masked,
+        it returns the SQL text string that will un-mask the value.
```
hadia206, line 75:
```suggestion
        Returns the format string used to un-mask the data in this column.
        Call `self.unprotect_protocol.format(value)` to generate SQL for unmasking, where `value` is the SQL text string corresponding to a value that is masked.
```

pydough/metadata/properties/masked_table_column_metadata.py
```diff
@@ -0,0 +1,189 @@
...
+
+    @property
+    def protect_protocol(self) -> str:
+        """
+        Returns the format string used to mask the data in this table column
+        property. The format string should create SQL text such that when
+        `self.protect_protocol.format(value)` is called, where `value` is the
+        SQL text string corresponding to a value that is to be masked, it
+        returns the SQL text string that will mask the value in a way that can
+        be reversed by `self.unprotect_protocol`.
```
hadia206, line 87:
```suggestion
        Returns the format string used to mask the data in this column.
        Should be reversible by `self.unprotect_protocol`.
```

pydough/metadata/properties/masked_table_column_metadata.py
```diff
@@ -0,0 +1,189 @@
...
+            column_name,
+            unprotect_protocol,
+            protect_protocol,
+            server_masked,
+            sample_values,
+            description,
+            synonyms,
+            extra_semantic_info,
+        )
+        collection.add_property(property)
```
hadia206, line 189: nit: add `__repr__`

pydough/metadata/properties/masked_table_column_metadata.py
```diff
@@ -0,0 +1,189 @@
...
+        if "sample values" in property_json:
+            sample_values = extract_array(property_json, "sample values", error_name)
+        if "description" in property_json:
+            description = extract_string(property_json, "description", error_name)
+        if "synonyms" in property_json:
+            synonyms = extract_array(property_json, "synonyms", error_name)
+        if "extra semantic info" in property_json:
+            extra_semantic_info = extract_object(
+                property_json, "extra semantic info", error_name
+            )
```
hadia206, line 173: This is duplicate of parent class. Instead of duplicating the code, let's move that to be in helper function (`get_optional_fileds` or something)in the parent class and re-use it here?

hadia206 (approved):
Looks good to me. 
Thanks Kian

pydough/metadata/properties/masked_table_column_metadata.py
knassre-bodo: It already has one waaaay in the furthest ancestor abstract base class `AbstractMetadata`:
- AbstractMetadata (for graphs, collections & properties):
  - PropertyMetadata (for both expressions & relationships of a collection)
    - ScalarAttributeMetadata (for expressions of a collection, e.g. columns of a table)
      - TableColumnMetadata (for expressions that are 1:1 with a column in the SQL table)
        - MaskedtableColumnMetadata (the new one

knassre-bodo:
After some discussions, going to expand on the definition further with a new addition: the metadata has **TWO** data types (one for the unmasked data, one for the masked data)

pydough/metadata/properties/general_join_metadata.py
```diff
@@ -48,9 +46,9 @@ def __init__(
         condition: str,
         self_name: str,
         other_name: str,
-        description: str | None,
-        synonyms: list[str] | None,
-        extra_semantic_info: dict | None,
+        description: str | None = None,
+        synonyms: list[str] | None = None,
+        extra_semantic_info: dict | None = None,
```
knassre-bodo, line 51: Still needs to be here (as an optional argument) to the constructor so it can be used for `build_reverse_relationship`.

PR #405, opened 2025-08-14
Adding additional relational simplification patterns for datetime comparisons

knassre-bodo:
Patterns include:
- Switch comparisons between `QUARTER(x)` and a literal use `MONTH(x)` instead, e.g. `QUARTER(x) == 2` becomes `ISIN(MONTH(x), [4, 5, 6])`
- Having patterns such as `MONTH(x) < 0` return False and `HOUR(x) <= 60` return True
- Simplify `DATETIME(DATETIME(u, v, w), x, y, z)` -> `DATETIME(u, v, w, x, y, z)`
- Simplify a chain of `DATETIME` information on a literal to produce the final literal (e.g. `DATETIME('2024-13-05', '+25 days', 'start of month')` -> `'2024-02-01`)
- Simplify datetime part extraction on literals, e.g. `YEAR('2023-01-01')` ->` 2023`
- Infer that a column `x` is non-null if it is used in a filter / join condition (for an inner/semi join) where the condition would be false if `x` were NULL
- Simplify more functions to a NULL literal if any of their arguments are a NULL literal.

### DIFF
--- a/tests/gen_data/init_defog_mysql.sql
+++ b/tests/gen_data/init_defog_mysql.sql
@@ -412,10 +412,10 @@ INSERT INTO `notifications` (`id`, `user_id`, `message`, `type`, `status`, `crea
  (10, 8, 'Playtime! New games and toys have arrived', 'promotion', 'archived', CAST('2023-06-01 18:00:00' AS DATETIME), 'email', NULL, 'https://kidzplayhouse.com/new-arrivals'),
  (11, 9, 'Here''s $10 to start your glow up!', 'promotion', 'unread', CAST('2023-06-01 10:15:00' AS DATETIME), 'email', NULL, 'https://beautytrending.com/new-customer'),
  (12, 10, 'Your order #ord_mjs337 is being processed', 'transaction', 'read', CAST('2023-06-04 19:31:30' AS DATETIME), 'web_app', 'web_d8180kaf', 'https://gamerush.co/orders/32e2b29c'),
- (13, 1, 'New promotion: Get 10% off your next order!', 'promotion', 'unread', NOW() - INTERVAL 7 DAY, 'email', NULL, 'https://techmart.com/promo/TECH10'),
- (14, 1, 'Your order #456def has been delivered', 'transaction', 'unread', NOW() - INTERVAL 14 DAY, 'mobile_app', 'mobile_8fh2k1', 'app://orders/456def'),
- (15, 2, 'Reminder: Your FitLife membership expires in 7 days', 'general', 'unread', NOW() - INTERVAL 21 DAY, 'email', NULL, 'https://fitlifegear.com/renew'),
- (16, 2, 'Weekend Flash Sale: 25% off all activewear!', 'promotion', 'unread', NOW() - INTERVAL 7 DAY + INTERVAL 2 DAY, 'mobile_app', 'mobile_yjp08q', 'app://shop/activewear');
+ (13, 1, 'New promotion: Get 10% off your next order!', 'promotion', 'unread', CURRENT_DATE - INTERVAL ((DAYOFWEEK(CURRENT_DATE) - 2) % 7) DAY - INTERVAL 7 DAY, 'email', NULL, 'https://techmart.com/promo/TECH10'),
+ (14, 1, 'Your order #456def has been delivered', 'transaction', 'unread', CURRENT_DATE - INTERVAL ((DAYOFWEEK(CURRENT_DATE) - 2) % 7) DAY - INTERVAL 14 DAY, 'mobile_app', 'mobile_8fh2k1', 'app://orders/456def'),
+ (15, 2, 'Reminder: Your FitLife membership expires in 7 days', 'general', 'unread', CURRENT_DATE - INTERVAL ((DAYOFWEEK(CURRENT_DATE) - 2) % 7) DAY - INTERVAL 21 DAY, 'email', NULL, 'https://fitlifegear.com/renew'),
+ (16, 2, 'Weekend Flash Sale: 25% off all activewear!', 'promotion', 'unread', CURRENT_DATE - INTERVAL ((DAYOFWEEK(CURRENT_DATE) - 2) % 7) DAY - INTERVAL 7 DAY + INTERVAL 2 DAY, 'mobile_app', 'mobile_yjp08q', 'app://shop/activewear');
 
 INSERT INTO `user_sessions` (`user_id`, `session_start_ts`, `session_end_ts`, `device_type`, `device_id`) VALUES
  (1, CAST('2023-06-01 09:45:22' AS DATETIME), CAST('2023-06-01 10:20:35' AS DATETIME), 'mobile_app', 'mobile_8fh2k1'),


tests/test_pipeline_defog_custom.py
```diff
@@ -1879,6 +1880,196 @@ def get_day_of_week(
...
+        pytest.param(
+            PyDoughPandasTest(
+                "result = ("
+                " transactions"
+                " .WHERE(YEAR(date_time) == 2023)"
+                " .WHERE((RANKING(by=date_time.ASC()) == 1) | (RANKING(by=date_time.DESC()) == 1))"
+                " .CALCULATE("
+                " date_time,"
+                " s00 = DATETIME(DATETIME(date_time, 'start of week'), '-8 weeks'),"  # -> DATETIME(date_time, 'start of week', '-8 weeks')
+                " s01 = QUARTER(date_time) == 0,"  # KEEP_IF(False, PRESENT(date_time))
```
knassre-bodo, line 1892: None of these tests, in the final SQL, contain `KEEP_IF(x, PRESENT(date_time))` because we can infer that `PRESENT(date_time)` is always True if it occurs after the filter `YEAR(date_time) == 2023)`

tests/test_sql_refsols/tpch_q10_sqlite.sql
```diff
@@ -8,21 +8,8 @@ WITH _s3 AS (
...
-      WHEN CAST(STRFTIME('%m', orders.o_orderdate) AS INTEGER) <= 9
-      AND CAST(STRFTIME('%m', orders.o_orderdate) AS INTEGER) >= 7
-      THEN 3
-      WHEN CAST(STRFTIME('%m', orders.o_orderdate) AS INTEGER) <= 12
-      AND CAST(STRFTIME('%m', orders.o_orderdate) AS INTEGER) >= 10
-      THEN 4
-    END = 4
-    AND CAST(STRFTIME('%Y', orders.o_orderdate) AS INTEGER) = 1993
+    CAST(STRFTIME('%Y', orders.o_orderdate) AS INTEGER) = 1993
+    AND CAST(STRFTIME('%m', orders.o_orderdate) AS INTEGER) IN (10, 11, 12)
```
knassre-bodo, line 12: Compressed `(YEAR(x) == 2023) & (QUARTER(x) == 4)` into (YEAR(x) == 2023) & (ISIN(MONTH(x), [10, 11, 12]))`

tests/test_sql_refsols/technograph_year_cumulative_incident_rate_overall_sqlite.sql
```diff
@@ -36,15 +36,15 @@ WITH _s2 AS (
 SELECT
   year AS yr,
   ROUND(
-    CAST(SUM(COALESCE(sum_n_rows, 0)) OVER (ORDER BY year ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) AS REAL) / SUM(COALESCE(sum_expr_3, 0)) OVER (ORDER BY year ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW),
+    CAST(SUM(COALESCE(sum_n_rows, 0)) OVER (ORDER BY year ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW) AS REAL) / SUM(sum_expr_3) OVER (ORDER BY year ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW),
```
knassre-bodo, line 39: Inferred that `sum_expr_3` is already non-null from the filter on line 62

tests/test_sql_refsols/smoke_b_sqlite.sql
```diff
@@ -59,7 +59,7 @@ SELECT
   DATE('2025-01-01 12:35:13', 'start of day') AS g,
   CONCAT_WS(
     ';',
-    CAST(STRFTIME('%H', DATETIME('2025-01-01 12:35:13')) AS INTEGER),
+    12,
```
knassre-bodo, line 62: Simplified directly to a literal

tests/test_sql_refsols/defog_ewallet_adv11_sqlite.sql
```diff
@@ -18,9 +18,9 @@ WITH _s1 AS (
 )
 SELECT
   users.uid,
-  COALESCE(_s1.sum_duration, 0) AS total_duration
+  _s1.sum_duration AS total_duration
```
knassre-bodo, line 21: We know that `_s1.sum_duration` is non-null because of the following (expand the code above to confirm):
- The `SUM` call is guaranteed to be non-null if its inputs are non-null (so long as there is a group by)
- The very large input expression to the `SUM` is guaranteed to be non-null if `session_end_ts` and `session_start_ts` are both non-null
- We know that `session_end_ts` and `session_start_ts` are both non-null because of the filter on line 15

tests/test_plan_refsols/technograph_year_cumulative_incident_rate_overall.txt
```diff
@@ -1,4 +1,4 @@
...
+ROOT(columns=[('yr', year), ('cum_ir', ROUND(RELSUM(args=[DEFAULT_TO(sum_n_rows, 0:numeric)], partition=[], order=[(year):asc_last], cumulative=True) / RELSUM(args=[sum_expr_3], partition=[], order=[(year):asc_last], cumulative=True), 2:numeric)), ('pct_bought_change', ROUND(100.0:numeric * sum_expr_3 - PREV(args=[sum_expr_3], partition=[], order=[(year):asc_last]) / PREV(args=[sum_expr_3], partition=[], order=[(year):asc_last]), 2:numeric)), ('pct_incident_change', ROUND(100.0:numeric * DEFAULT_TO(sum_n_rows, 0:numeric) - PREV(args=[DEFAULT_TO(sum_n_rows, 0:numeric)], partition=[], order=[(year):asc_last]) / PREV(args=[DEFAULT_TO(sum_n_rows, 0:numeric)], partition=[], order=[(year):asc_last]), 2:numeric)), ('bought', sum_expr_3), ('incidents', DEFAULT_TO(sum_n_rows, 0:numeric))], orderings=[(year):asc_first])
```
knassre-bodo, line 1: All of these `DEFAULT_TO(sum_expr_3, 0)` calls can be reaplced with `sum_expr_3` because, in the line below, we know that if `sum_expr_3` is NULL then `DEFAULT_TO(sum_expr_3, 0:numeric) > 0:numeric` will be False, ergo all the remaining rows must have a non-null value of `sum_expr_3`

pydough/conversion/relational_simplification.py
```diff
@@ -279,6 +295,362 @@ def visit_window_expression(
...
+                            self.quarter_month_array(lit_expr.value),
+                        ],
+                    )
+                # QUARTER(x) == 3 <=> KEEP_IF(False, PRESENT(x))
+                else:
+                    result = conditional_false
+            case (pydop.NEQ, pydop.QUARTER, NumericType()) if isinstance(
+                lit_expr.value, int
+            ):
+                # QUARTER(x) == 4 <=> NOT(ISIN(MONTH(x), [10, 11, 12]))
```
john-sanchez31, line 399:
```suggestion
# QUARTER(x) != 4 <=> NOT(ISIN(MONTH(x), [10, 11, 12]))
```

pydough/conversion/relational_simplification.py
```diff
@@ -279,6 +295,362 @@ def visit_window_expression(
...
+                NumericType(),
+            ) if isinstance(lit_expr.value, int) and lit_expr.value < 1:
+                result = conditional_false
+            # HOUR(x) <= -1 <=> KEEP_IF(False, PRESENT(x)) (same for minute/second)
+            # HOUR(x) == -1 <=> KEEP_IF(False, PRESENT(x)) (same for minute/second)
+            case (
+                pydop.LEQ | pydop.EQU,
+                pydop.HOUR | pydop.MINUTE | pydop.SECOND,
+                NumericType(),
+            ) if isinstance(lit_expr.value, int) and lit_expr.value < 0:
```
john-sanchez31, line 530: Can `HOUR(x)` return negative numbers? I don't really get this case

pydough/conversion/relational_simplification.py
```diff
@@ -279,6 +295,362 @@ def visit_window_expression(
...
+
+        Returns:
+            The simplified expression if possible, otherwise the original
+            expression.
+        """
+        # Extract a pandas Timestamp from the literal if possible. Allows cases
+        # where the literal is a native Python datetime/date, a pandas
+        # Timestamp, or a string without any alphabetic characters (to avoid
+        # parsing things like 'now' that depend on the current date).
+        ts: pd.Timestamp | None = None
```
john-sanchez31, line 616: Can we change its name for something more meaningful e.g `timestp` or `timestamp` if not reserved

pydough/conversion/relational_simplification.py
knassre-bodo: The whole point is that it can't return negatives, therefore `HOUR(x) <= -5` will always return False

juankx-bodo:
Remember to fix typo in PR title before merge

pydough/conversion/relational_simplification.py
```diff
@@ -279,6 +317,511 @@ def visit_window_expression(
...
+                match unit:
+                    case DateTimeUnit.YEAR:
+                        timestamp_value = timestamp_value + pd.DateOffset(years=amt)
+                    case DateTimeUnit.QUARTER:
+                        timestamp_value = timestamp_value + pd.DateOffset(
+                            months=amt * 3
+                        )
+                    case DateTimeUnit.MONTH:
+                        timestamp_value = timestamp_value + pd.DateOffset(months=amt)
+                    case DateTimeUnit.WEEK:
```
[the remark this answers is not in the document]
knassre-bodo: The issue wasn't that `+1 month` becomes `+30 days`, moreso that `+1 month` when the current date is the last day of the month is dicey. This is a more general fix which should be considered correct.

tests/test_pipeline_defog_custom.py
```diff
@@ -2011,6 +2012,196 @@ def get_day_of_week(
...
+                        "s81": [1, 1],
+                        "s82": [0, 0],
+                        "s83": [1, 1],
+                        "s84": [0, 0],
+                    }
+                ),
+                "simplification_4",
+                kwargs={"pd": pd, "datetime": datetime},
+            ),
+            id="simplification_4",
```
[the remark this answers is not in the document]
knassre-bodo: > DATETIME with Alphabetic strings (like 'now', 'today')

We have a bunch of those already in existing tests that don't get simplified

> DATETIME with Malformed dates ('2024-13-99')

That is undefined behavior in PyDough. For now the simplifier just ignores those cases, but there is no good/bad test for that situation since there isn't a good way to test for it (e.g. in sqlite, it just returns null if you do anything weird)

pydough/conversion/relational_simplification.py
```diff
@@ -279,6 +317,511 @@ def visit_window_expression(
...
+                    result = CallExpression(
+                        pydop.LET,
+                        expr.data_type,
+                        [
+                            self.switch_operator(func_expr, pydop.MONTH),
+                            LiteralExpression((lit_expr.value * 3) - 2, NumericType()),
+                        ],
+                    )
+                # QUARTER(x) < 1 <=> KEEP_IF(False, PRESENT(x))
+                elif lit_expr.value < 2:
```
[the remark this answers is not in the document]
knassre-bodo: I'm moving this into your boundary dictionary idea, for greater simplification.

tests/test_pipeline_defog_custom.py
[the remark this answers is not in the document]
knassre-bodo: > Test for adding/subtracting months that cross February in a leap year (high priority based on what I have seen with dialects)

Sure but there won't be a dialect difference in this case because, once simplified, there is no DATEADD call that would be different dialect-to-dialect

pydough/conversion/relational_simplification.py
```diff
@@ -279,6 +317,511 @@ def visit_window_expression(
...
+        # as a literal.
+        match op:
+            case pydop.YEAR:
+                return LiteralExpression(timestamp_value.year, NumericType())
+            case pydop.QUARTER:
+                quarter: int = ((timestamp_value.month - 1) // 3) + 1
+                return LiteralExpression(quarter, NumericType())
+            case pydop.MONTH:
+                return LiteralExpression(timestamp_value.month, NumericType())
+            case pydop.DAY:
```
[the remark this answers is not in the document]
knassre-bodo: You know what... I'll add it. Why not :)

hadia206 (approved):
Looks good to me.
Great Job Kian!

PR #406, opened 2025-08-15
Fixing bug with complex combinations of PARTITION and CROSS

knassre-bodo:
The bug originated in a pattern such as this: `A.PARTITION(...).CROSS(B.PARTITION(...)).PARTITION(...)`, where the final `PARTITION` did not work due to this error message: `PyDoughUnqualifiedException: Unsupported collection node for `split_partition_ancestry`: UnqualifiedCross`. This task fixes that bug so the afformentioned case works, and adds test for it and several other uses of `x.CROSS(y)` where `x` and `y` contain `PARTITION`

### DIFF
--- a/tests/test_sql_refsols/simplification_3_sqlite.sql
+++ b/tests/test_sql_refsols/simplification_3_sqlite.sql
@@ -93,26 +93,3 @@ SELECT
   MAX(rsiz1) AS s35,
   ROUND(SUM(rsiz2), 2) AS s36
 FROM _t1
-
-
-SELECT
-  sbcustname,
-   COALESCE(
-      SUM(ABS(COALESCE(CAST(sbcustpostalcode AS INTEGER), 0))) OVER (ORDER BY sbcustname ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW),
-      0.1
-    ) AS rsum2,
-    ROW_NUMBER() OVER (ORDER BY sbcustname) AS rnk
-FROM main.sbcustomer
-ORDER BY sbcustname
-;
-
-SELECT
-  sbCustname,
-   COALESCE(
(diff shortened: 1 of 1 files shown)


pydough/conversion/filter_pushdown.py
```diff
@@ -187,6 +187,15 @@ def visit_join(self, join: Join) -> RelationalNode:
...
 
+        # If the join type is LEFT or SEMI but the condition is TRUE, convert it
+        # to an INNER join.
+        if (
+            isinstance(join.condition, LiteralExpression)
+            and bool(join.condition.value)
+            and join.join_type in (JoinType.LEFT, JoinType.SEMI)
+        ):
+            join_type = JoinType.INNER
+
```
knassre-bodo, line 198: This was a nice-to-have since I noticed we had a few `LEFT JOIN ON TRUE` in the generated plans for the `part_cross_part_xxx` tests.

pydough/conversion/hybrid_decorrelater.py
```diff
@@ -43,7 +43,7 @@ def __init__(self) -> None:
         self.children_indices: list[int] = []
 
     def make_decorrelate_parent(
-        self, hybrid: HybridTree, child_idx: int, min_steps: int, max_steps: int
+        self, hybrid: HybridTree, child_idx: int, max_steps: int
```
knassre-bodo, line 46: Turns out, the `min_steps` parameter here is unneeded, and was causing bugs in situations like `part_cross_part_b`.

pydough/unqualified/qualification.py
```diff
@@ -905,6 +905,7 @@ def split_partition_ancestry(
                 | UnqualifiedSingular()
                 | UnqualifiedPartition()
                 | UnqualifiedBest()
+                | UnqualifiedCross()
```
knassre-bodo, line 908: This was the bugfix for `part_cross_part_c` (has more to do with PARTITION than with CROSS).

PR #407, opened 2025-08-19
Adding relational optimization for Join-Aggregate transpose

knassre-bodo:
Adding an optimization pass which identifies cases where a join occurs after an aggregation, but the join is on the aggregation keys and placing the join before the aggregation would reduce the number of rows inputted to the aggregation, so it switched the two:
- Only allowed on a semi join (if the aggregation is on the left) or an inner join (if either side is the aggregation)
  - Also allowed for left joins (if the aggregation is on the right) with special circumstances (referred to as the left join case)
- Only allowed if the cardinality from the aggregate to the other side is singular & filtering
- All columns from the non-aggregation side are passed through via `ANY_VALUE` (except equijoin keys, which are just replaced with the equivalent join key from the aggregation side, which should be one of the aggregation key).
- If the left join case occurs, then `COUNT(*)` is replaced with `COUNT(x)` (where `x` is one of the equi-join keys from the right hand side, aborting the attempt if there is no such key).
 - Similarly, `COUNT(x)` calls have an additional `KEEP_IF(..., y != 0)` post-processing step, where `y` is a sentinel column to determine whether there was a match or not from the left join. If there was a `COUNT(*)` rewritten as `COUNT(x)`, then that column is used as `y`. Otherwise, a new `COUNT(*)` is inserted in the aggregation and used as `y`.

Also added some new simplification rules to the SQLGlot simplifier override file to account for newly generated patterns:
- `CASE WHEN x != y THEN x ELSE NULL END` -> `NULLIF(x, y)`
- `COALESCE(NULLIF(x, y), y)` -> `COALESCE(x, y)`
- `COALESCE(NULLIF(x, y), z)` -> `CASE WHEN x = y THEN z ELSE x END`
- `SUM(NULLIF(x, 0))` -> `SUM(x)`
- `COALESCE(COUNT(x)) -> `COUNT(x)`
- `COALESCE(COUNT_IF(x))` -> `COUNT_IF(x)`
- `NULLIF(COALESCE(x, y), y)` -> `NULLIF(x, y)`

The net result of all of these changes is the following:
- Plans will tend to be slightly more compact, and closer to how a human would write them
- Performance of queries will sometimes change, most often with a slight performance gain (though occasionally with a slight performance regression)

A side effect of these changes is the sometimes-extraneous and unnecessary inclusion of the extra `ANY_VALUE` calls, when those columns could instead be switched out with the grouping keys. A followup may potentially be warranted to change how uniqueness information is propagated so that column pruning can make this switch when the grouping key is unused, but one of these `MIN` / `MAX` / `ANY_VALUE` columns is also unique.


tests/test_pydough_functions/common_prefix_pydough_functions.py
```diff
@@ -698,7 +698,26 @@ def common_prefix_al():
...
+                67393,
+                87784,
+                148252,
+                176947,
+                196620,
+                103099,
+                169275,
+            ),
+        )
+    )
```
knassre-bodo: This test was being a pill and randomly hanging forever after the optimization, even though the new query was being written better (likely because of some sqlite nonsense with how joins are/aren't optimized) so I made a minor tweak to the query so it has the same answer, but the intermediate tables are much smaller.

tests/test_pipeline_common_prefix.py
```diff
@@ -623,12 +623,13 @@
...
                             f"Customer#{i:09}"
-                            for i in (138841, 36091, 54952, 103768, 46081)
+                            for i in (6434, 45280, 60493, 87616, 132775)
                         ],
-                        "n_orders": [21, 20, 19, 19, 17],
+                        "n_orders": [2, 2, 2, 2, 2],
                     }
                 ),
                 "common_prefix_y",
+                order_sensitive=True,
```
knassre-bodo: Changed the test slightly so it isn't as time consuming to compute

pydough/conversion/join_aggregate_transpose.py
```diff
@@ -0,0 +1,409 @@
...
+
+    def generate_name(self, base: str, used_names: Iterable[str]) -> str:
+        """
+        Generates a new name for a column based on the base name and the existing
+        columns in the join. This is used to ensure that the new column names are
+        unique and do not conflict with existing names.
+        """
+        if base not in used_names:
+            return base
+        i = 0
```
john-sanchez31: Check type hints on `i` and `name`

pydough/conversion/join_aggregate_transpose.py
```diff
@@ -0,0 +1,409 @@
...
+
+        # The cardinality with regards to the input being considered must be
+        # singular (unless the aggregations allow plural), and must be
+        # filtering (since the point of joining before aggregation is to reduce
+        # the number of rows to aggregate).
+        cardinality: JoinCardinality = (
+            join.cardinality if is_left else join.reverse_cardinality
+        )
+
+        left_join_case = (
```
john-sanchez31: Type hint missing

pydough/conversion/join_aggregate_transpose.py
```diff
@@ -0,0 +1,409 @@
...
+        non_agg_alias: str | None = (
+            join.default_input_aliases[1] if is_left else join.default_input_aliases[0]
+        )
+
+        # Now that the transpose is deemed possible, if in the left join
+        # scenario, transform any `COUNT(*)` calls into `COUNT(col)`, where
+        # `col` is one of the aggregation keys. If this is not possible, then
+        # abort. Also abort if any of the aggregation keys are not used as
+        # equi-join keys.
+        sentinel_column: RelationalExpression | None = None
```
john-sanchez31: What is a sentinel column?

pydough/conversion/join_aggregate_transpose.py
```diff
@@ -0,0 +1,409 @@
...
+                for col_name, col_expr in join.columns.items():
+                    if (
+                        isinstance(col_expr, ColumnReference)
+                        and col_expr.name == existing_sentinel
+                    ):
+                        sentinel_join_name = col_name
+                        break
+            else:
+                # Otherwise, create a new COUNT(*) column for that purpose.
+                agg_name = self.generate_name("n_rows", aggregate.columns)
```
john-sanchez31: Missing type hint

pydough/sqlglot/override_simplify.py
```diff
@@ -124,6 +125,11 @@ def _simplify(expression, root=True):
         node = simplify_concat(node)
         node = simplify_conditionals(node)
 
+        # PyDough Change: new pre-order transformations
```
john-sanchez31: Are these functions overridden from sqlglot or PyDough specific functions?

tests/test_metadata/defog_graphs.json
```diff
@@ -7,7 +7,7 @@
         "name": "customers",
         "type": "simple table",
         "table path": "main.sbCustomer",
-        "unique properties": ["_id"],
+        "unique properties": ["_id", "name", "email", "address1"],
```
john-sanchez31: Why did this chage?

tests/test_pydough_functions/defog_test_functions.py
```diff
@@ -287,7 +287,7 @@ def impl_defog_broker_adv8():
         & (date_time >= DATETIME("now", "start of week", "-1 week"))
     )
     return Broker.CALCULATE(
-        n_transactions=KEEP_IF(COUNT(selected_txns), COUNT(selected_txns) > 0),
+        n_transactions=KEEP_IF(COUNT(selected_txns), COUNT(selected_txns) != 0),
```
john-sanchez31: Why do this queries change?

tests/test_metadata/defog_graphs.json
knassre-bodo: These were columns that are actually unique, but previously weren't marked as such in the defog metadata. I happened to correct these while making other changes, since I was looking through these.

juankx-bodo: Not sure about this change. Is this required for a specific test case?
While this set of fields may be unique in the actual data, we generally can't assume the same for 'real' databases where data is changing. I think simulating live database behavior with the defog datasets would be better for testing purposes, as this reflects how they would be used in the real world.

pydough/sqlglot/override_simplify.py
```diff
@@ -225,3 +231,111 @@ def simplify_datetrunc(expression: exp.Expression, dialect: Dialect) -> exp.Expr
...
+        return expr
+
+    first = expr.this
+    second = expr.expressions[0]
+
+    if not isinstance(first, exp.Nullif):
+        return expr
+
+    lhs = first.args.get("this")
+    rhs = first.args.get("expression")
```
juankx-bodo: Type hints for `first`, `second`, `lhs` & `rhs`

PR #413, opened 2025-08-25
Keep track of reverse-cardinality in Joins for optimizations

knassre-bodo:
Modifying the cardinality set up for relational join nodes to have a notion of "reverse cardinality", e.g. what is the cardinality from the perspective of the RHS input with regards to the LHS input. For example, when the left hand side is the join is the tpch CUSTOMER table and the right hand side is the tpch ORDERS table, the cardinality is `PLURAL_FILTER` (since each customer can have 0, 1, or multiple matching orders) but the reverse cardinality is `SINGULAR_ACCESS` (since each order has exactly 1 matching customer). This reverse cardinality can be used for two things right away:
- Adjusting the partial aggregation splitting protocol to infer when to push / not push an aggregate into the _right_ hand side based on the reverse cardinality (e.g. if the reverse cardinality is filtering, don't push because the join will actually reduce the number of rows).
- Modifying the column pruning protocol for joins, which currently removes the RHS for certain kinds of joins if the RHS columns are unused and the cardinality is `SINGULAR_ACCESS`: can now dot he same to prune the LHS entirely if the reverse cardinality is `SINGULAR_ACCESS` (e.g. in the `CUSTOMER` -> `ORDERS` example, if it is an inner join and every column in `CUSTOMER` is unused, we can just prune the `CUSTOMER` side of the join entirely)


pydough/relational/relational_nodes/join.py
```diff
@@ -259,9 +275,16 @@ def to_string(self, compact: bool = False) -> str:
...
             ""
             if self.cardinality == JoinCardinality.UNKNOWN_UNKNOWN
+            or self.join_type in (JoinType.SEMI, JoinType.ANTI)
             else f", cardinality={self.cardinality.name}"
         )
-        return f"JOIN(condition={self.condition.to_string(compact)}, type={self.join_type.name}{cardinality_suffix}, columns={self.make_column_string(self.columns, compact)}{correl_suffix})"
+        reverse_cardinality_suffix: str = (
+            ""
+            if self.reverse_cardinality == JoinCardinality.UNKNOWN_UNKNOWN
+            or self.join_type in (JoinType.SEMI, JoinType.ANTI)
```
knassre-bodo, line 284: This ensures that we don't include the cardinalities in the plan file for SEMI/ANTI joins (since they only make sense for left/inner).

pydough/conversion/hybrid_tree.py
```diff
@@ -571,13 +573,27 @@ def add_child(
...
+            min_steps,
+            max_steps,
+            {},
+            reverse_cardinality,
         )
         self._children.append(connection)
 
+        # Augment the reverse cardinality if the parent does not always exist.
+        if (not reverse_cardinality.filters) and (not self.always_exists()):
+            connection.reverse_cardinality = reverse_cardinality.add_filter()
```
knassre-bodo, line 595: Doing this **after** is importaint because `.always_exists()` doesn't work if `self` contains a partition node unless we have already appended the partitioned data to `self._children`

pydough/conversion/hybrid_tree.py
```diff
@@ -605,6 +623,97 @@ def add_child(
...
+        # Once we find the root, infer the cardinality of the join that would
+        # connect just this node to the parent context. The rest of the nodes in
+        # the tree don't matter since they will not affect how many matches
+        # there are back to the parent context, or whether there is always a
+        # match or not for each record in the current context.
+        match self.pipeline[0]:
+            case HybridRoot():
+                # If the parent of the child is a root, it means a cross join
+                # is occurring, so the cardinality depends on the size of the
+                # parent.
```
hadia206, line 691: not clear to me what `size` here means.
```suggestion
                # is occurring, so the cardinality depends on whether
                # the parent context is singular or plural.
```

hadia206 (approved):
Overall looks good to me.
Good job Kian!

pydough/conversion/relational_converter.py
```diff
@@ -1501,6 +1549,8 @@ def optimize_relational_tree(
     # objects inside the plan.
     pruner: ColumnPruner = ColumnPruner()
     root = pruner.prune_unused_columns(root)
+    # print()
+    # print(root.to_tree_string())
```
knassre-bodo, line 1553:
```suggestion
```

PR #417, opened 2025-08-26
Add masked table column relational unprotect rewrite

knassre-bodo:
Adding crucial rewrite step for masked table columns: During relational conversion, if a column is a masked table column, place a `PROJECT` on top of the `SCAN` node fetching data from the table. This `PROJECT` node will invoke an `UNMASK` operator (containing information from the metadata for the masked table column) which will transform the masked columns into their unmasked forms.

Updates existing tests to account for this transformation, ensuring that the `UNMASK` calls are injected into the underlying SQL (and pulled up as late as possible by projection pullup). All the tests in `test_masked_sqlite.py` will now run with the _correct_ e2e answers, as the decryption of the underlying data is now undone by the generated query.


pydough/pydough_operators/expression_operators/masked_expression_function_operator.py
```diff
@@ -0,0 +1,78 @@
...
+    def format_string(self) -> str:
+        """
+        The format string to use for this operator to either mask or unmask the
+        operand.
+        """
+        return (
+            self.masking_metadata.unprotect_protocol
+            if self.is_unprotect
+            else self.masking_metadata.protect_protocol
+        )
```
knassre-bodo, line 65: This switch logic makes the SQL conversion step seemless

tests/test_sql_refsols/cryptbank_agg_04_sqlite.sql
```diff
@@ -1,15 +1,15 @@
 WITH _s1 AS (
   SELECT
     a_branchkey,
-    SUM(a_balance) AS sum_a_balance
+    SUM(SQRT(a_balance)) AS sum_unmask_a_balance
```
knassre-bodo, line 4: Similarly, `a_balance` was encrypted by squaring it, so it is decrypted by taking the square root.

tests/conftest.py
```diff
@@ -575,6 +575,24 @@ def sqlite_technograph_connection() -> DatabaseContext:
...
+def enable_mask_rewrites(request):
+    """
+    Temporarily enable the mask rewrites by setting the environment variable to
+    the specified value, returning the name that should be used to identify
+    the situation ("raw" for disabled, "rewrite" for enabled).
+    """
+    old_value: str = os.environ.get("PYDOUGH_ENABLE_MASK_REWRITES", "0")
+    os.environ["PYDOUGH_ENABLE_MASK_REWRITES"] = request.param[0]
+    yield request.param[1]
+    os.environ["PYDOUGH_ENABLE_MASK_REWRITES"] = old_value
```
knassre-bodo, line 593: **FOR NOW** this environment variable has no effect, but it will in #418. We use it to clearly separate the difference in executing with vs without it, and so we can verify everything still runs correctly in both versions.

This means that, for the time being, the `_raw` vs `_rewrite` files are the same.

juankx-bodo (approved):

pydough/pydough_operators/expression_operators/masked_expression_function_operator.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+
+    def __init__(
+        self,
+        masking_metadata: MaskedTableColumnMetadata,
+        is_unprotect: bool,
+    ):
+        # Create a dummy verifier that requires exactly one argument, since all
+        # masking/unmasking operations are unary.
+        verifier: TypeVerifier = RequireNumArgs(1)
```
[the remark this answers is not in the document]
knassre-bodo: Yes since, from a PyDough perspective, the operator is always invoked the form `UNMASK(arg)` or `MASK(arg)`. Internally, when the macro gets expanded, it may contain multiple arguments, but it is only _parameterized_ on one argument (the thing getting masked/unmasked).

pydough/conversion/relational_converter.py
```diff
@@ -806,7 +822,31 @@ def build_simple_table_scan(
...
+            node.collection.collection.table_path, scan_columns, uniqueness
+        )
+
+        # If any of the columns are masked, insert a projection on top to unmask
+        # them.
+        if any(self.is_masked_column(expr) for expr in node.terms.values()):
+            unmask_columns: dict[str, RelationalExpression] = {}
+            for name, hybrid_expr in node.terms.items():
+                if self.is_masked_column(hybrid_expr):
+                    assert isinstance(hybrid_expr, HybridColumnExpr)
```
juankx-bodo, line 835: Are these assert required? We are here because `is_masked_column(hybrid_expr) == true`. Am I missing something?

PR #427, opened 2025-09-15
Add a `session` keyword argument to `to_sql` and `to_df`

### LINKED ISSUE #421: Allow a session to be passed in as a keyword argument to `to_sql` and `to_df`
Goal: extend the keyword arguments of the `to_sql` and `to_df` arguments to optionally take in a `session` argument, which is a PyDough session object.
- If this argument is omitted, the behavior of `to_sql` and `to_df` is unchanged.
- If this argument is provided, the metadata/configs/database used by `to_sql` and `to_df` will come from the session object instead of `pydough.active_session` or the other keyword arguments
- If this argument is provided **and** any of the other keyword arguments that conflict with it are provided (`metadata`, `configs`, `database`), an exception is raised.

Additionally, to streamline utilization of these parameters throughout the codebase, and trivialize the extension of them, they will now be parceled together within a session object throughout the lifetime of the `to_sql` or `to_df` execution. This means:
- If no keyword argument is used, `pydough.active_session` is communicated throughout the workflow.
- If the `session` keyword argument is used, its value is communicated throughout the workflow.
- If the other keyword arguments are used, a new session object is constructed using their values (and the ones from active session, where omitted) is communicated throughout the workflow.

Examples:
```py
import pydough
from pydough.configs import PyDoughSession
pydough_code = pydough.from_string(...)

# Create metadata objects
meta1 = ...
meta2 = ...

# Create config objects
configs1 = ...
configs2 = ...

# Create database connection objects
db1 = ...
db2 = ...

# Create a custom session object
s1 = PyDoughSession()
s1.metadata = ...
s1.configs = ...
s1.metadata = ...

# Uses pydough.active_session
df1 = to_sql(pydough_code)

# Same as df1
df2 = to_sql(pydough_code, session=pydough.active_session)

# Uses session containing meta1/configs1/db1 (aka uses s1)
df3 = to_sql(pydough_code, session=s1)

# Uses session containing meta2/configs2, and uses the database from `pydough.active_session`
df4 = to_sql(pydough_code, metadata=meta2, configs=configs2)

# Uses session containing meta2/configs2/db2
df5 = to_sql(pydough_code, metadata=meta2, configs=configs2, database=db2)
```

These changes must be thoroughly documented in the usage API documentation, and a small amount of testing is required (both for good and bad cases). Note: the error builder is **not** encapsulated in this manner, since it is only invoked from `pydough.active_session`.

Regarding the encapsulation of sessions, an example would be the arguments to `convert_ast_to_relational`, which currently contain the configs & dialect, but will instead just pass in the session.

knassre-bodo:
Resolves #421.

### DIFF
--- a/documentation/usage.md
+++ b/documentation/usage.md
@@ -788,7 +788,9 @@ The `explain` API is a more generic explanation interface that can be called on 
 - A specific property within a specific collection within a metadata graph object (can be accessed as `graph["collection_name"]["property_name"]`)
 - The PyDough code for a collection that could have `to_sql` or `to_df` called on it.
 
-The `explain` API also has an optional `verbose` argument (default=False) that enables displaying additional information. It also has an optional `session` argument to specify what configs etc. to use when explaining certain terms (if not provided, uses `pydough.active_session`).
+The `explain` API has the following optional arguments:
+* `verbose` (default False): specifies whether to include a more detailed explanation, as opposed to a more compact summary. 
+* `session` (default None): if provided, specifies what configs etc. to use when explaining PyDough code objects (if not provided, uses `pydough.active_session`).
 
 Below are examples of each of these behaviors, using a knowledge graph for the TPCH schema.
 
@@ -1006,8 +1008,9 @@ The `explain` API is limited in that it can only be called on complete PyDough c
 
 To handle cases where you need to learn about a term within a collection, you can use the `explain_term` API. The first argument to `explain_term` is PyDough code for a collection, which can have `explain` called on it, and the second is PyDough code for a term that can be evaluated within the context of that collection (e.g. a scalar term of the collection, or one of its sub-collections).
 
-The `explain_term` API also has a `verbose` keyword argument (default False) to specify whether to include a more detailed explanation, as opposed to a more compact summary. The `explain_term` API also has an optional `verbose` argument (default=False) that enables displaying additional information. It also has an optional `session` argument to specify what configs etc. to use when explaining certain terms (if not provided, uses `pydough.active_session`).
-
+The `explain_term` API has the following optional arguments:
+* `verbose` (default False): specifies whether to include a more detailed explanation, as opposed to a more compact summary. 
+* `session` (default None): if provided, specifies what configs etc. to use when explaining certain terms (if not provided, uses `pydough.active_session`).
 
 Below are examples of using `explain_term`, using a knowledge graph for the TPCH schema. For each of these examples, `european_countries` is the "context" collection, which could have `to_sql` or `to_df` called on it, and `term` is the term being explained with regards to `european_countries`.
 
--- a/pydough/conversion/agg_split.py
+++ b/pydough/conversion/agg_split.py
@@ -477,7 +477,7 @@ def attempt_join_aggregate_transpose(
     # top of the top aggregate.
     if need_projection:
         new_node: RelationalNode = node.copy(
-            inputs=[split_partial_aggregates(input, config) for input in node.inputs]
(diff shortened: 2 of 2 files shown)


pydough/evaluation/evaluate_unqualified.py
```diff
@@ -27,19 +27,47 @@
...
+
+    # If the session is provided, use it directly. Verify it has a metadata
+    # graph attached, and there are no other keyword arguments.
+    if "session" in kwargs:
+        session = kwargs.pop("session")
+        if not isinstance(session, PyDoughSession):
+            raise PyDoughSessionException(
+                f"Expected `session` to be a PyDoughSession, got {session.__class__.__name__}."
+            )
+        if session.metadata is None:
```
juankx-bodo, line 58: We are only checking for `metadata`. Should we check also for `session.config` and `session.database`?
Could we build `new_session` from `session` in `kwargs` if it exists, then if metadata, config or database are provided we overwrite them in `new_session` and finally if any of them is null in `new_session` then we set that value with its value in `active_session`.

tests/test_exploration.py
```diff
@@ -1317,6 +1318,7 @@ def test_unqualified_node_exploration(
     ],
     verbose: bool,
     get_sample_graph: graph_fetcher,
```
juankx-bodo, line 1320: Should we provide the graph as part of the session?

juankx-bodo (approved):

documentation/usage.md
```diff
@@ -716,27 +700,27 @@ The value of `sql` is the following SQL query text as a Python string:
...
   customer.c_name AS name,
   _s1.n_rows AS n_orders,
-  _s1.total
+  _s1.sum_o_totalprice AS total
 FROM main.customer AS customer
 JOIN _s1 AS _s1
   ON _s1.o_custkey = customer.c_custkey
 ORDER BY
-  total DESC
+  3 DESC
```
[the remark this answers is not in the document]
knassre-bodo: While I was updating the documentation, I also updated the example generated SQL used in the documentation to reflect the most recent version of PyDough (since these are not _automatically_ generated)

pydough/evaluation/evaluate_unqualified.py
knassre-bodo: Because `session.metadata` can be none, but `session.condig` and `session.database` cannot.

tests/test_exploration.py
knassre-bodo: Because the way this test is written doesn't use sessions, it is just calling `pydough.explain` directly on metadata objects. The fetcher gets passed into this function:
```py
graph_name: str = args[0]
collection_name: str | None = args[1]
property_name: str | None = args[2]
def wrapped_test_impl(fetcher: graph_fetcher, verbose: bool):
        graph: GraphMetadata = fetcher(graph_name)
        if collection_name is None:
            return pydough.explain(graph, verbose=verbose)
        elif property_name is None:
            return pydough.explain(graph[collection_name], verbose=verbose)
        else:
            return pydough.explain(
                graph[collection_name][property_name], verbose=verbose
            )
```

pydough/evaluation/evaluate_unqualified.py
```diff
@@ -27,19 +27,47 @@
...
+            raise PyDoughSessionException(
+                "Cannot evaluate Pydough without a metadata graph. "
+                "Please use `session.load_metadata_graph` to attach a graph to the session."
+            )
+        if kwargs:
+            raise ValueError(f"Unexpected keyword arguments: {kwargs}")
```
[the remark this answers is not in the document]
knassre-bodo: They are all required. The difference is that the other two can't be None.

PR #429, opened 2025-09-19
Adjusting how hybrid backrefs are derived to prevent bugs with partitions

knassre-bodo:
TODO: ADD DESCRIPTION


tests/test_technograph.py
```diff
@@ -480,6 +479,147 @@
             ),
             id="country_x_year_analysis",
         ),
+        pytest.param(
+            PyDoughPandasTest(
+                """
```
knassre-bodo, line 484: These two tests are identical. The first one was a reproducer for a bug with correlated partitions, and the second is an alternative way to write it with `CROSS` to avoid much of the correlation/de-correlation.

tests/test_pydough_functions/simple_pydough_functions.py
```diff
@@ -795,7 +795,7 @@ def supplier_pct_national_qty():
         .suppliers.WHERE((account_balance >= 0.0) & CONTAINS(comment, "careful"))
         .CALCULATE(
             supplier_name=name,
-            nation_name=name,
+            nation_name=nation_name,
```
knassre-bodo, line 798: I have no clue why this was working before, but this change ensures that the test _continues_ passing (I guess that before we had a bug which was causing this to work correctly even though it shouldn't have, but my other patches caused that good-bug to go away).

hadia206 (approved):
Looks good to me.

Issue #435, opened 2025-09-24
SQL query optimization error regression


juankx-bodo:
Pydough v1.00.10 is showing a `SQL query optimization error` when generating SQL for querys with:
- [Invalid SQL when a column or table name includes a quoting character](https://github.com/bodo-ai/PyDough/issues/434)
- [Invalid double escape to names already escaped in metadata](https://github.com/bodo-ai/PyDough/issues/433)
- [Invalid use of table alias name when table_name is a SQL reserved word](https://github.com/bodo-ai/PyDough/issues/432)

This `SQL query optimization error` is not happening on v1.00.09

Issue #434, opened 2025-09-24
Invalid SQL when a column or table name includes a quoting character


juankx-bodo:
If a `column_name` or `table_name` includes a quoting character `` ` `` or `"` the SQL generated is invalid.

Issue #432, opened 2025-09-24
Invalid use of table alias name when table_name is a SQL reserved word


juankx-bodo:
When a `table_name` is a reserved word, the alias chosen by Pydough is invalid in `JOIN` and `SELECT` statements if a property from that table is selected.

(1) If there's a table called `cast` that's used in a `JOIN` condition, Pydough will use the same table name as an alias. This creates invalid SQL in the `JOIN` condition because the alias name isn't escaped.

(2) The list of columns in the `SELECT` will also use the unescaped, invalid alias name.

PR #441, opened 2025-10-08
Alias and query optimization error patch

john-sanchez31:
This patch resolves: #435, #432 and #434


demos/metadata/tpch_demo_graph.json
```diff
@@ -817,5 +817,5 @@
     "additional definitions": [],
     "verified pydough analysis": [],
     "extra semantic info": {}
-  },
+  }
 ]
```
knassre-bodo: Add the newline char

pydough/sqlglot/override_qualify.py
```diff
@@ -0,0 +1,288 @@
...
+        for name, source in scope.sources.items():
+            if isinstance(source, exp.Table):
+                pivots = source.args.get("pivots")
+                quoted = None
+                if not source.alias:
+                    # Don't add the pivot's alias to the pivoted table, use the table's name instead
+                    if pivots and pivots[0].alias == name:
+                        name = source.name
+
+                    # CHANGE: preserve quoting of original table name
```
john-sanchez31: This is the only change made in `qualify.py` and `qualify_table.py` from sqlglot

demos/metadata/tpch_demo_graph.json
john-sanchez31: Add new line

pydough/sqlglot/sqlglot_helpers.py
```diff
@@ -45,13 +53,77 @@ def set_glot_alias(expr: SQLGlotExpression, alias: str | None) -> SQLGlotExpress
...
+
+            if isinstance(expr, Identifier):
+                quoted = expr.args.get("quoted", False)
+            elif isinstance(expr, Column):
+                quoted = expr.this.args.get("quoted", False)
+
+        return generate_glot_alias(expr, alias, quoted=quoted)
+
+
+def generate_glot_alias(
```
john-sanchez31: Overridden function of `alias_` defined in `expressions.py`. Some code was omitted because its not required for this specific use.

pydough/sqlglot/sqlglot_helpers.py
```diff
@@ -45,13 +53,77 @@ def set_glot_alias(expr: SQLGlotExpression, alias: str | None) -> SQLGlotExpress
...
+def generate_identifier(name, quoted=None, copy=True):
+    if name is None:
+        return None
+
+    if isinstance(name, Identifier):
+        identifier = maybe_copy(name, copy)
+    elif isinstance(name, str):
+        identifier = Identifier(
+            this=name,
+            quoted=quoted if quoted is not None else False,
```
john-sanchez31: Before: 
```py 
quoted=not SAFE_IDENTIFIER_RE.match(name) if quoted is None else quoted,
```

pydough/sqlglot/sqlglot_relational_expression_visitor.py
```diff
@@ -372,11 +372,35 @@ def make_sqlglot_column(
...
+            # Strip one layer of surrounding quotes or backticks from the column
+            # name. For example: ""column name"" -> "column name"
+            #
+            # This ensures that the internal representation does not include
+            # redundant quote characters, while marking the name as requiring
+            # quoting in the generated SQL.
+            column_name = column_name[1:-1]
+            # Deteles the escaped quotes inside the name because sqlglot takes
+            # care of re-escaping them when generating SQL.
+            column_name = column_name.replace('""', '"')
```
john-sanchez31: sqlglot escapes automatically the quotes inside the column's name but I doesn't escape quotes in table names. For consistency and to pass the `valid_sql_identifier` from `error_utils.py` those escape quotes are require in the metadata for columns as well. But this replaces it with just one because sqlglot takes care of this at the end. For example:
```
metadata: "column name": "\"`name\"\"[\""
here we delete the extra quote: `name\"[ and set quoted = True
sqlglot's result: "`name""[" (which is the correct answer)
```

pydough/sqlglot/sqlglot_helpers.py
```diff
@@ -45,13 +53,77 @@ def set_glot_alias(expr: SQLGlotExpression, alias: str | None) -> SQLGlotExpress
...
-        return expr.as_(alias)
+        quoted: bool = False
+
+        if (
+            (alias.startswith('"') and alias.endswith('"'))
+            or (alias.startswith("'") and alias.endswith("'"))
+            or (alias.startswith("`") and alias.endswith("`"))
+        ):
+            alias = alias[1:-1]
+            alias = alias.replace('""', '"')
```
juankx-bodo: I send some comments about this in slack. I think we should do:
- if `startswith('"') and endswith('"')`  >> replace from `""`  to `"`
- if ```startswith('`') and endswith('`')``` >> replace from ``` `` ```  to ``` ` ```
- No need to check if `(alias.startswith("'") and alias.endswith("'")` because this should fail the SQL name validation.

pydough/sqlglot/sqlglot_helpers.py
```diff
@@ -45,13 +53,77 @@ def set_glot_alias(expr: SQLGlotExpression, alias: str | None) -> SQLGlotExpress
...
+
+            if isinstance(expr, Identifier):
+                quoted = expr.args.get("quoted", False)
+            elif isinstance(expr, Column):
+                quoted = expr.this.args.get("quoted", False)
```
juankx-bodo: This is because of the issue you told me about the behavior of sqlglot with column vs table names right?

pydough/sqlglot/sqlglot_helpers.py
```diff
@@ -45,13 +53,77 @@ def set_glot_alias(expr: SQLGlotExpression, alias: str | None) -> SQLGlotExpress
...
+    # [1]: https://cloud.google.com/bigquery/docs/reference/standard-sql/window-function-calls
+
+    if "alias" in exp.arg_types and not isinstance(exp, Window):
+        exp.set("alias", alias)
+        return exp
+
+    return SQLGlotAlias(this=exp, alias=alias)
+
+
+def generate_identifier(name, quoted=None, copy=True):
```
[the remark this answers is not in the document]
juankx-bodo: LGTM because these identifiers should be enclosed in the metadata. Because this is a code removal change, maybe add a comment in the code about the change.

pydough/sqlglot/sqlglot_helpers.py
juankx-bodo: We can mention the issue # in the code comment to be clear about the root cause of the change. What do you think @knassre-bodo ?

pydough/sqlglot/sqlglot_relational_expression_visitor.py
john-sanchez31: Yes absolutely, the thing is where should I place this new function that can be used in `sqlglot_relational_expression_visitor.py` and `sqlglot_helpers.py`?

pydough/sqlglot/sqlglot_helpers.py
```diff
@@ -45,13 +53,99 @@ def set_glot_alias(expr: SQLGlotExpression, alias: str | None) -> SQLGlotExpress
...
+        `expr`: The expression to wrap in an alias.
+        `alias`: The alias to use.
+
+    Returns:
+        The generated Alias expression.
+    """
+    exp = maybe_parse(expr, dialect=None, copy=True)
+    alias = generate_identifier(alias, quoted=quoted)
+
+    # Part of this code is ommitted because por this particular case the table
```
knassre-bodo:
```suggestion
    # Part of this code is omitted because for this particular case the table
```

knassre-bodo (commented):
A few comments, but overall LGTM. Great job @john-sanchez31 at investigating this! I'll let Hadia or Juan Carlos be the approving review once you have e2e tests for the other dialects

pydough/sqlglot/execute_relational.py
```diff
@@ -102,7 +104,7 @@ def apply_sqlglot_optimizer(
 
     # Apply each rule explicitly with appropriate kwargs
 
-    kwargs = {
+    kwargs: dict[str, Any] = {
```
hadia206: why `Any`? shouldn't it be boolean?

pydough/sqlglot/sqlglot_helpers.py
```diff
@@ -45,13 +53,99 @@ def set_glot_alias(expr: SQLGlotExpression, alias: str | None) -> SQLGlotExpress
...
+    if name is None:
+        return None
+
+    if isinstance(name, Identifier):
+        identifier = maybe_copy(name, copy)
+    elif isinstance(name, str):
+        identifier = Identifier(
+            this=name,
+            # originally was:
+            # quoted=not SAFE_IDENTIFIER_RE.match(name) if quoted is None else quoted,
```
hadia206: Reminder to add PyDough change comment and why you did this change

tests/test_pipeline_custom_datasets.py
```diff
@@ -343,13 +348,11 @@ def test_pipeline_e2e_custom_datasets(
     custom_datasets_test_data: PyDoughPandasTest,
     get_test_graph_by_name: graph_fetcher,
     sqlite_custom_datasets_connection: DatabaseContext,
-    coerce_types=True,
```
hadia206: why was this removed?

hadia206 (commented):
Nice work John!
Overall looks good.
I left minor comments/questions.

pydough/sqlglot/execute_relational.py
john-sanchez31: I had to add this because I was having problems with pre-commit. `Any` type is because right below a dialect could be added which seems to `Any` type as well.

pydough/sqlglot/sqlglot_helpers.py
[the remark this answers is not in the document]
john-sanchez31: I meant this concrete call or usage of the function. Originally this function among other params received `table` but those are not used or passed from this place that why I omitted them.

juankx-bodo:
Remember to run CI

PR #443, opened 2025-10-09
Adding optimization rewrite pass to utilize server with information about masked columns

knassre-bodo:
Augmenting relational optimization to rewrite expressions containing an UNMASK operator when a server is mounted to the PyDough session (and the environment variable is activated):
- When this is the case, two additional shuttles/visitors are added to the `additional_shuttles` list, before the masking literal comparisons shuttle.
- The first, `MaskServerCandidateVisitor` is a no-op shuttle that just traverses the entire tree to find expressions that can potentially be rewritten and adds them to a pool. It also maintains other datastructures about the heirarchy fo expressions.
- The second, `MaskServerRewriteShuttle` looks for expressions in the candidate shuttle's pool, and once it finds one it sends every candidate in the pool into a batch request to the mask server, processing the output results to create the new relational node. The candidate pool is then emptied so future invocations will not re-do the same batch calculation. Before sending the full batch, it first does dry runs to algorithmically determine which requests to truly send.

### DIFF
--- a/.github/workflows/sf_masked_testing.yml
+++ b/.github/workflows/sf_masked_testing.yml
@@ -41,6 +41,7 @@ jobs:
       SF_NONE_USERNAME: ${{ secrets.SF_NONE_USERNAME }}
       SF_NONE_PASSWORD: ${{ secrets.SF_NONE_PASSWORD }}
       SF_MASKED_ACCOUNT: ${{ secrets.SF_MASKED_ACCOUNT }}
+      PYDOUGH_MASK_SERVER_PATH: ${{ secrets.PYDOUGH_MASK_SERVER_PATH }}
 
     steps:
       - uses: actions/checkout@v4
--- a/tests/test_metadata_errors.py
+++ b/tests/test_metadata_errors.py
@@ -875,7 +875,7 @@ def test_missing_property(get_sample_graph: graph_fetcher) -> None:
         pytest.param(
             "MASKED_TABLE_COLUMN_EXTRA_FIELDS",
             re.escape(
-                "masked table column property 'name' of simple table collection 'customers' in graph 'MASKED_TABLE_COLUMN_EXTRA_FIELDS' must be a JSON object containing no fields except for ['column name', 'data type', 'description', 'extra semantic info', 'name', 'protect protocol', 'protected data type', 'sample values', 'server masked', 'synonyms', 'type', 'unprotect protocol']"
+                "masked table column property 'name' of simple table collection 'customers' in graph 'MASKED_TABLE_COLUMN_EXTRA_FIELDS' must be a JSON object containing no fields except for ['column name', 'data type', 'description', 'extra semantic info', 'name', 'protect protocol', 'protected data type', 'sample values', 'server dataset id', 'server masked', 'synonyms', 'type', 'unprotect protocol']"
             ),
             id="MASKED_TABLE_COLUMN_EXTRA_FIELDS",
         ),


pydough/conversion/mask_server_candidate_visitor.py
```diff
@@ -0,0 +1,312 @@
...
+        2. Either None, or the linear serialization of the input expression as
+           a list, where invocations of UNMASK(input_expr) are replaced with
+           the token "__col__".
+        """
+
+    def reset(self):
+        self.stack.clear()
+
+    def visit_call_expression(self, expr: CallExpression) -> None:
+        # TODO: ADD COMMENTS
```
knassre-bodo:
```suggestion
```

pydough/conversion/mask_server_candidate_visitor.py
```diff
@@ -0,0 +1,312 @@
...
+        # Besides ISIN, if the function call is not one of the operators that
+        # can be handled by the Mask Server, return None since it cannot be
+        # converted.
+        elif call.op not in self.OPERATORS_TO_SERVER_NAMES:
+            return None
+
+        # Build up the list with the first two entries: the name of the function
+        # call operator, and the number of inputs to the function call.
+        result: list[str | int | float | None | bool] = []
+        operator_name = self.OPERATORS_TO_SERVER_NAMES[call.op]
```
john-sanchez31: Missing type hint

pydough/conversion/mask_server_candidate_visitor.py
```diff
@@ -0,0 +1,312 @@
...
+                self.convert_literal_to_server_expression(
+                    LiteralExpression(v, UnknownType())
+                )
+            )
+            if literal_list is None:
+                return None
+            in_list.extend(literal_list)
+
+        # The result list is:
+        # 1. The operator name "IN"
```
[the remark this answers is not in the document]
knassre-bodo: This is how we convert PyDough relational terms to the mask server terms. Specifically, we are converting a call of `ISIN` to to the mask server operator `IN`. If the PyDough code is `NOT(ISIN(...))`, then the list returned by this function will get wrapped in a `NOT` call.

tests/test_mock_mask_server.py
```diff
@@ -22,27 +22,27 @@
...
-                    table_path="srv.db.tbl",
+                    table_path="db.tbl",
                     column_name="col",
                     expression=["OR", 2, "__col__", 5],
                 ),
                 MaskServerInput(
-                    table_path="srv.db.orders",
+                    table_path="db.orders",
                     column_name="order_date",
                     expression=["BETWEEN", 3, "__col__", "2025-01-01", "2025-02-01"],
```
juankx-bodo: We need to test the use of `QUOTE`, for values using predicate reserved words like an OP name or `__col__`

juankx-bodo (commented):
I'll let the approval to @hadia206 and @john-sanchez31

pydough/conversion/mask_server_candidate_visitor.py
```diff
@@ -0,0 +1,688 @@
...
+        step_literal = input_exprs[3]
+        if (
+            start_literal is None
+            or stop_literal is None
+            or len(start_literal) != 1
+            or len(stop_literal) != 1
+            or step_literal not in ([1], ["NULL"])
+        ):
+            return None
+        print(start_literal, stop_literal, step_literal)
```
john-sanchez31: Is this `print` part of the code? If not it should be deleted

pydough/conversion/mask_server_candidate_visitor.py
```diff
@@ -0,0 +1,688 @@
...
+        #    versus the number of elements in the list.
+        # 3. The linear serialization of the first input expression.
+        # 4. The unfolded elements of the literal list from the second input.
+        result: list[str | int | float | None | bool] = ["IN"]
+        result.append(len(inputs[1].value) + 1)
+        result.extend(input_exprs[0])
+        result.extend(in_list)
+        return result
+
+    def convert_slice_call_to_server_expression(
```
john-sanchez31: As my understanding this function only supports positives values, but PyDough can support slicing with negative values. How does that work? I assume those values wouldn't enter this function so is it managed somewhere else?

pydough/conversion/mask_server_candidate_visitor.py
```diff
@@ -0,0 +1,688 @@
...
+        return result
+
+    def convert_dateadd_call_to_server_expression(
+        self,
+        input_expr: list[str | int | float | None | bool],
+        sign_str: str,
+        amount: int,
+        unit_str: str,
+    ) -> list[str | int | float | None | bool] | None:
+        """
```
john-sanchez31: Add a `Returns` section

john-sanchez31 (commented):
Left some comments, after those are fixed or clarified I'll approve but overall LGTM!

pydough/mask_server/mask_server.py
```diff
@@ -242,24 +302,37 @@ def generate_result(self, response: dict) -> list[MaskServerOutput]:
...
-
-                if response_case in (
-                    MaskServerResponse.IN_ARRAY,
-                    MaskServerResponse.NOT_IN_ARRAY,
-                ):
-                    payload = materialization.get("values", [])
+                response: dict = item.get("response", None)
+                if response is None:
+                    # In this case, use a dummy value as a default to indicate
+                    # the dry run was successful
```
john-sanchez31: Did you mean `to indicate the dry run was unsuccessful`?

pydough/mask_server/mask_server_candidate_visitor.py
```diff
@@ -0,0 +1,703 @@
...
+        2. Either None, or the linear serialization of the input expression as
+           a list, where invocations of UNMASK(input_expr) are replaced with
+           the token "__col__".
+        """
+
+        self.heritage_tree: dict[
+            RelationalExpression, set[RelationalExpression | None]
+        ] = {}
+        """
+        TODO
```
john-sanchez31: Reminder

pydough/mask_server/mask_server_candidate_visitor.py
```diff
@@ -0,0 +1,703 @@
...
+        """
+
+        self.ancestry_stack: list[RelationalExpression | None] = [None]
```
john-sanchez31: Remenber the `docstring` below

pydough/mask_server/min_cover_set.py
```diff
@@ -0,0 +1,43 @@
+"""
```
john-sanchez31: Reminder

pydough/conversion/relational_converter.py
```diff
@@ -860,7 +865,9 @@ def build_simple_table_scan(
                     )
                     unmask_columns[name] = CallExpression(
                         pydop.MaskedExpressionFunctionOperator(
-                            hybrid_expr.column.column_property, True
+                            hybrid_expr.column.column_property,
+                            node.collection.collection.table_path,
```
juankx-bodo: Is this the reason why we need to use the full table path in metadata?

pydough/conversion/relational_converter.py
```diff
@@ -1707,10 +1722,19 @@ def convert_ast_to_relational(
...
     # Invoke the optimization procedures on the result to clean up the tree.
-    additional_shuttles: list[RelationalExpressionShuttle] = []
+    additional_shuttles: list[
+        RelationalExpressionShuttle | RelationalExpressionVisitor
+    ] = []
     # Add the mask literal comparison shuttle if the environment variable
-    # PYDOUGH_ENABLE_MASK_REWRITES is set to 1.
+    # PYDOUGH_ENABLE_MASK_REWRITES is set to 1. If a masking rewrite server has
+    # been attached to the session, include the shuttles for that as well.
     if os.getenv("PYDOUGH_ENABLE_MASK_REWRITES") == "1":
```
juankx-bodo: Is there a reson why `PYDOUGH_ENABLE_MASK_REWRITES` is not in `PyDoughConfigs`?

pydough/mask_server/mask_server.py
```diff
@@ -135,100 +142,153 @@ def simplify_simple_expression_batch(
...
+        if dry_run:
+            pyd_logger.info(
+                f"Batch request (dry run) to Mask Server ({len(batch)} items):"
+            )
+        else:
+            pyd_logger.info(f"Batch request to Mask Server ({len(batch)} items):")
+        for idx, item in enumerate(batch):
+            pyd_logger.info(
+                f"({idx + 1}) {item.table_path}.{item.column_name}: {item.expression}"
+            )
```
juankx-bodo: Should this log entry be `debug` level instead of `info`?

pydough/mask_server/mask_server.py
```diff
@@ -135,100 +142,153 @@ def simplify_simple_expression_batch(
...
 
         path: str = "v1/predicates/batch-evaluate"
         method: RequestMethod = RequestMethod.POST
-
-        request: ServerRequest = self.generate_request(batch, path, method)
-
+        request: ServerRequest = self.generate_request(
+            batch, path, method, dry_run, hard_limit
+        )
         response_json = self.connection.send_server_request(request)
```
juankx-bodo: In case of a predicate_server failure, users will not be able to query the database at all. Not even with the MASK functions. This could be a critical point of failure for the system.

pydough/mask_server/mask_server.py
```diff
@@ -135,100 +142,153 @@ def simplify_simple_expression_batch(
...
             "expression_format": {"name": "linear", "version": "0.2.0"},
+            "hard_limit": hard_limit,
         }
 
         for item in batch:
             evaluate_request: dict = {
-                "column_reference": f"{item.table_path}.{item.column_name}",
+                "column_ref": {
+                    "kind": "fqn",
+                    "value": f"{self.server_address}.{item.table_path}.{item.column_name}",
```
juankx-bodo: separator should be `"/"`. `item.table_path` is a composed name with elements separated by `"."`. Any element could be enclosed with `double-quotes` or `backtick` and have `"."` as part of the element name. Additionally, any character in the name equals to the enclosure char will be escaped using the same character twice.

juankx-bodo: We could wait to see the real thing implementation before this kind of changes.

pydough/mask_server/mask_server.py
```diff
@@ -135,100 +142,153 @@ def simplify_simple_expression_batch(
...
         result: list[MaskServerOutput] = self.generate_result(response_json)
 
         return result
 
     def generate_request(
-        self, batch: list[MaskServerInput], path: str, method: RequestMethod
+        self,
+        batch: list[MaskServerInput],
+        path: str,
+        method: RequestMethod,
```
juankx-bodo: Including `path` and `method` in parameters looks like an attempt of doing `generate_request()` more general. However, due to all other specific parameters and actions I think this method is very specific for `batch-evaluate`. Maybe `path` and `method` could be class properties since them will not change for this method. If more request methods are required in future those paths and methods could also be part of the class.

pydough/mask_server/mask_server_candidate_visitor.py
```diff
@@ -0,0 +1,703 @@
...
+        # 1. The operator name "IN"
+        # 2. The total number of arguments, including the element to check
+        #    versus the number of elements in the list.
+        # 3. The linear serialization of the first input expression.
+        # 4. The unfolded elements of the literal list from the second input.
+        result: list[str | int | float | None | bool] = ["IN"]
+        result.append(len(inputs[1].value) + 1)
+        result.extend(input_exprs[0])
+        result.extend(in_list)
+        return result
```
juankx-bodo: Is this OK? Shouldn't it be `[input_exprs[0], len(inputs[1].value) + 1, in_list]`?

pydough/mask_server/mask_server.py
knassre-bodo: 🤷 Rn I'm keeping everything the same logging level for simplicity. We can revise down if we think it is appropriate.

pydough/mask_server/mask_server.py
```diff
@@ -135,100 +142,153 @@ def simplify_simple_expression_batch(
...
         Returns:
             A server request including payload to be sent.
 
         Example payload:
+        ```
         {
             "items": [
                 {
-                    "column_reference": "srv.db.tbl.col",
+                    "column_ref": {"kind": "fqn", "value": "srv.db.schema.table.name"},
```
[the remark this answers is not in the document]
knassre-bodo: Good catch

pydough/mask_server/mask_server.py
```diff
@@ -135,100 +142,153 @@ def simplify_simple_expression_batch(
...
+        else:
+            pyd_logger.info(f"Batch request to Mask Server ({len(batch)} items):")
+        for idx, item in enumerate(batch):
+            pyd_logger.info(
+                f"({idx + 1}) {item.table_path}.{item.column_name}: {item.expression}"
+            )
+
         assert batch != [], "Batch cannot be empty."
 
         path: str = "v1/predicates/batch-evaluate"
```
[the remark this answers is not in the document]
knassre-bodo: Good point, moved into a class var of `MaskServerInfo` that gets passed into the `ServerRequest` by `generate_request`

pydough/mask_server/mask_server.py
knassre-bodo: I think method doesn't need to be a class property, can just get baked into the method's construciton of an `ServerRequest` instance.

pydough/mask_server/mask_server_candidate_visitor.py
```diff
@@ -0,0 +1,688 @@
...
+        2. The input expression that is being unmasked.
+        3. The linear serialization of the entire expression as a list, where
+           invocations of UNMASK(input_expr) are replaced with the token
+           "__col__".
+        """
+
+        self.processed_candidates: set[RelationalExpression] = set()
+        """
+        The set of all relational expressions that have already been added to
+        the candidate pool at lest once. This is used to avoid adding the same
```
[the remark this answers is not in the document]
john-sanchez31: Reminder

pydough/mask_server/mask_server_candidate_visitor.py
```diff
@@ -0,0 +1,688 @@
...
+                return None
+
+        result.append(start_int)
+        result.append(length_int)
+        return result
+
+    def convert_join_strings_call_to_server_expression(
+        self, input_exprs: list[list[str | int | float | None | bool] | None]
+    ) -> list[str | int | float | None | bool] | None:
+        """
```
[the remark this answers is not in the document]
john-sanchez31: Reminder

pydough/errors/error_utils.py
```diff
@@ -203,7 +203,8 @@ def __init__(self):
             "sql_keyword": "must have a SQL name that is not a reserved word",
         }
 
-    def _split_identifier(self, name: str) -> list[str]:
+    @staticmethod
```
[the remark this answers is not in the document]
knassre-bodo: So it can be used from other files w/o creating an instance of this class

pydough/mask_server/mask_server_rewrite_shuttle.py
```diff
@@ -0,0 +1,323 @@
...
+        # lists, the batch and ancillary info, remain in sync by index so they
+        # can be zipped together later.
+        for expr, (
+            mask_op,
+            input_expr,
+            expression_list,
+        ) in self.candidate_visitor.candidate_pool.items():
+            ancillary_info.append((expr, input_expr))
+            assert mask_op.masking_metadata.server_masked
+            assert mask_op.masking_metadata.server_dataset_id is not None
```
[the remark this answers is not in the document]
knassre-bodo: I'm treating these as assertions since if these conditions are _not_ true, then it shouldn't have been placed in the candidate pool in the first place.

pydough/mask_server/mask_server_candidate_visitor.py
```diff
@@ -0,0 +1,703 @@
...
+                result: list[str | int | float | None | bool] = []
+                operator_name: str = self.OPERATORS_TO_SERVER_NAMES[call.op]
+                result.append(operator_name)
+                result.append(len(call.inputs))
+                # For each input to the function call, append its linear
+                # serialization to the result list. We know they are not None
+                # from the earlier check.
+                for inp in input_exprs:
+                    assert inp is not None
+                    result.extend(inp)
```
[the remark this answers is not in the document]
knassre-bodo: This is no longer a feature of the API.

PR #445, opened 2025-10-23
Load additional definitions into metadata

juankx-bodo:
This PR is to load `additional definitions` into metadata. The markdown generator tool uses the metadata clases to generate the markdown. Only the fields loaded can be included.


knassre-bodo:
Closed due to pivoting to an alternative path.

PR #446, opened 2025-10-23
MySQL & Postgres image fix

john-sanchez31:
Minor adjustments in the waiting time in `conftest.py`. This change allows local testing and ci use the new image for testing created with MySQL and Postgres workflow respectively.


tests/conftest.py
```diff
@@ -907,29 +908,6 @@ def mysql_conn_db_context(
     # The first time, set up the defog data
     import mysql.connector as mysql_connector
 
-    mysql_username = os.getenv("MYSQL_USERNAME")
```
john-sanchez31: This is not longer needed, defog is already included in the new image

knassre-bodo (approved):

PR #447, opened 2025-10-27
Adding warning logs for potentially unsafe masking/unmasking calls

knassre-bodo:
Adds a relational visitor pass at the end of the relational optimizer designed to traverse the final relational tree and identify any mask/unmask calls that would cause a critical logical error if the user does not have permission. The pass finds all mask/unmask calls that, at some point in the plan, are used in a sensitive manner so one of the following will happen if the user cannot make the mask/unmask plan:
- The answer will have the wrong number of rows 
- The answer will include the wrong subset of rows
- The answer will have the rows in the wrong order

Those three types of critical errors will happen if the mask/unmask call without sufficient permissions is eventually used in one of the following:
- The condition of a join
- The condition of a filter
- An ordering key for a limit
- An ordering key for the relational root (ordering the final table)

If the values are garbled in one of the following ways, it is not considered a critical failure unless the garbled value is used in one of the sensitive manners described above:
- Calling a function/aggregation/window function on mask/unmask calls
- Using mask/unmask on the data presented as the final answer
- Using mask/unmask calls as the partition or order keys for a window function

Since PyDough does not (currently) have a way to determine whether a user has permission to mask/unmask a column, the visitor will identify all sensitive dependencies that would cause a critical error **if** the user does not have permission. Then, it will dump all of these dependencies via logger warnings in the following format (per column):

```
WARNING  pydough.logger.logger:masking_critical_detection_visitor.py:148 Query will not produce a valid output unless user has permission to mask column `CRBNK.CUSTOMERS.c_fname`
WARNING  pydough.logger.logger:masking_critical_detection_visitor.py:152 Query will not produce a valid output unless user has permission to unmask column `CRBNK.CUSTOMERS.c_key`
WARNING  pydough.logger.logger:masking_critical_detection_visitor.py:152 Query will not produce a valid output unless user has permission to unmask column `CRBNK.ACCOUNTS.a_key`
```


pydough/conversion/masking_critical_detection_visitor.py
```diff
@@ -0,0 +1,283 @@
...
+
+    def __init__(self) -> None:
+        self.critical_mask_columns: set[str] = set()
+        """
+        The set of fully qualified column names where a MASK operation on the
+        column is critical to the output of the query.
+        """
+
+        self.critical_unmask_columns: set[str] = set()
+
```
john-sanchez31:
```suggestion
```

juankx-bodo (commented):
I will let the final approval to @john-sanchez31 and @hadia206

tests/test_masked_sqlite.py
```diff
@@ -1374,14 +1506,12 @@ def test_cryptbank_mask_server_logging(
...
+    with redirect_stdout(io.StringIO()):
+        to_sql(root, metadata=graph, mask_server=mock_server_info)
 
     # Retrieve the output from the captured logger output, while capturing
     # stdout to avoid polluting the console with logging calls
-    with redirect_stdout(io.StringIO()):
-        batch_requests_made: list[set[str]] = extract_batch_requests_from_logs(
-            caplog.text
-        )
+    batch_requests_made: list[set[str]] = extract_batch_requests_from_logs(caplog.text)
```
hadia206: why remove `with redirect_stdout` here?

PR #450, opened 2025-10-30
Delete redundant joins by substitution equi-join keys for their mirror to render one side pruneable

knassre-bodo:
Adds an optimization pass which substitutes columns used by a join from one side to the other when passing through a join key. This is done under a certain set of conditions to ensure that the one side of the join no longer has any of its columns being used, thus ensuring that under those same conditions, the join can be optimized out by the column pruner.


pydough/conversion/hybrid_tree.py
```diff
@@ -858,7 +858,7 @@ def is_singular(self) -> bool:
             case _:
                 return False
         # The current level is fine, so check any levels above it next.
-        return True if self.parent is None else self.parent.always_exists()
+        return True if self.parent is None else self.parent.is_singular()
```
hadia206: why this change?

pydough/conversion/join_key_substitution.py
```diff
@@ -0,0 +1,101 @@
...
+                lhs_keys: set[ColumnReference] = set(lhs_keys_list)
+                rhs_keys: set[ColumnReference] = set(rhs_keys_list)
+                col_finder = ColumnReferenceFinder()
+                for value in join.columns.values():
+                    value.accept(col_finder)
+                col_refs: set[ColumnReference] = col_finder.get_column_references()
+                lhs_refs = {
+                    ref
+                    for ref in col_refs
+                    if ref.input_name == join.default_input_aliases[0]
```
hadia206: For my understanding, how the `default_input_aliases` identify LHS refs?
I thought it includes both LHS and RHS inputs

hadia206 (approved):
Looks good to me.
Only have some questions

pydough/conversion/hybrid_tree.py
knassre-bodo: This was a bug in the previous version that I happened to notice. This function is checking if a hybrid tree is singular with regards to its parent context, which mis true if the current level is singular + all levels above it are also singular.

PR #451, opened 2025-10-31
Update synthea and wdi metadata

juankx-bodo:
Update metadata with AI enriched version based on latest metadata generation tool version
Sample values and all definitions are based on the actual BIRD full dataset.


tests/test_metadata/world_development_indicators_graph.json
```diff
@@ -4,1104 +4,1390 @@
...
+            "synonyms": [
+              "Seriescode"
+            ]
           },
           {
-            "name": "Year",
+            "name": "year_",
             "type": "table column",
-            "column name": "Year",
+            "column name": "\"Year\"",
```
john-sanchez31: Why is quoted?

hadia206:
I agree with John, do we need that after the changes he made in this [PR](https://github.com/bodo-ai/PyDough/pull/457)?

PR #456, opened 2025-11-05
Move defog update to its own job

hadia206:
Previously, the call to data update (`DEFOG_DAILY_UPDATE`) was triggered from `sf_conn_db_context` fixture that ran three times, once for each Python version in the test matrix. 

This could lead to duplicate data if all three versions executed the procedure simultaneously. 

This PR moves the update step into its own CI job, ensuring the procedure runs only once per CI run. Then, Snowflake tests now run as usual after the update.


tests/gen_data/sf_task.sql
```diff
@@ -527,6 +527,75 @@ VALUES
 ('TX054', 'C018', 'T004', DATE_TRUNC('month', CURRENT_DATE) - INTERVAL '2 months' + INTERVAL '4 days', 'sell', 30, 180.00, 5400.00, 'USD', 27.00, 5.00, 'KP054', TO_CHAR(DATE_TRUNC('month', CURRENT_DATE) - INTERVAL '2 months' + INTERVAL '4 days', '%Y%m%d %H:%i:%s'), 'success'),
 ('TX055', 'C019', 'T005', DATE_TRUNC('month', CURRENT_DATE) - INTERVAL '1 month' + INTERVAL '5 days', 'buy', 10, 2500.00, 25000.00, 'USD', 125.00, 15.00, 'KP055', TO_CHAR(DATE_TRUNC('month', CURRENT_DATE) - INTERVAL '1 month' + INTERVAL '5 days', '%Y%m%d %H:%i:%s'), 'success'),
 ('TX056', 'C002', 'T006', DATE_TRUNC('month', CURRENT_DATE) + INTERVAL '1 day', 'sell', 20, 200.00, 4000.00, 'USD', 20.00, 10.00, 'KP056', TO_CHAR(DATE_TRUNC('month', CURRENT_DATE) + INTERVAL '1 day', '%Y%m%d %H:%i:%s'), 'success');
+-------------------------------------------------------------------------------
```
hadia206: Some Inserts that are in the procedure but not included in that file

knassre-bodo (approved):

PR #457, opened 2025-11-07
Adding custom & s3 tests infrastructure

john-sanchez31:
Includes the infrastructure for testing custom defined locally and s3 datasets from the bucket `llm-fixtures`. Adding the name of the dataset in `S3_DATASETS` sets up  the needed files (creating the `.db` file from `.sql` file). `S3_DATASETS_SCRIPTS` define the s3 datasets that uses a `.sql` file to create the `.db` file. S3 datasets not included in `S3_DATASETS_SCRIPTS` will be downloaded from s3.

This setup allows local and ci testing using the tests defined in `test_pipeline_custom_datasets.py` and `test_pipeline_s3_datasets.py`.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -642,7 +642,7 @@ def sqlite_cryptbank_connection() -> DatabaseContext:
 def sqlite_custom_datasets_connection() -> Callable[[str], DatabaseContext]:
     """
     This fixture is used to connect to the sqlite database of the custom datasets.
-    Returns a DatabaseContext for the MySQL TPCH database.
+    Returns a DatabaseContext for the given custom database name.
     """
     custom_datasets_dir: str = "tests/gen_data"
     # Setup the directory to be the main PyDough directory.


tests/conftest.py
```diff
@@ -643,45 +638,179 @@ def sqlite_cryptbank_connection() -> DatabaseContext:
...
+            # setting up with script
+            # assuming the metadata is already in the repo
+            init_sql = f"{data_folder}/{scripts[dataset]}.sql"
+            subprocess.run(f"sqlite3 {db_file} < {init_sql}", shell=True, check=True)
+        else:
+            # download from s3
+            local_metadata_path: str = f"{metadata_folder}/{dataset}_graph.json"
+
+            key_data: str = f"data/{dataset}.db"
+            key_metadata: str = f"metadata/{dataset}.json"
```
hadia206: nit: move newline to be after key_metadata definition
```suggestion
           key_data: str = f"data/{dataset}.db"
            key_metadata: str = f"metadata/{dataset}.json"
            
```

pyproject.toml
```diff
@@ -43,6 +43,7 @@ snowflake = ["snowflake-connector-python[pandas]"]
 mysql = ["mysql-connector-python"]
 postgres = ["psycopg2-binary"]
 server = ["fastapi", "httpx", "uvicorn"]
+boto3 = ["boto3"]
```
hadia206: This should be part of the `dev-dependencies`.
It's only used in testing, not related to any PyDough functionality.

tests/conftest.py
```diff
@@ -643,45 +638,179 @@ def sqlite_cryptbank_connection() -> DatabaseContext:
...
+            except Exception as e:
+                print(f"An error occurred: {e}")
+
+        database_path = f"{db_folder}/{dataset}.db"
+        try:
+            os.remove(database_path)
+        except FileNotFoundError:
+            print(f"Error: File '{database_path}' not found.")
+        except Exception as e:
+            print(f"An error occurred: {e}")
```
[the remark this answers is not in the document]
juankx-bodo: I think we should keep db files from "local" datasets, like those created from `CUSTOM_DATASETS_SCRIPTS`. It helps to run `pdunit` once and then you can use those db files with jupyter when you need to reproduce a new bug, for example.

tests/test_pipeline_custom_datasets.py
```diff
@@ -305,9 +323,10 @@ def custom_datasets_test_data(request) -> PyDoughPandasTest:
     return request.param
 
 
+@pytest.mark.custom
```
juankx-bodo: We should use `@pytest.mark.custom` only for s3 datasets, that's why I was thinking on it like `@pytest.mark.s3_datasets` instead. Local datasets like `keywords` should be tested always in the same way `defog` and `tpch` do.

tests/conftest.py
```diff
@@ -643,45 +638,179 @@ def sqlite_cryptbank_connection() -> DatabaseContext:
...
 
 
+CUSTOM_DATASETS = ["synthea", "soccer_2016", "wdi", "keywords"]
+"""
+    Contains the name of all the custom datasets that will be used for testing.
+    This includes the datasets from S3 and initialized with a .sql file.
+"""
+CUSTOM_DATASETS_SCRIPTS = {
+    "keywords": "init_reserved_words_sqlite",
+    "wdi": "init_world_indicators_sqlite",
```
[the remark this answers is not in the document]
john-sanchez31: Yes, but these tests are using the local version (not the 2gb version) so It's a custom version for our testing. This variable determine which datasets are build from local scripts and which are downloaded from s3 that's why wdi is included.

tests/conftest.py
```diff
@@ -643,45 +638,179 @@ def sqlite_cryptbank_connection() -> DatabaseContext:
...
+    when the script is executed the metadata must be created manually for testing
+    """
+
+    bucket: str = "llm-fixtures"
+
+    for dataset in datasets:
+        db_file: str = f"{data_folder}/{dataset}.db"
+
+        if dataset in scripts:
+            # setting up with script
```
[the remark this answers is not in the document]
john-sanchez31: No sure if I understand the question. As mentioned before this variable is used to determine which dataset are build from the local `.sql` and which ones are downloaded from s3

tests/conftest.py
```diff
@@ -643,45 +638,179 @@ def sqlite_cryptbank_connection() -> DatabaseContext:
...
+            os.remove(database_path)
+        except FileNotFoundError:
+            print(f"Error: File '{database_path}' not found.")
+        except Exception as e:
+            print(f"An error occurred: {e}")
+
+
 @pytest.fixture(scope="session")
-def sqlite_custom_datasets_connection() -> DatabaseContext:
+def custom_datasets_setup():
```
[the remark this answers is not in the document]
john-sanchez31: Yes s3 datasets and custom local datasets are in what we call custom datasets, those tests run with `[run all]` or `[run custom]`

tests/test_pipeline_custom_datasets.py
john-sanchez31: I think both `keywords` and `s3_datesets` are on the same category. Nether of both are part of `defog` or `tpch` and both are used for edge cases or bug testing that's why there is custom datasets. I though the new flag was created just for that so we don't have to run those every time.

john-sanchez31: But in that case in what type of tests or category falls `keywords`? What about wdi? Since it's using the local version of it not the s3 version (it's more than 2gb)

tests/conftest.py
```diff
@@ -202,21 +204,28 @@ def sample_graph_names(request) -> str:
 
 
 @pytest.fixture(scope="session")
-def get_test_graph_by_name() -> graph_fetcher:
+def get_s3_datasets_graph(s3_datasets_setup) -> graph_fetcher:
```
knassre-bodo: NIT: if this fixture depends on `s3_datasets_setup`, then place its definition right below `s3_datasets_setup`

tests/conftest.py
```diff
@@ -644,44 +653,209 @@ def sqlite_cryptbank_connection() -> DatabaseContext:
...
     )
-    commands.append(rm_command)
-    # Add one sqlite3 command per dataset
-    for _, db_file, init_sql in SQLite_datasets:
-        commands.append(f"sqlite3 {db_file} < {init_sql}")
-    # Get the shell commands required to re-create all the db files
-    shell_cmd: str = "; ".join(commands)
+    print("Datasets downloaded")
+    yield
+    print("\nRemoving datasets")
```
knassre-bodo: Should we ditch the prints?

tests/conftest.py
```diff
@@ -644,44 +653,209 @@ def sqlite_cryptbank_connection() -> DatabaseContext:
 
 
 @pytest.fixture(scope="session")
-def sqlite_custom_datasets_connection() -> DatabaseContext:
+def sqlite_custom_datasets_connection() -> Callable[[str], DatabaseContext]:
     """
-    Returns the SQLITE database connection with all the custom datasets attached.
+    This fixture is used to connect to the sqlite database of the custom datasets.
+    Returns a DatabaseContext for the MySQL TPCH database.
```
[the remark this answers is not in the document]
hadia206: Don't forget to fix that^

hadia206 (approved):
Thanks John

PR #459, opened 2025-11-14
Call Expression Alias Patch

### LINKED ISSUE #458: Invalid composed SQL alias where column_name is quoted.
In some scenarios where a SQL column name is quoted in the metadata, pydough generates an invalid composed alias name where it adds a function name plus the quoted column name.

It is possible to reproduce it in the reserved words dataset using this pydough code:
```
invalid_alias_name = keywords.CALCULATE(
    max_len=MAX(partition_.integer)
).calculate_.WHERE(
    where_ == max_len
).CALCULATE(key=where_, len=length)

```

The SQL generated is:
```
ERROR WHILE OPTIMIZING QUERY:
SELECT
  _s1.".WHERE" AS key,
  _s1.LENGTH AS len
FROM (
  SELECT
    MAX("INTEGER") AS max_""INTEGER""
  FROM (
    SELECT
      "INTEGER" AS "INTEGER"
    FROM keywords."PARTITION"
  ) AS _t0
) AS _s0
INNER JOIN (
  SELECT
    LENGTH,
    ".WHERE" AS ".WHERE"
  FROM keywords.CALCULATE
) AS _s1
  ON _s1.".WHERE" = _s0.max_""INTEGER""
```

john-sanchez31:
This resolves #458


hadia206 (approved):
Good job John.
Please address comment before merging.

pydough/conversion/column_bubbler.py
```diff
@@ -76,7 +76,25 @@ def generate_cleaner_names(expr: RelationalExpression, current_name: str) -> lis
         if len(expr.inputs) == 1:
             input_expr = expr.inputs[0]
             if isinstance(input_expr, ColumnReference):
-                result.append(f"{expr.op.function_name.lower()}_{input_expr.name}")
+                input_name: str = input_expr.name
+                quoted: bool = False
```
knassre-bodo: I think the better fix here is to not do any of this special handling, but instead just remove all characters from input_name except letters/numbers/underscores. We don't have to worry about conflicts because the point of this function is that it returns _potential_ alternative names, and it only uses if them if those names are not already in use.

PR #461, opened 2025-11-28
Adding tpch custom tests for all dialects

john-sanchez31:
This PR adds the tpch custom tests for all dialects fixing the bugs along the way. List with most of the fixes:
  - Getting the microseconds mysql fixed
  - The case when expression is a CollationExpression added in the functions `identify_connection_types` and `make_hybrid_expr `. Fix for test rank_parts_per_supplier_region_by_size.
  - Faster version of the test `supplier_pct_national_qty` for mysql (was taking too long).
  - String format specifiers for mysql, postgres and snowflake added (added some helpers that change the test and result based on the dialect see: `tpch_custom_test_data_dialect_replacements`)
  - Quantile function fixed: Use `FLOOR` instead of `INTEGER` which provides more consistent result through dialects. Also use of `Decimal` instead of `float` and `Integer`. For 0 and 1 use `MIN` and `MAX` function
  - No qualification of a column in order by with Collation bug fixed test simple_cross_1 and 2
  - Faster version of the tests simple cross 6 and 7 for mysql (before was taking too long)
  - Bug fixed: AVG(Null) for postgres doesn’t work the NULL needs to be casted as INT. This fixed the test window_filter_order_10
  - Bug fixed: 1 as _ for snowflake in test `window_filter_order_10`  (this was in the sqlglot function `default_selection`)
  - INTEGER function in postgres with decimal string param fixed (CAST( ‘2.5’ AS INT) is not allowed in postgres found in test `simple_int_float_string_cast`)
  - `current` name changed because it is a reserved word in sf. Found in test `prev_next_regions`
  - For nation sf test `nation_window_aggs, wrapped the args for RELAVG in CAST AS DOUBLE, giving more precision for the final result 13.666 -> 13.66667 (which is the correct answer)
  - Sf test `simple_int_float_string_cast` fixed for `convert_integer` in sf and postgres check if the arg is literal to apply the new version which cast it as double/numeric and then truncate it
  - Sf transform bindings `convert_current_timestamp` added. This gets the current timestamp without timezone (by default sf uses timezones) like the rest of the dialects.
  - sf test `part_reduced_size` fixed, order by `dmy` added in the pydough query. This gives an consistent answer through the dialects.


pydough/conversion/hybrid_translator.py
```diff
@@ -871,10 +889,8 @@ def rewrite_quantile_call(
         max_input_arg = self.inject_expression(
             child_connection.subtree, keep_largest, create_new_calc
         )
-        max_call: HybridFunctionExpr = HybridFunctionExpr(
-            pydop.MAX, [max_input_arg], expr.typ
-        )
-
+        max_call = HybridFunctionExpr(pydop.MAX, [max_input_arg], expr.typ)
+        # breakpoint()
```
john-sanchez31: Delete this

pydough/sqlglot/override_pushdown_projections.py
```diff
@@ -94,3 +95,71 @@ def pushdown_projections(expression, schema=None, remove_unused_selections=True)
...
+        scope.clear_cache()
+
+
+# Selection to use if selection list is empty
+def default_selection(is_agg: bool) -> exp.Alias:
+    # PyDough Change: quoted=True avoiding `as _` which may be incorrect for some
+    # dialects
+    return alias(
+        exp.Max(this=exp.Literal.number(1)) if is_agg else "1", "_", quoted=True
+    )
```
john-sanchez31: Here is the change

pydough/sqlglot/override_qualify.py
```diff
@@ -293,3 +303,192 @@ def _qualify(table: exp.Table) -> None:
...
+) -> None:
+    """Disambiguate columns, ensuring each column specifies a source"""
+    # PYDOUGH CHANGE: using our custom get_scope_columns function instead of
+    # scope.columns
+    for column in get_scope_columns(scope):
+        column_table = column.table
+        column_name = column.name
+
+        if column_table and column_table in scope.sources:
+            # PYDOUGH CHANGE: using our custom get_scope_columns function
```
john-sanchez31: The important change

pydough/sqlglot/transform_bindings/postgres_transform_bindings.py
```diff
@@ -578,9 +578,21 @@ def convert_avg(
     def convert_integer(
         self, args: list[SQLGlotExpression], types: list[PyDoughType]
     ) -> SQLGlotExpression:
-        return sqlglot_expressions.Cast(
-            this=args[0], to=sqlglot_expressions.DataType.build("INTEGER")
-        )
+        assert len(args) == 1
+
+        if isinstance(args[0], sqlglot_expressions.Literal):
```
knassre-bodo: Adda comment explaining the difference here

pydough/sqlglot/override_pushdown_projections.py
```diff
@@ -94,3 +95,71 @@ def pushdown_projections(expression, schema=None, remove_unused_selections=True)
...
+def _remove_unused_selections(scope, parent_selections, schema, alias_count):
+    order = scope.expression.args.get("order")
+
+    if order:
+        # Assume columns without a qualified table are references to output columns
+        order_refs = {c.name for c in order.find_all(exp.Column) if not c.table}
+    else:
+        order_refs = set()
+
+    new_selections = []
```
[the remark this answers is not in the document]
john-sanchez31: Yes this is all from `sqlglot` except for the small changed below

hadia206 (approved):
Great work John, Thanks for the effort to fix all that!
Minor comments so approving but please make sure to address them before merging.

PR #463, opened 2025-12-19
Fixed infinite cycle in qualification caused by reference equality

juankx-bodo:
- A customer use case with multiple layers of partition/cross caused a seeming infinite loop (actually seems to be infinite recursion, but without a stack overflow for some reason)

- Removing the step where reference expression equality depends on the underlying collections' equality broke the infinite cycle/recursion since there was no longer a loop of the same collection checking for equality of underlying expressions, but one of those expressions pointed back to the original colection.

- Removing `self.collection.equals(other.collection)` from `Reference.equals()` in `reference.py` breaks the loop


knassre-bodo (approved):
Let's change the title to be "Fixed infinite cycle in qualification caused by reference equality" and add some clarifying information to the PR description:
* A customer use case with multiple layers of partition/cross caused a seeming infinite loop (actually seems to be infinite recursion, but without a stack overflow for some reason)
* Removing the step where reference expression equality depends on the underlying collections' equality broke the infinite cycle/recursion since there was no longer a loop of the same collection checking for equality of underlying expressions, but one of those expressions pointed back to the original colection.

Also, if possible, can we add a new test in the `technograph` files which emulates the failing customer query (that graph was designed to emulate them) which fails before this change but passes after?

hadia206 (approved):
Uh, what a silly mistake!
Good job finding this out.

PR #464, opened 2025-12-19
Add optional max_rows keyword argument to to_sql/to_df

### LINKED ISSUE #466: Add optional `max_rows` argument to `to_sql`/`to_df`
If the optional `max_rows` argument is provided, it must be a positive integer or None. If it is provided, PyDough will ensure that the final table outputted by `to_df` has at most `max_rows` records in it by placing a LIMIT in the SQL generated by `to_sql`.
- If the SQL query already ends in a limit `n` which is larger than `max_rows`, then `n` is replaced with `max_rows`.
- If the SQL query already ends in a limit `n` which is less than or equal to `max_rows`, then `nothing happens.

knassre-bodo:
Resolves #466.

### DIFF
--- a/tests/test_sql_refsols/nested_topk_nations_max_3_sqlite.sql
+++ b/tests/test_sql_refsols/nested_topk_nations_max_3_sqlite.sql
@@ -1,0 +1,19 @@
+WITH _s0 AS (
+  SELECT
+    n_name,
+    n_nationkey,
+    n_regionkey
+  FROM tpch.nation
+  ORDER BY
+    1
+  LIMIT 8
+)
+SELECT
+  _s0.n_nationkey AS key,
+  _s0.n_name AS name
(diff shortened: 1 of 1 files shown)


tests/test_max_rows.py
```diff
@@ -0,0 +1,129 @@
...
+            PyDoughPandasTest(
+                "result = nations.CALCULATE(key, name).TOP_K(3, by=name.ASC())",
+                "TPCH",
+                lambda: pd.DataFrame(
+                    {
+                        "key": [0, 1, 2],
+                        "name": ["ALGERIA", "ARGENTINA", "BRAZIL"],
+                    }
+                ),
+                "nations_top3_max6",
```
knassre-bodo: In this test, the query will still say LIMIT 3 because the max_rows is already larger than the existing limit, so it is irellevant.

PR #469, opened 2026-01-08
Dataframe Collection (User Collections)

### LINKED ISSUE #162: Add support for user-created collections in PyDough
The goal of this issue is to add support for a new type of collection that, instead of corresponding to a collection in the metadata, corresponds to a collection defined by the user, such as:
- Using a PyDough API to parametrically generate a table (e.g. a date range or numerical range)
- Converting a user-providing Pandas DataFrame, or a constant list of values, to a collection (either writing to the database connection, perhaps as a temporary table, or treating as a `VALUES` clause)
- Referring to a table in the database that is not part of the metadata

This issue requires designing the approach for such collections in unqualified nodes, qualification, QDAG nodes, hybrid conversion, hybrid nodes, relational conversion, relational nodes, and SQLGlot conversion, and implementing them for at least one of the choices mentioned above, as well as picking the user syntax for it. The other forms of user-created collections will be followup issues.

Example of what this could look like, with a range of integers becoming a collection:

```py
# Creates a collection `sizes` with a single property `part_size` whose values are the
# integers from 1 (inclusive) to 100 (exclusive), skipping by 5s, then for each size value,
# counts how many turquoise parts have that size.
sizes = pydough.range_collection(1, 100, 5, name='sizes', column=`part_size`)
turquoise_parts = parts.WHERE(CONTAINS(name, 'turquoise'))
result = sizes.CALCULATE(part_size).CALCULATE(part_size, n_parts=COUNT(CROSS(turquoise_parts).WHERE(size == part_size)))
```
In this example, `sizes` could be translated as a VALUES clause, or in Snowflake as a `GENERATOR` table function.

Example of what this could look like, with a pandas DataFrame becoming a collection:

```py
# Creates a collection named `months92` from the DataFrame with columns
# `m` and `d`, with 1 row for each start-of-month date in 1992. Then, for each
# month, count how many urgent orders were placed that month and the change
# from the previous month.
df = pd.DataFrame({
  "m": ["JAN", "FEB", "MAR", "APR", "MAY", "JUN", "JUL", "AUG", "SEP", "OCT", "NOV", "DEC"],
  "d": pd.date_range("1992", "1993", freq="MS", inclusive="left")
})
months = pydough.dataframe_collection(df, name="months92")
urgent_orders = orders.WHERE(order_priority == "1-URGENT")
result = months.CALCULATE(m, d).CALCULATE(n_orders=COUNT(CROSS(urgent_orders).WHERE(d == DATETIME(order_date, "start of month")))).CALCULATE(month=m, n_orders=n_orders, change_from_prev_month=n_orders-PREV(n_orders, by=d.ASC()))
```
In this example, `months` could be translated as a VALUES clause.

The spec for how this should be implemented:
- Each collection-creating API (`dataframe_collection`, `range_collection` etc.) should return a new unqualified node `UnqualifiedGeneratedCollection` which contains an instance of a `PyDoughUserGeneratedCollection` (created by the API).
- There is an abstract base class for `PyDoughUserGeneratedCollection`, with concrete implementation classes for the integer range, DataFrame, and other future implementations, each with some common APIs (e.g. how to stringify, whether it is guaranteed to be non-empty, and what column names/types it has) as the `PyDoughUserGeneratedCollection` instance is daisy-chained throughout the codebase (like how metadata is passed around between IRs)
- There should be a new QDAG node for `GeneratedCollection` (subclass of `ChildAccess`), also containing a `PyDoughUserGeneratedCollection` instance. The qualification of `UnqualifiedGeneratedCollection` to `GeneratedCollection` should be simple by just unwrapping the `PyDoughUserGeneratedCollection` instance from inside the unqualified node and wrapping it in the QDAG node.
- There should be a new hybrid operation node for `HybridGeneratedCollection` (works similar to `HybridCollectionAccess`), also containing a `PyDoughUserGeneratedCollection` instance. The conversion from `GeneratedCollection` to `HybridGeneratedCollection` should also be simple by doing a similar unwrapping & re-wrapping of the underlying `PyDoughUserGeneratedCollection`. Any hybrid utility logic that cases on `HybridCollectionAccess` should probably also case on `HybridGeneratedCollection` (e.g. `always_exists`)
- There should be a new relational node for `GeneratedTable` which works like `Scan` except it takes in the `PyDoughUserGeneratedCollection` (same unwrapping & re-wrapping)
- During SQLGlot conversion, convert the `GeneratedTable` to SQL code based on which underlying `PyDoughUserGeneratedCollection` it uses (& possibly based on the dialect -> can add more stuff to the binding infrastructure)

john-sanchez31:
Resolves #162

### DIFF
--- a/tests/test_pipeline_tpch_custom.py
+++ b/tests/test_pipeline_tpch_custom.py
@@ -226,6 +226,7 @@ from tests.test_pydough_functions.user_collections import (
     simple_dataframe_collection_1,
     simple_dataframe_collection_2,
     simple_dataframe_collection_3,
+    simple_dataframe_collection_4,
     simple_range_1,
     simple_range_2,
     simple_range_3,
@@ -4401,6 +4402,17 @@ from .testing_utilities import PyDoughPandasTest, graph_fetcher, run_e2e_error_t
                 "simple_dataframe_collection_3",
             ),
             id="simple_dataframe_collection_3",
+        ),
+        pytest.param(
+            PyDoughPandasTest(
+                simple_dataframe_collection_4,
+                "TPCH",
+                lambda: pd.DataFrame(
+                    {"user_id": [1, 2, 3, 4], "country": ["US", "CR", "US", "MX"]}
+                ),
+                "simple_dataframe_collection_4",
+            ),
+            id="simple_dataframe_collection_4",
         ),
         pytest.param(
             PyDoughPandasTest(
--- a/tests/test_pydough_functions/user_collections.py
+++ b/tests/test_pydough_functions/user_collections.py
@@ -233,15 +233,8 @@ def simple_dataframe_collection_4():
         {
             "user_id": [1, 2, 3, 4],
             "country": ["US", "CR", "US", "MX"],
-            "age": [25, 30, 22, 30],
-            "signup_date": pd.to_datetime(
-                [
-                    "2024-01-10",
-                    "2024-01-12",
-                    "2024-02-01",
-                    "2024-02-01",
-                ]
-            ),
+            "user_list": [[12, 4], [23, 5], [56, 3], [23, 6, 7]],
+            "date": [{"a": 1}, {"b": 2}, {"c": 3}, {"d": 4}],
         }
     )
 
@@ -249,6 +242,7 @@ def simple_dataframe_collection_4():
         name="users",
         dataframe=df,
         unique_column_names=["user_id"],
+        column_subset=["user_id", "country"],
     )
 
 
--- a/tests/test_plan_refsols/simple_dataframe_collection_4.txt
+++ b/tests/test_plan_refsols/simple_dataframe_collection_4.txt
@@ -1,0 +1,2 @@
+ROOT(columns=[('user_id', user_id), ('country', country)], orderings=[])
+ GENERATED_TABLE(DataframeCollection(name='users', shape=(4, 2), columns=['user_id', 'country']))
--- a/tests/test_sql_refsols/simple_dataframe_collection_4_ansi.sql
+++ b/tests/test_sql_refsols/simple_dataframe_collection_4_ansi.sql
@@ -1,0 +1,8 @@
+SELECT
+  column1 AS user_id,
+  column2 AS country
+FROM (VALUES
+  (1, 'US'),
+  (2, 'CR'),
+  (3, 'US'),
+  (4, 'MX')) AS users(user_id, country)
--- a/tests/test_sql_refsols/simple_dataframe_collection_4_mysql.sql
+++ b/tests/test_sql_refsols/simple_dataframe_collection_4_mysql.sql
@@ -1,0 +1,8 @@
+SELECT
+  users.user_id,
+  users.country
+FROM (VALUES
+  ROW(1, 'US'),
+  ROW(2, 'CR'),
+  ROW(3, 'US'),
+  ROW(4, 'MX')) AS users(user_id, country)
--- a/tests/test_sql_refsols/simple_dataframe_collection_4_postgres.sql
+++ b/tests/test_sql_refsols/simple_dataframe_collection_4_postgres.sql
@@ -1,0 +1,8 @@
+SELECT
+  users.user_id,
+  users.country
+FROM (VALUES
+  (1, 'US'),
+  (2, 'CR'),
+  (3, 'US'),
+  (4, 'MX')) AS users(user_id, country)
--- a/tests/test_sql_refsols/simple_dataframe_collection_4_snowflake.sql
+++ b/tests/test_sql_refsols/simple_dataframe_collection_4_snowflake.sql
@@ -1,0 +1,8 @@
+SELECT
+  users.user_id,
+  users.country
+FROM (VALUES
+  (1, 'US'),
+  (2, 'CR'),
+  (3, 'US'),
+  (4, 'MX')) AS users(user_id, country)
--- a/tests/test_sql_refsols/simple_dataframe_collection_4_sqlite.sql
+++ b/tests/test_sql_refsols/simple_dataframe_collection_4_sqlite.sql
@@ -1,0 +1,8 @@
+SELECT
+  users.column1 AS user_id,
+  users.column2 AS country
+FROM (VALUES
+  (1, 'US'),
+  (2, 'CR'),
+  (3, 'US'),
+  (4, 'MX')) AS users


documentation/dsl.md
```diff
@@ -1575,6 +1575,45 @@ Output:
...
 ```
 
+<!-- TOC --><a name="dataframe_collection"></a>
+### `pydough.dataframe_collection`
+
+The `dataframe_collection` creates a collection within a specified pandas dataframe. This is useful for building datasets dynamically.
+It takes in the following arguments:
+
+- `name`: The name of the dataframe collection.
+- `dataframe`: The panda dataframe containing the corresponding data
```
```suggestion
- `dataframe`: Pandas dataframe containing the corresponding data.
```

pydough/user_collections/dataframe_collection.py
```diff
@@ -0,0 +1,170 @@
+"""
+A user-defined collection of pandas Dataframes.
+Usage:
+`pydough.dataframe_collection(name, dataframe)`
+
+This module defines a collection that generated from a given pandas Dataframe.
+The user must specify the name of the collection and the
+name of the dataframe that will hold the values.
```
hadia206:
```suggestion
name of the dataframe that has the values.
```

pydough/user_collections/dataframe_collection.py
```diff
@@ -0,0 +1,170 @@
...
+
+The pandas dataframe must have at least one column with one item. The accepted
+types are: NumericType, BooleanType, DatetimeType, StringType and UnknownType.
```
hadia206: This should list the acceptable Pandas Types by the feature not the PyDough ones or you can have both.

pydough/user_collections/dataframe_collection.py
```diff
@@ -0,0 +1,170 @@
...
+Mixed types columsn are not allowed.
```
hadia206: typo.
```suggestion
Mixed types columns are not allowed.
```
Also, does Pandas allow mixed types in a column? My understanding is that will be `object` type. If that's what you're referring to let's clarify that in the docstring as well (i.e. object type is not allowed)

pydough/user_collections/dataframe_collection.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def __len__(self) -> int:
+        return len(self._dataframe)
+
+    def is_singular(self) -> bool:
+        """Returns True if the collection is guaranteed to contain at most one row."""
+        return len(self) <= 1
+
+    def always_exists(self) -> bool:
+        """Check if the range collection is always non-empty."""
```
hadia206:
```suggestion
        """Check if the dataframe collection is always non-empty."""
```

pydough/user_collections/README.md
```diff
@@ -13,6 +13,12 @@ This module defines the user collections that can be created on the fly and used
     - `end`: The ending value of the range (exclusive).
     - `step`: The step value for incrementing the range. Default is 1.
 
+### [dataframe_collection.py](dataframe_collection.py)
+
+  - `DataframeGeneratedCollection`: Class used to create a dataframe collection using the given dataframe and name.
+    - `name`: The name of the dataframe collection.
+    - `dataframe`: The pandas dataframe containing all data (rows and columns).
```
hadia206:
```suggestion
    - `dataframe`: The Pandas DataFrame containing all data (rows and columns).
```

pydough/user_collections/README.md
```diff
@@ -21,6 +27,11 @@ This module defines the user collections that can be created on the fly and used
     - `end`: The ending value of the range (exclusive).
     - `step`: The step value for incrementing the range. Default is 1.
     - Returns: An instance of `RangeGeneratedCollection`.
+  - `dataframe_collection`: Function to create a dataframe collection with the specified parameters.
+    - `name`: The name of the dataframe collection.
+    - `dataframe`: The pandas dataframe.
```
hadia206:
```suggestion
    - `dataframe`: The Pandas DataFrame.
```

pydough/user_collections/user_collection_apis.py
```diff
@@ -43,3 +46,32 @@ def range_collection(
...
 
     return UnqualifiedGeneratedCollection(range_collection)
+
+
+def dataframe_collection(
+    name: str, dataframe: pd.DataFrame
+) -> UnqualifiedGeneratedCollection:
+    """
+    Implementation of the `pydough.dataframe_collection` function, which provides
+    a way to create a collection of pandas dataframe in PyDough.
```
hadia206:
```suggestion
    a way to create a collection of Pandas DataFrame in PyDough.
```

hadia206 (commented):
Good work John. Please see my comment below.

pydough/conversion/relational_converter.py
```diff
@@ -1430,7 +1430,9 @@ def rel_translation(
                 assert context is not None, "Malformed HybridTree pattern."
                 result = self.translate_hybridroot(context)
             case HybridUserGeneratedCollection():
-                assert context is not None, "Malformed HybridTree pattern."
+                if context is None:
+                    context = TranslationOutput(EmptySingleton(), {})
```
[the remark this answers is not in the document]
knassre-bodo: I think this is a change that either I pushed or that I pair-programmed with John, which centers on the edge cases with DataFrame collection testing which somehow flew under the radar with range collections.

The difference here is that in certain situations, the root level of a child subtree can be the generated collection (instead of a normal root node like a collection access or a HybridRoot). In these situations, we need to account for the fact that there is no node above the generated collection, therefore no context for us to join from. The simple fix is to create a sentinel parent to join from (the empty singleton), which is already how we handle other similar things.

In terms of a comment, something like:
`# Account for cases where there is no ancestor preceding the generated collection`

documentation/dsl.md
```diff
@@ -1575,6 +1575,45 @@ Output:
...
+
+Supported Signatures:
+- `dataframe_collection(dataframe, name)`: generates collection with the given datafram and name.
+
+#### Example
```
knassre-bodo: Should we perhaps also have a more involved example that does more things with the collection?

pydough/user_collections/dataframe_collection.py
```diff
@@ -0,0 +1,170 @@
...
+        return self._dataframe
+
+    @property
+    def column_names_and_types(self) -> list[tuple[str, PyDoughType]]:
+        assert len(self.columns) == len(self.types)
+        return [(self.columns[i], self.types[i]) for i in range(len(self.columns))]
+
+    @property
+    def unique_column_names(self) -> list[str]:
+        return list(dict.fromkeys(self.columns))
```
knassre-bodo: Aww heck I knew I forgot something in the spec: we need the option to specify subsets of the columns that are unique (just like we do in metadata). It can be an optional argument in `pydough.dataframe_collection` which has type `list[str | list[str]]` where the strings are the column names. For instance, in your `dsl.md` exmaple, it would be `["color", "idx"]` since both columns are unique.

We should then have tests on the behavior of window function / correlation tests with DataFrames that use this argument.

tests/test_pipeline_tpch_custom.py
```diff
@@ -3241,6 +3262,61 @@
             ),
             id="simple_range_5",
         ),
+        pytest.param(
+            PyDoughPandasTest(
+                "r = pydough.range_collection('tbl', 'v', 0, 500, 13).CALCULATE(first_digit=INTEGER(STRING(v)[:1]))\n"
+                "result = r.PARTITION(name='digits', by=first_digit).CALCULATE(first_digit, n=COUNT(tbl))",
+                "TPCH",
```
knassre-bodo: These new tests cover some categories of edge cases John discovered while testing DataFrames which we realized hadn't been covered/handled yet for simple range (hence all the weird HybridTree changes).

pydough/user_collections/dataframe_collection.py
```diff
@@ -0,0 +1,170 @@
...
+
+            case _ if pa.types.is_string(field_type) or pa.types.is_large_string(
+                field_type
+            ):
+                return StringType()
+
+            case _ if pa.types.is_binary(field_type) or pa.types.is_large_binary(
+                field_type
+            ):
+                return StringType()
```
knassre-bodo: Shouldn't this be rejected for now?

pydough/user_collections/dataframe_collection.py
```diff
@@ -0,0 +1,170 @@
...
+
+            case _ if pa.types.is_struct(field_type):
+                raise ValueError("Structs are not supported for dataframe collections")
+
+            case _ if pa.types.is_dictionary(field_type):
+                raise ValueError(
+                    "Dictionaries are not supported for dataframe collections"
+                )
+            case _:
+                return UnknownType()
```
knassre-bodo: Are there cases where this happens that are valid? If so what are there? (should be documented in comments). Otherwise, we should explicitly ban things with clear error messages.

pydough/user_collections/dataframe_collection.py
```diff
@@ -0,0 +1,201 @@
...
+        return self._dataframe
+
+    @property
+    def column_names_and_types(self) -> list[tuple[str, PyDoughType]]:
+        assert len(self.columns) == len(self.types)
+        return list(zip(self.columns, self.types))
+
+    @property
+    def unique_column_names(self) -> list[str]:
+        return list(dict.fromkeys(self.columns))
```
knassre-bodo: Reminder to address the comment I made about about this; this is a serious thing that needs to be added & tested.

pydough/user_collections/dataframe_collection.py
```diff
@@ -0,0 +1,201 @@
...
+            f"shape={self.dataframe.shape}, "
+            f"columns={list(self.dataframe.columns)!r}"
+            f")"
+        )
+
+    def equals(self, other) -> bool:
+        return (
+            isinstance(other, DataframeGeneratedCollection)
+            and self.name == other.name
+            and self.columns == other.columns
```
hadia206: Do we need these? If we already check `self.dataframe.equals` in the line below it?

documentation/dsl.md
```diff
@@ -1575,6 +1575,57 @@ Output:
...
+- `name`: The name of the dataframe collection.
+- `dataframe`: Pandas DataFrame containing the corresponding data.
+
+#### Datatypes
+`dataframe_collection` doesn't support mix datatypes in one column.
+
+The supported PyDough types for `dataframe_collection` are:
+- `NumericType`: includes float, integer, infinity, Nan.
+- `BooleanType`: includes classic true and false.
+- `StringType`: includes all aphanumeric caracters.
```
hadia206: Removed and fixed the typo
```suggestion
- `StringType`: alphanumeric characters.
```

documentation/dsl.md
```diff
@@ -1575,6 +1575,57 @@ Output:
...
+- `Datetype`: includes date and datetime.
```
hadia206: Timestamp? NaT?

```suggestion
- `Datetype`: date and datetime.
```

pydough/user_collections/dataframe_collection.py
```diff
@@ -0,0 +1,201 @@
...
+        elif pa_types.is_list(field_type) or pa.types.is_large_list(field_type):
+            raise ValueError("Arrays are not supported for dataframe collections")
+
+        elif pa_types.is_struct(field_type):
+            raise ValueError("Structs are not supported for dataframe collections")
+
+        elif pa_types.is_dictionary(field_type):
+            raise ValueError("Dictionaries are not supported for dataframe collections")
+        else:
+            return UnknownType()
```
hadia206: Let's clarify what are these if any are still missing. 
See my comment in the dsl.md as well.

pydough/user_collections/dataframe_collection.py
hadia206: Reminder to do this^

pydough/user_collections/dataframe_collection.py
```diff
@@ -0,0 +1,170 @@
...
+                key=lambda t: t.__name__,
+            )
+            if len(types_list) > 1:
+                raise TypeError(
+                    f"Mixed types in column '{col}': {types_list}. All values in a column must be of the same type."
+                )
+            elif len(dataframe[col]) == 0:
+                raise TypeError(
+                    f"Column '{col}' is empty. All columns must have at least one value."
+                )
```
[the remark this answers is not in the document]
hadia206: My understanding from the code is that there's no rows, the code will raise an error. 
@knassre-bodo your test suggestion makes me think that we should allow generating a collection from a DataFrame with no data?

pydough/conversion/agg_removal.py
```diff
@@ -237,7 +237,7 @@ def aggregation_uniqueness_helper(
     match node:
         # Scans are unchanged, and their uniqueness is based on the unique sets
         # of the underlying table.
-        case Scan():
+        case Scan() | GeneratedTable():
```
hadia206: Why moved it here?

pydough/user_collections/user_collection_apis.py
```diff
@@ -49,7 +49,7 @@ def range_collection(
 
 
 def dataframe_collection(
-    name: str, dataframe: pd.DataFrame
+    name: str, dataframe: pd.DataFrame, unique_column_names: list[str | list[str]]
```
hadia206: Should this be optional? Because now it means the user has to specify unique column names which is not intuitive.
Also, I have another request (but that can be a followup if it's too much)
which is having another optional argument specifying some column names to be able create a collection from these columns only.

pydough/user_collections/user_collection_apis.py
```diff
@@ -66,12 +68,26 @@ def dataframe_collection(
         raise TypeError(f"Expected 'name' to be a string, got {type(name).__name__}")
     if not isinstance(dataframe, pd.DataFrame):
         raise TypeError(
-            f"Expected 'dataframe' to be a pandas DataFrame, got {type(dataframe).__name__}"
+            f"Expected 'dataframe' to be a Pandas DataFrame, got {type(dataframe).__name__}"
+        )
+    if not isinstance(unique_column_names, list):
+        raise TypeError(
+            f"Expected 'unique_column_names' to be a list of string, got {type(unique_column_names).__name__}"
```
hadia206: it could also be a list of list of strings

documentation/dsl.md
```diff
@@ -1585,6 +1585,15 @@ It takes in the following arguments:
 
 - `name`: The name of the dataframe collection.
 - `dataframe`: Pandas DataFrame containing the corresponding data.
+- `unique_column_names`: List of strings or list of strings
```
hadia206: This should be `list of strings or list of list of strings`

tests/test_pydough_functions/user_collections.py
```diff
@@ -149,3 +152,776 @@ def user_range_collection_6():
...
+
+    classes_collection = pydough.dataframe_collection(
+        "classes", class_df, ["key", "class_name"]
+    )
+
+    other_classes_same_language = (
+        CROSS(classes_collection.CALCULATE(language, key)).WHERE(
+            (language == original_language) & (key != original_key)
+        )
+        # .CALCULATE(metric=(key != original_key))
```
knassre-bodo:
```suggestion
```

tests/test_pipeline_tpch_custom.py
```diff
@@ -4743,6 +5515,72 @@ def test_pipeline_e2e_tpch_custom(
...
+        ),
+        pytest.param(
+            dataframe_collection_bad_5,
+            None,
+            re.escape(
+                "Arrays in column 'col1', are not supported for dataframe collections"
+            ),
+            id="dataframe_collection_bad_5",
+        ),
+        pytest.param(
```
knassre-bodo: Note: we should have a regular test with a column that is a bad type but is NOT included in the filtered columns, which means it does NOT have an error (e.g. column `x` is an array type, but not included in the filtered columns) since the type checking/inference only matters for the columns we are using.

pydough/user_collections/dataframe_collection.py
```diff
@@ -0,0 +1,291 @@
...
+                f"Fixed-size arrays / tuples in column '{field_name}', are not supported for dataframe collections"
+            )
+
+        else:
+            raise ValueError(
+                f"Unknown type in column '{field_name}' for dataframe collections"
+            )
+
+    @staticmethod
+    def valid_unique_column_names(unique_columns_name: list[str | list[str]]) -> bool:
```
knassre-bodo: We already have something like this in `error_utils.py` used for metadata, called `unique_properties_predicate`. Calling `unique_properties_predicate.verify(unique_columns_names, "unique_columns_names")` verifies that `unique_columns_names` is a non-empty list of objects that are either strings or non-empty lists of strings, and if not raises an exception with the name `"unique_columns_names"` inside the message

PR #472, opened 2026-01-16
Fix logging issues

juankx-bodo:


pydough/logger/logger.py
```diff
@@ -7,9 +7,21 @@
 import sys
 
 
+def get_level_source_logger(logger: logging.Logger) -> logging.Logger:
+    """
+    Returns the logger from which the effective level is inherited.
```
hadia206: missing args/returns for docstring format

pydough/logger/logger.py
```diff
@@ -18,27 +30,30 @@ def get_logger(
     The default handler redirects to standard output. Additional handlers can be sent as a list.
 
     Args:
-        `name` : Logger name, usually the `__name__` from the calling module.
-        `default_level` : Default logging level if not set externally.
+        `name` : The logger name you want to get or create (in case it does not exists)
+        `default_level` : Logging level. PYDOUGH_LOG_LEVEL value will be used if not set and not ancestor level has been already set.
```
hadia206:
```suggestion
        `default_level` : Logging level. PYDOUGH_LOG_LEVEL value will be used if not set and ancestor level has not been already set.
```

hadia206 (commented):
The code looks correct to me.
I'm not familiar with logging so will let Kian or John give approval.

pydough/logger/logger.py
```diff
@@ -18,27 +30,30 @@ def get_logger(
...
     Returns:
         `logging.Logger` : Configured logger instance.
     """
     logger: logging.Logger = logging.getLogger(name)
-    level_env: str | None = os.getenv("PYDOUGH_LOG_LEVEL")
-    level: int
+    level: int = logging.INFO
+    if default_level is None:
+        # Get log level from PYDOUGH_LOG_LEVEL
+        level_env = os.getenv("PYDOUGH_LOG_LEVEL")
```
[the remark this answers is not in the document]
knassre-bodo: Yeah I'm pretty sure these are always strings when obtained from `os.getenv`?

PR #473, opened 2026-01-23
PyDough performance optimizations with multiple COUNT on different filters

### LINKED ISSUE #477: Combine child subtrees being counted with different filters using SUM(IFF(...))
Consider a PyDough query such as this:
```
nations.CALCULATE(
  name,
  c1 = COUNT(customers.WHERE(market_segment == "BUILDING")),
  c2 = COUNT(customers.WHERE((market_segment == "BUILDING") & MONOTONIC(500, account_balance, 600))),
)
```

Currently, this will be implemented by taking `nations` and joining it onto two filtered/aggregated copies of `customers`, one filtered with just the market_segment and one on the market_segment & account_balance. However, there is a more performant way to write this where only 1 copy of `customers` is scanned/filtered/aggregated:
```
child = customers.WHERE(market_segment == "BUILDING").CALCULATE(expr=MONOTONIC(500, account_balance, 600))
nations.CALCULATE(
  name,
  c1 = COUNT(child),
  c2 = COUNT(SUM(IFF(child.expr, 1, 0))),
)
```

The goal of this optimization is to add a new pass to the hybrid tree that automatically transforms the datastructures to make it so if the PyDough query is written in the first way, it is transformed as if it were written the second way and thus improve performance. In concrete terms, the algorithm is:
- Recursively traverse the entire hybrid tree and invoke the rest of the procedure on each level with multiple children
- Identify any children that are only used in a `COUNT` call that is not an `ONLY_MATCH` child; these children are "mergeable"
- For each child, identify the set of all filters (e.g. the conjunction) in the bottom-most layer of the child (only looking at the ones after any limits or window functions)
- Identify which of the children subtrees are **identical** except for these filters
- Find each pair of childA and childB where childA is "mergeable", the child subtree is identical to child B, and the set of filters from childA is a **superset** of the filters in childB; if multiple such childB exist for childA, only focus on the childB with the smallest set of filters
- Merge childA into childB by creating a new expression `SUM(IIF(cond, 1, 0))` in childB where `cond` are all of the excess filters in childA that are not in childB; replace all references to the count of childA with this new SUM call in childB

This optimization must be done after hybrid decorrelation to remove a lot of complicated unnecessary edge cases. If it is done before the dead child removal pass, then any redundant childA subtrees will automatically be cleaned up.

knassre-bodo:
Resolves #477.


pydough/conversion/hybrid_tree.py
```diff
@@ -976,7 +976,7 @@ def remove_dead_children(self, must_remove: set[int]) -> dict[int, int]:
                 continue
             if (
                 self.children[child_idx].connection_type.is_semi
-                and not self.children[child_idx].subtree.always_exists()
+                and not self.children[child_idx].get_always_exists()
```
knassre-bodo: This change is crucial to avoid a subtle bug with pruning dead children

tests/test_sql_refsols/defog_restaurants_gen8_sqlite.sql
```diff
@@ -12,7 +12,7 @@ WITH _s1 AS (
 ), _s7 AS (
   SELECT
     _s3.region,
-    SUM(IIF(NOT restaurant.rating IS NULL, 1, 0)) AS sum_expr,
+    SUM(NOT restaurant.rating IS NULL) AS sum_expr,
```
knassre-bodo: A cosmetic improvement as a result of the changes to the relational optimizer

pydough/conversion/hybrid_filter_merger.py
```diff
@@ -0,0 +1,560 @@
...
+)
+from .hybrid_tree import HybridTree
+
+if TYPE_CHECKING:
+    from .hybrid_translator import HybridTranslator
+
+
+class HybridFilterMerger:
+    """
+    TODO
```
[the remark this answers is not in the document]
knassre-bodo: Oop good catch

pydough/conversion/hybrid_filter_merger.py
```diff
@@ -0,0 +1,564 @@
...
+
+            # Create a secondary mapping to indicate pools of children that were
+            # not merged by the dag because there was no child with a filter
+            # subset relationship, but are still isomorphic to one another.
+            # These are stored in the form of a pool of isomorphic children,
+            # where one member of the pool is the key and the rest are the
+            # value.
+            secondary_merges: dict[int, set[int]] = self.make_secondary_merges(
+                mergeable_children, child_isomorphisms, filter_dag
+            )
```
hadia206: nit: could we move this part to be closer to when it's used? It's confusing to have this step between ` self.make_filter_dag` and building the dictionary to do merging

pydough/conversion/hybrid_filter_merger.py
```diff
@@ -0,0 +1,564 @@
...
+        # replacement map to point from the old COUNT(*) reference in the source
+        # subtree to the new SUM expression reference in the target subtree.
+        agg_name: str = self.translator.gen_agg_name(tree.children[target_idx])
+        tree.children[target_idx].aggs[agg_name] = sum_expr
+        agg_ref: HybridExpr = HybridChildRefExpr(agg_name, target_idx, NumericType())
+        old_agg_ref = HybridChildRefExpr(
+            next(
+                name
+                for name, expr in tree.children[source_idx].aggs.items()
+                if repr(expr) == "COUNT()"
```
hadia206: why exact name match? why not use operator check like`expr.op == pydop.COUNT and len(expr.args) == 0`?

pydough/conversion/hybrid_filter_merger.py
```diff
@@ -0,0 +1,564 @@
...
+        """
+        Identify the subset of child indices from a hybrid tree where the child
+        is an aggregation where the only aggregate is a single COUNT(*).
+
+        Args:
+            `tree`: The tree whose children we are checking.
+
+        Returns:
+            A set of the indices of the children that are only used by a COUNT
+            aggregation that is not ONLY_MATCH.
```
hadia206: Is that true? The code has includes `AGGREGATION_ONLY_MATCH`

pydough/conversion/hybrid_translator.py
```diff
@@ -1725,10 +1739,11 @@ def convert_qdag_to_hybrid(self, node: PyDoughCollectionQDAG) -> HybridTree:
...
+        # 6. Run the filter-merging procedure, then re-run ejecting aggregate
+        # inputs to clean up any new aggregates created by filter merging.
+        self.run_filter_merging(hybrid)
+        self.eject_aggregate_inputs(hybrid)
+        # 8. Run any final rewrites, such as turning MEDIAN into an average
         # of the 1-2 median rows, that must happen after de-correlation.
         self.run_rewrites(hybrid)
-        # 7. Remove any dead children in the hybrid tree that are no longer
-        # being used.
-        hybrid.remove_dead_children(set())
```
hadia206: This is removed because it's not done in the  run_filter_merge step, right?
We don't need to check again after run_rewrites?

tests/test_pipeline_tpch_custom.py
```diff
@@ -2690,6 +2690,632 @@
             ),
             id="quarter_function_test",
         ),
+        pytest.param(
```
hadia206: Recommended tests:
- make_condition with empty set (no extra filters)
- Limit test
- Window test
- `COUNT(column)`

pydough/conversion/hybrid_filter_merger.py
```diff
@@ -0,0 +1,564 @@
...
+        # Now go back through the target subtree, find any existing filters
+        # after any window/limit, and make them a disjunction of the existing
+        # filters and the disjunction of source conjunctions.
+        for operation in reversed(tree.children[target_idx].subtree.pipeline):
+            if isinstance(operation, HybridFilter):
+                if operation.condition.contains_window_functions():
+                    break
+                operation.condition = HybridFunctionExpr(
+                    pydop.BOR,
+                    [operation.condition, new_disjunction],
```
hadia206: That works if the `operation.condition` is already a disjunction right?

pydough/conversion/hybrid_filter_merger.py
```diff
@@ -0,0 +1,564 @@
...
+        # Run the main procedure on subtrees with multiple children.
+        if len(tree.children) > 1:
+            # Identify which children are only used by a COUNT aggregation that
+            # is not ONLY_MATCH.
+            mergeable_children: set[int] = self.identify_mergeable_children(tree)
+
+            # Extract the set of filters in the bottom level of each child tree,
+            # only considering filters after a critical point (limits, windows).
+            child_filters: list[set[HybridExpr]] = [
+                self.get_final_filters(child.subtree) for child in tree.children
```
hadia206: What happens if `.get_final_filters` returned an empty set?
Wouldn't that mean it's a subset of all other sets?

pydough/conversion/hybrid_filter_merger.py
```diff
@@ -0,0 +1,564 @@
...
+            strict superset of filters compared to that child, and they are
+            isomorphic after stripping away filters.
+        """
+        # Build up the initial DAG as all-None, then fill in as connections
+        # are formed.
+        dag: list[int | None] = [None for _ in range(len(child_filters))]
+
+        # Build initial edges from each mergeable child to another isomorphic
+        # child that is a subset of its filter list.
+        for idx in mergeable_children:
```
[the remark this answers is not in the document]
knassre-bodo: Good point!

pydough/conversion/hybrid_filter_merger.py
```diff
@@ -0,0 +1,564 @@
+"""
+Logic to merge multiple subtrees in the hybrid tree into one if they are the
+same except one of them has more filters than the other and is only used in
+a COUNT aggregation, meaning the filter can be implemented by doing a SUM on
+the less-filtered subtree where the SUM argument is the additional filters.
```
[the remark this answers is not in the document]
knassre-bodo: Will do.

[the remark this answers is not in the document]
knassre-bodo: What I'm adding:
```
Suppose the current tree has five children `A`, `B`, `C`, `D` and `E` which
are identical except that their lowest levels have the following pipelines, and
the only reference from the current tree to these children is a `COUNT(*)`
aggregation on each of them:
- `A`: `Collection[table1] -> FILTER(u) -> FILTER(v & w)`
- `B`: `Collection[table1] -> FILTER(v)`
- `C`: `Collection[table1] -> FILTER(x) -> LIMIT(1000) -> FILTER(u) -> FILTER(v)`
- `D`: `Collection[table1] -> FILTER(x) -> LIMIT(1000) -> FILTER(v) -> FILTER(y)`
- `E`: `Collection[table1] -> FILTER(x) -> LIMIT(1000) -> FILTER(v & w) -> FILTER(z)`

The protocol will detect that `A` can be merged onto `B` since they are
identical except that `A` has extra filters `u` and `w`, so the a subset will
occur on A->B.

The protocol will then detect that `C`, `D` and `E` all have the same common
prefix with different filters after the LIMIT clause (u & v, v & y, v & w & z),
so it will attempt to merge all three of them. For the sake of arbitrariness,
assume that it does (D, E) -> C as a a non-subset merge.

When doing the subset merge of A->B, child A will be deleted and the reference
to `COUNT(*)` on A will be replaced with a reference to `SUM(IFF(u & w, 1, 0))`
on B, since the SUM will count how many of the rows where `v` is true also
satisfy the conditions `u` and `w`, which is exactly the same as the original
COUNT on A which had `u`, `v` and `w`.

When doing the subset merge of (D, E) -> C, child D and E will be deleted, and
child C will be transformed into the following subtree (call this `C'`):

`Collection[table1] -> FILTER(x) -> LIMIT(1000) -> FILTER(u | (v & y) | (v & w & z)) -> FILTER(v | (v & y) | (v & w & z))`

This will ensure that `C'` contains all the rows that were in `C`, `D` or `E`.
Then to get the correct counts for `C`, `D` and `E`, the original `COUNT(*)`
references on each of them will be transformed into the following aggregations
on `C'` (note that `v` appeared in all 3 children, so it is a given than `v` is
true in `C'`):
- `C`: `SUM(IFF(u, 1, 0))`, since `C` had `u & v` and `C'` covers `v`.
- `D`: `SUM(IFF(y, 1, 0))`, since `C` had `v & y` and `C'` covers `v`.
- `E`: `SUM(IFF(w & z, 1, 0))`, since `C` had `v & w & z` and `C'` covers `v`.

This does result in some redundant filters in `C'` which can be optimized away
by later transformations at the relational/SQLGlot level into the equivalent of
the following:

`Collection[table1] -> FILTER(x) -> LIMIT(1000) -> FILTER(v & (u | y | (w & z))`
```

PR #480, opened 2026-01-28
Optimize SEMI joins to INNER joins when HAS is used on singular relationships

### LINKED ISSUE #476: Ensure that SEMI joins on singular data are promoted to INNER joins
Consider a PyDough query such as this:
```
customers.WHERE(HAS(nation.WHERE(region.name == "EUROPE")))
```

Currently, this will create a SEMI join between the `customers` table and the result of `nation.WHERE(region.name == "EUROPE"))`. However, a SEMI join here is overkill since there will be at most 1 matching record of nations/regions for each customer. In performance terms, this will result in an unnecessary additional layer of aggregation on nations/regions before joining it onto customers. Instead, we can promote this SEMI join directly to an INNER join when converting from the hybrid tree to the relational tree.

The general rule: if joining the current context to the child context and the join type is `SEMI`, replace with `INNER` if the child context is singular with regards to the current context.

hadia206:
-  Optimizes using `INNER JOIN` instead of `SEMI JOIN` when `HAS` is used on a singular sub-collection, the existence check is redundant since singular relationships always have exactly one match.
- Qualification Bug Fix for issue where queries failed when the graph name matched a collection name

closes #476


tests/test_pipeline_tpch_custom.py
```diff
@@ -2739,6 +2739,123 @@
...
+                lambda: pd.DataFrame(
+                    {
+                        "n": [115066],
+                    }
+                ),
+                "redundant_has_on_plural_lineitems",
+            ),
+            id="redundant_has_on_plural_lineitems",
+        ),
+        # HASNOT on singular relationship - should optimize to ANTI join or similar
```
knassre-bodo: No optimization here, it just stays as ANTI the entire time

PR #488, opened 2026-02-10
Fix bug with HASNOT(x)==1

knassre-bodo:
If the expression `HASNOT(x) == 1` is written in PyDough, this will be translated as `(COUNT(x) == 0) == 1`. The count becomes a left join count reference `count_expr` and the condition expression becomes `(COALESCE(count_expr, 0) == 0) == 1`, which the sqlglot simplifier will rewrite as `(count_expr == 0 OR count_expr IS NULL) == 1`. This is problematic because in some dialects, such as SQL, doing comparisons in the form `(bool_expr) == 1` is invalid.

The workaround fix for this is to modify PyDough's relational simplifier to detect cases in the form `(bool_expr) == 1` and change them to `bool_expr`, so for the example above it would replace `(COALESCE(count_expr, 0) == 0) == 1` with `COALESCE(count_expr, 0) == 0`, which becomes the non-problematic expression `count_expr == 0 OR count_expr IS NULL`.

### DIFF
--- a/tests/test_pipeline_tpch_custom.py
+++ b/tests/test_pipeline_tpch_custom.py
@@ -2983,6 +2983,17 @@ from .testing_utilities import PyDoughPandasTest, graph_fetcher, run_e2e_error_t
         ),
         pytest.param(
             PyDoughPandasTest(
+                bad_child_reuse_5,
+                "TPCH",
+                lambda: pd.DataFrame(
+                    {"key": [2487, 43044, 69321, 76146], "n_orders": [0] * 4}
+                ),
+                "bad_child_reuse_5",
+            ),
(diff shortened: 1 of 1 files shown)


tests/test_pipeline_tpch_custom.py
```diff
@@ -2981,6 +2981,24 @@
             ),
             id="bad_child_reuse_4",
         ),
+        pytest.param(
```
john-sanchez31: NIT: Is there a reason for these tests to be bewteen `bad_child_reuse` tests? Can we move these tests before `bad_child_reuse_1` or after `bad_child_reuse_5`?

hadia206 (approved):
Looks good to me

PR #490, opened 2026-02-11
Add BodoSQL support

### LINKED ISSUE #228: Add support for BodoSQL context in PyDough
Goals of this issue:
- Add the ability to create a PyDough database context that is attatched to a BodoSQL context (either local, Snowflake, or Iceberg catalog). E.g. `pydough.active_session.load_database_context("BodoSQL", context=bc)`
   - This may require creating a DB API 2.0 compliant wrapper around the BodoSQL context, which can simply raise an exception for everything except for `cursor` which fails for eveyrthing except `execute` and `fetchall` for the time being. 
   - Alternatively, a refactor can be done to the DatabaseConnection API to work around this by allowing either a DB API 2.0 connection or the BodoSQL context, and all methods of the context switch on whether it is the connection or the BodoSQL context.
- Add any necessary changes to SQLGlot transformation to work with a BodoSQL context (the Snowflake dialect should work?)
- Add the ability to use the Pydough BodoSQL context to execute a query in BodoSQL and return the answer as a pandas DataFrame.
- Add any necessary documentation
- Add e2e testing of BodoSQL with local catalog & filesystem-based Iceberg catalog (will need PyDough metadata).
- Add non-CI testing of BodoSQL with a Snowflake catalog & non-filesystem-based Iceberg catalog (requires using secrets to configure the catalogs). If it is possible to have a CI that accesses these secrets without exposing them to users, create a special workflow that runs these tests that triggers nightly or with special permissions. Otherwise, mark these tests so they are always skipped unless the secrets are present, which can only be done in local development.

knassre-bodo:
Resolves #228. Testing includes a new workflow that is triggered by `[RUN BODOSQL]`, which runs a bevy of queries on 3 custom datasets:
- `COLORSHOP` uses a BodoSQL context with tables defined in local pandas DataFrames
- `CORPUS` uses a BodoSQL context with an Iceberg file system catalog of several tables that are written into Iceberg format from loaded compressed CSV files, where the writing process has a first pass with PyIceberg that adds partitioning and a second pass that adds puffin files.
- `FASHIONSTORE` is the same idea as `CORPUS`, just with a different data set with more tables in it.

It also runs all of the TPCH/Defog queries (standard & custom) on BodoSQL.

### DIFF
--- a/tests/test_pipeline_bodosql.py
+++ b/tests/test_pipeline_bodosql.py
@@ -12,8 +12,6 @@ import pandas as pd
 import pytest
 import datetime
 import logging
-import subprocess
-import sys
 import bodo
 from io import StringIO
 from bodo.spawn.utils import set_global_config
@@ -29,10 +27,6 @@ from functools import cache
 import pyarrow as pa
 from pyarrow import Table as PyArrowTable
 import shutil
-from pyiceberg.schema import Schema as IcebergSchema
-from pyiceberg.partitioning import PartitionSpec, PartitionField
-
-from tests.conftest import sf_conn_db_context
 from tests.testing_utilities import graph_fetcher, temp_env_override
 import numpy as np
 import numpy.typing as npt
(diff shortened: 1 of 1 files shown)


tests/test_bodosql.py
```diff
@@ -0,0 +1,1225 @@
...
+                "result = CORPUS.CALCULATE(n=COUNT(shakespeare.WHERE((play == 'macbeth') & (act == 5))))",
+                "CORPUS",
+                lambda: pd.DataFrame({"n": [62]}),
+                "corpus_q03",
+            ),
+            id="corpus_q03",
+        ),
+        pytest.param(
+            # How many words appear in the Shakespeare dialogue sample of act
+            # 5 of Macbeth?
```
hadia206: Wrong description

tests/test_metadata/bodosql_graphs.json
```diff
@@ -0,0 +1,434 @@
...
+          {
+            "name": "name",
+            "type": "table column",
+            "column name": "supname",
+            "data type": "string",
+            "description": "The company name",
+            "sample values": ["Pallette Emporium", "Rainbow Inc.", "Hue Depot"]
+          }
+        ],
+        "description": "The customers in the database",
```
hadia206: companies?

demos/notebooks/bodosql_color_tables.py
```diff
@@ -0,0 +1,115 @@
+import datetime
```
hadia206: nit: add higl-level description of the purpose of this file

hadia206 (approved):
I have minor test comment but otherwise looks good.
Thanks Kian!

tests/test_pipeline_bodosql.py
```diff
@@ -0,0 +1,1225 @@
+"""
+Tests for the PyDough workflow integrated with BodoSQL
```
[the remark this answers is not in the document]
knassre-bodo: Some of these were more indirect through similar tests, but yeah I can add direct ones. HASNOT actually hasn't been done yet, good catch I think I checked it off on accident. Puffin/NDV is being checked by examining the plans via `bodosql_plan_log_messages` to look for specific lines in the plan indicating NDV values (which can only be there if Puffin is beign done correctly), but it looks like the test I had checking those values got hit by some kind of conflict that reverted them, so I'll add them back.

As for some other topics... I may spend some time to try and get defog + TPCH on BodoSQL so we can check all of those functionalities with BodoSQL.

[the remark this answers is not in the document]
knassre-bodo: Update: all the defog/tpch tests (standard benchmark & custom queries) have been added with BodoSQL, so anything covered by ANY of those tests is now covered by BodoSQL. The main thing that isn't working is STARTSWITH with Iceberg tables because of some Bodo/BodoSQL bug they are aware of.

PR #498, opened 2026-03-04
Add support for Trino dialect and database context in PyDough

knassre-bodo:
Adding support for the Trino dialect in Pydough, including testing with a docker container that has four Trino connections to other databases: MySQL, Postgres, Mongo, and Cassandra.


pydough/unqualified/unqualified_node.py
```diff
@@ -243,7 +243,7 @@ def __pos__(self):
         return self
 
     def __neg__(self):
-        return 0 - self
+        return -1 * self
```
knassre-bodo, line 246: This change had a ripple effect to elide some obscure SQLGlot bugs that I did not have the bandwidth to investigate as part of this PR.

pydough/sqlglot/override_simplify.py
```diff
@@ -385,3 +388,88 @@ def rewrite_nullif_coalesce(expr: exp.Expression) -> exp.Expression:
         return exp.Nullif(this=lhs.args.get("this"), expression=rhs, copy=False)
     else:
         return expr
+
+
+def rewrite_length_literal(expr: exp.Expression) -> exp.Expression:
```
knassre-bodo, line 398: These various changes allowed us to avoid some very nasty and borderline unreadable code generated by the Trino bindings.

tests/test_pipeline_mysql.py
```diff
@@ -410,92 +402,6 @@ def test_pipeline_e2e_mysql_custom_functions(
     )
 
 
-@pytest.mark.mysql
-@pytest.mark.execute
-def test_pipeline_e2e_mysql_tpch_simple_week(
```
knassre-bodo, line 415: Combined all of these into one when adding the Trino version.

tests/test_pipeline_tpch.py
```diff
@@ -67,4 +67,6 @@ def test_pipeline_e2e_tpch(
 
     db_context, graph = all_dialects_tpch_db_context
 
-    tpch_pipeline_test_data.run_e2e_test(lambda _: graph, db_context, coerce_types=True)
+    tpch_pipeline_test_data.run_e2e_test(
+        lambda _: graph, db_context, coerce_types=True, atol=5e-3
```
knassre-bodo, line 71: This is because when Trino takes the average of decimal types, it keeps the existing scale, so all of the `AVG` calls were only accurate up to 2 digits.

tests/test_pipeline_custom_datasets.py
```diff
@@ -233,6 +233,7 @@ def custom_datasets_test_data(request) -> PyDoughPandasTest:
     return request.param
 
 
+@pytest.mark.custom_datasets
```
knassre-bodo, line 236: This made my job easier in testing by allowing me to deselect these.

tests/test_plan_refsols/sign.txt
```diff
@@ -1,2 +1,2 @@
-ROOT(columns=[('high', sbDpHigh), ('high_neg', -1:numeric * sbDpHigh), ('high_zero', 0:numeric), ('sign_high', SIGN(sbDpHigh)), ('sign_high_neg', SIGN((-1:numeric * sbDpHigh))), ('sign_high_zero', SIGN(0:numeric))], orderings=[(sbDpHigh):asc_first], limit=5:numeric)
- SCAN(table=main.sbDailyPrice, columns={'sbDpHigh': sbDpHigh})
+ROOT(columns=[('ticker_id', sbDpTickerId), ('exp', sbDpHigh - 185:numeric), ('sign_exp', SIGN((sbDpHigh - 185:numeric))), ('sign_neg_exp_a', SIGN((-1:numeric * (sbDpHigh - 185:numeric)))), ('sign_neg_exp_b', SIGN((-1.0:numeric * (sbDpHigh - 185:numeric)))), ('sign_pos', SIGN(13:numeric)), ('sign_neg', SIGN(-0.5:numeric)), ('sign_zero', SIGN(0:numeric)), ('sign_exp_zero', SIGN(0:numeric)), ('sign_abs_exp', SIGN(ABS((sbDpHigh - 185:numeric)))), ('sign_neg_abs_exp', SIGN((-1:numeric * ABS((sbDpHigh - 185:numeric)))))], orderings=[(sbDpDate):asc_first, (sbDpTickerId):asc_first], limit=5:numeric)
+ SCAN(table=main.sbDailyPrice, columns={'sbDpDate': sbDpDate, 'sbDpHigh': sbDpHigh, 'sbDpTickerId': sbDpTickerId})
```
knassre-bodo, line 2: Drastically reworked this test to show more interesting information, contain more useful edge cases, and also catch some bugs I found along the way.

.github/workflows/trino_testing.yml
```diff
@@ -0,0 +1,146 @@
...
+          # Set environment variables for MySQL container
+          MYSQL_ROOT_PASSWORD: ${{ secrets.MYSQL_PASSWORD }}
+          MYSQL_DATABASE: tpch
+        ports:
+          - 3306:3306
+      mongo:
+        image: bodoai1/pydough-mongo-tpch:latest
+        env:
+          # Set environment variables for MongoDB container
+          MONGO_HOST: 127.0.0.1
```
john-sanchez31, line 55: I think It'd be better to create the corresponding environment variables for mongo

.github/workflows/trino_testing.yml
```diff
@@ -0,0 +1,146 @@
...
+            POSTGRES_HOST: postgres
+            POSTGRES_DB: "pydough_test"
+            POSTGRES_USER: ${{ secrets.POSTGRES_USER }}
+            POSTGRES_PASSWORD: ${{ secrets.POSTGRES_PASSWORD }}
+            MYSQL_HOST: mysql
+            MYSQL_DB: test
+            MYSQL_USER: ${{ secrets.MYSQL_USERNAME }}
+            MYSQL_PASSWORD: ${{ secrets.MYSQL_PASSWORD }}
+            MONGO_HOST: mongo
+            MONGO_USER: "root_user"
```
john-sanchez31: Those can be reused here (avoiding typing errors)

pydough/sqlglot/transform_bindings/trino_transform_bindings.py
```diff
@@ -0,0 +1,453 @@
...
+            # dow2 = DAYOFWEEK(date2)
+            # result = INTEGER((raw_delta + dow1 - dow2) / 7)
+            date1 = self.make_datetime_arg(args[1])
+            date2 = self.make_datetime_arg(args[2])
+            raw_delta = self.convert_datediff(
+                [sqlglot_expressions.convert("DAY"), date1, date2], types
+            )
+            dow1 = self.convert_dayofweek([date1], [types[1]])
+            dow2 = self.convert_dayofweek([date2], [types[2]])
+            divion = sqlglot_expressions.Div(
```
john-sanchez31:
```suggestion
            division = sqlglot_expressions.Div(
```

tests/test_pydough_functions/defog_test_functions.py
```diff
@@ -1502,7 +1490,7 @@ def impl_defog_ewallet_adv12():
     What is the total transaction amount for each coupon offered by merchant
     with ID 1? Return the coupon ID and total amount transacted with it.
     """
-    return coupons.WHERE(merchant_id == "1").CALCULATE(
+    return coupons.WHERE(merchant_id == 1).CALCULATE(
```
john-sanchez31, line 1493: Why this change? Does it work for all dialects? I know some (Oracle) might have trouble handling this change.

tests/test_sql_refsols/division_by_zero_database_postgres.sql
```diff
@@ -2,5 +2,5 @@ SELECT
   CAST(l_extendedprice AS DOUBLE PRECISION) / l_discount AS computed_value
 FROM tpch.lineitem
 ORDER BY
-  1 NULLS FIRST
+  l_discount NULLS FIRST
```
john-sanchez31, line 5: Why is it using the actual column name instead of the index?

tests/conftest.py
```diff
@@ -1186,6 +1207,346 @@ def container_is_running(name: str) -> bool:
...
+                        "-e",
+                        "MONGO_USER=root_user",
+                        "-e",
+                        "MONGO_PASSWORD=mongo_pwd",
+                        "-e",
+                        "FILTER_TABLES=true",
+                        "-e",
+                        "TPCH_TABLES=region,supplier",
+                        "-e",
+                        "DEFOG_TABLES=sbCustomer,salespersons,payments_received,doctors,diagnoses,concomitant_meds,merchants,wallet_user_balance_daily,user_sessions,cite,domain_author,domain_keyword,keyword,publication_keyword,location",
```
john-sanchez31, line 1274: Why is this needed? When more defog db are added does this need to be updated?

tests/conftest.py
```diff
@@ -1186,6 +1207,346 @@ def container_is_running(name: str) -> bool:
...
+    except ImportError as e:
+        raise RuntimeError("pymongo is not installed") from e
+
+    # Wait for Mongo to be ready
+    for _ in range(300):
+        try:
+            client = MongoClient(
+                host=MONGO_HOST,
+                port=MONGO_PORT,
+                username="root_user",
```
john-sanchez31: Maybe would be better use environment variables. I know the data in the container are not sensitive but still would be a better practice and for consistency.

tests/test_pipeline_custom_datasets.py
john-sanchez31: I agree, needed sooner or later

tests/test_pipeline_trino.py
```diff
@@ -0,0 +1,419 @@
...
+        trino_conn_db_context,
+        config=defog_config,
+        coerce_types=True,
+        atol=5e-3,
+        display_sql=True,
+    )
+
+
+@pytest.mark.skip(
+    "Trino does not properly handle all of the weird quoting cases; investigate at a later time."
```
john-sanchez31, line 401: We can leave it like this but just take in count that other type of tests can be added as `custom_datasets` not just the quoted ones.

tests/testing_utilities.py
```diff
@@ -1536,8 +1542,17 @@ def run_e2e_test(
                 )
 
         # Perform the comparison between the result and the reference solution
+        # print()
```
john-sanchez31: Don't forget to delete these comments

john-sanchez31 (commented):
Great job Kian! There are some comments below but overall almost ready

.github/workflows/trino_testing.yml
hadia206: +1. Moving to variables an re-using

.github/workflows/trino_testing.yml
```diff
@@ -0,0 +1,146 @@
...
+      - name: Install uv
+        uses: astral-sh/setup-uv@v3
+        with:
+          version: "0.6.0"
+
+      - name: Create virtual environment
+        run: uv venv
+
+      - name: Install dependencies
+        run: uv pip install -e ".[trino]"; uv pip install pytest
```
hadia206: Do we still need the `uv pip install pytest`?

pydough/sqlglot/transform_bindings/base_transform_bindings.py
```diff
@@ -254,7 +254,10 @@ def convert_call_to_sqlglot(
             else:
                 fmt_string = operator.macro_text
             combined_string: str = fmt_string.format(*arg_strings)
-            return parse_one(combined_string)
+            return parse_one(
+                combined_string,
+                dialect=self._visitor._session.database.dialect.sqlglot_dialect,
```
hadia206, line 259: Why this change?

pydough/sqlglot/transform_bindings/base_transform_bindings.py
```diff
@@ -359,6 +362,28 @@ def convert_call_to_sqlglot(
...
+            `typ`: The PyDough type of the expression, used to determine if
+            conversion is necessary.
+
+        Return:
+            A SQLGlot expression that is guaranteed to be of string type, either
+            by being the original expression (if it was already a string) or a
+            cast of the original expression to string.
+        """
+        if not isinstance(typ, StringType):
+            return sqlglot_expressions.Cast(this=expr, to="VARCHAR")
```
hadia206: Shouldn't this be
```suggestion
            return sqlglot_expressions.Cast(this=expr, to=sqlglot_expressions.DataType.build("VARCHAR"))
```

pydough/sqlglot/transform_bindings/trino_transform_bindings.py
```diff
@@ -0,0 +1,453 @@
...
+    ) -> SQLGlotExpression:
+        if operator in self.PYDOP_TO_TRINO_FUNC:
+            return sqlglot_expressions.Anonymous(
+                this=self.PYDOP_TO_TRINO_FUNC[operator], expressions=args
+            )
+
+        return super().convert_call_to_sqlglot(operator, args, types)
+
+    def convert_sum(
+        self, arg: SQLGlotExpression, types: list[PyDoughType]
```
hadia206:
```suggestion
        self, args: list[SQLGlotExpression], types: list[PyDoughType]
```

Then update use of arg to args

pydough/sqlglot/transform_bindings/trino_transform_bindings.py
hadia206: Reminder to fix it where it's used too.

pydough/database_connectors/builtin_databases.py
```diff
@@ -136,6 +140,46 @@ def load_snowflake_connection(**kwargs) -> DatabaseConnection:
...
+        import trino
+    except ImportError:
+        raise ImportError(
+            "Trino connector is not installed. Please install it with `pip install trino`."
+        )
+
+    connection = kwargs.pop("connection", None)
+    if connection:
+        return DatabaseConnection(connection)
+    required_keys = ["user", "host", "port"]
```
hadia206, line 172: docstring says that `host` and `port` have default values but they are listed here as a required argument. Shouldn't they be optional and set the values for them internally if not provided?

pydough/database_connectors/builtin_databases.py
```diff
@@ -136,6 +140,46 @@ def load_snowflake_connection(**kwargs) -> DatabaseConnection:
...
+def load_trino_connection(**kwargs) -> DatabaseConnection:
+    """
+    Loads a Trino database connection. This is done by providing a
+    wrapper around the DB 2.0 connect API.
+
+    Args:
+        **kwargs: The keyword arguments which are expected to include either a
+        connection object for Trino or the required connection parameters:
+        - user: Trino username (str)
+        - host: Trino server host (str, default: "127.0.01"/"localhost")
```
hadia206:
```suggestion
        - host: Trino server host (str, default: "127.0.0.1"/"localhost")
```

pydough/sqlglot/transform_bindings/trino_transform_bindings.py
```diff
@@ -0,0 +1,453 @@
...
+        ):
+            # Ensure that REPLACE(x, "", z) -> x
+            return args[0]
+        # Wrap the result in a case statement to ensure the result is the
+        # input if the delimiter is an empty string, unless it is also a
+        # literal that is not an empty string.
+        result: SQLGlotExpression = super().convert_replace(args, types)
+        if (
+            not isinstance(delim_arg, sqlglot_expressions.Literal)
+            and delim_arg.is_string
```
hadia206: Do we need this check `and delim_arg.is_string` ?
Isn't `is_string` only valid on literals?

pydough/sqlglot/transform_bindings/trino_transform_bindings.py
```diff
@@ -0,0 +1,453 @@
...
+                    args[0],
+                )
+                .else_(result)
+            )
+        return result
+
+    def convert_integer(
+        self, args: list[SQLGlotExpression], types: list[PyDoughType]
+    ) -> SQLGlotExpression:
+        # First cast to a DOUBLE, then to an INTEGER, for safety.
```
hadia206: Why? what's the problem if we go directly to INTEGER?

pydough/sqlglot/transform_bindings/trino_transform_bindings.py
```diff
@@ -0,0 +1,453 @@
...
+        return sqlglot_expressions.Cast(
+            this=sqlglot_expressions.Cast(
+                this=args[0], to=sqlglot_expressions.DataType(this="DOUBLE")
+            ),
+            to=sqlglot_expressions.DataType(this="BIGINT"),
+        )
+
+    def generate_dataframe_item_dialect_expression(
+        self, item: Any, item_type: PyDoughType
+    ) -> SQLGlotExpression:
```
hadia206, line 441: Write comment explaining what's the difference between that and the base case.

pydough/sqlglot/transform_bindings/trino_transform_bindings.py
```diff
@@ -0,0 +1,453 @@
...
+        self, args: list[SQLGlotExpression], types: list[PyDoughType]
+    ) -> SQLGlotExpression:
+        # Need to manually ensure that all of the types are strings before
+        # proceeding, since CONCAT_WS in Trino does not support implicit type
+        # conversion.
+        new_args: list[SQLGlotExpression] = []
+        new_types: list[PyDoughType] = []
+        for arg, typ in zip(args, types):
+            new_args.append(self.ensure_string(arg, typ))
+            new_types.append(StringType())
```
hadia206, line 256: nit:
`new_types` is not really needed. We can just pass `types` as it's. It's not used by `convert_join_strings`

pydough/sqlglot/execute_relational.py
```diff
@@ -629,6 +642,15 @@ def reset_sqlglot_dialect_configuration(dialect: DatabaseDialect) -> None:
                     ]
                 )
             )
+        case DatabaseDialect.TRINO:
+            # Replace the DAYOFWEEK override for the Trino generator with the
+            # default version, since PyDough handles the conversion logic
+            # correctly whereas the SQLGlot version gets confused with multiple
+            # rounds of parsing and unparsing.
```
hadia206: Do we need the comment about PyDough here again?  It makes it confusing, means that we're doing something specific to PyDough.
While my understanding is that, this is resetting back to original value.

pydough/sqlglot/override_simplify.py
```diff
@@ -152,6 +154,7 @@ def _simplify(expression, root=True):
 
         # PyDough Change: new post-order transformations
         node = rewrite_nullif_coalesce(node)
+        node = rewrite_length_literal(node)
```
hadia206, line 162: Please add a small comment on why doing it again post-order.

pydough/sqlglot/override_simplify.py
```diff
@@ -385,3 +388,88 @@ def rewrite_nullif_coalesce(expr: exp.Expression) -> exp.Expression:
...
+    if (
+        isinstance(numerator, exp.Literal)
+        and numerator.is_number
+        and isinstance(denominator, exp.Literal)
+        and denominator.is_number
+    ):
+        num_term = float(numerator.this)
+        denom_term = float(denominator.this)
+        if denom_term != 0 and num_term / denom_term == int(num_term / denom_term):
+            return exp.Literal.number(int(num_term / denom_term))
```
hadia206: Suggested by AI. 
Reason:
`sqlglot.expressions.Div` can represent either integer (`//`) or float (`/`) division depending on context. Replacing `4 / 2` with the integer literal `2` silently changes the type from `float` to `int` for float-division contexts. 
Additionally, the equality check `float == int` can produce false positives for very large numbers due to floating-point representation (e.g., `10**16 + 1 / 1` would fold incorrectly). Using `num_term % denom_term == 0` with integer arithmetic (`int(num_term)` and `int(denom_term)`) where both values are whole numbers would be safer.

```suggestion
        num_term = int(float(numerator.this))
        denom_term = int(float(denominator.this))
        if denom_term != 0 and num_term % denom_term == 0:
            return exp.Literal.number(num_term // denom_term)
```

pydough/sqlglot/execute_relational.py
```diff
@@ -629,6 +642,15 @@ def reset_sqlglot_dialect_configuration(dialect: DatabaseDialect) -> None:
...
+            Trino.Generator.TRANSFORMS[sqlglot_expressions.DayOfWeek] = (
+                lambda self, e: f"(({self.func('DAY_OF_WEEK', e.this)} % 7) + 1)",
```
hadia206: Assuming it's a typo. There was no `,` in the case written beforehand.
```suggestion
                lambda self, e: f"(({self.func('DAY_OF_WEEK', e.this)} % 7) + 1)"
```

tests/test_pipeline_tpch_custom.py
```diff
@@ -6504,6 +6601,17 @@ def _strip_temp_for_oracle(test_data: PyDoughPandasTest) -> PyDoughPandasTest:
     )
 
 
+def get_table_prefix_for_dialect(dialect: DatabaseDialect) -> str:
+    """Return the appropriate table name prefix for the given database dialect."""
+    match dialect:
+        case DatabaseDialect.SNOWFLAKE:
```
hadia206, line 6610: Let's re-add comment here in docstring or above each case.
```
# For Snowflake, use cross-database write (read from SNOWFLAKE_SAMPLE_DATA,
    # write to E2E_TESTS_DB.PUBLIC)
```

tests/test_pydough_functions/simple_pydough_functions.py
```diff
@@ -1982,45 +1982,6 @@ def minutes_seconds_datediff():
     )
 
 
-def simple_week_sampler():
```
hadia206, line 1985: Why was this removed? I see the tpc-h one all combined into one. but this one I don't see it re-added somewhere else or I may have missed it.

tests/testing_utilities.py
hadia206: Reminder

hadia206 (commented):
Good work, Kian!
Please see my comments below.

Also, don't forget the documentation.

tests/test_pydough_functions/defog_test_functions.py
knassre-bodo: It technically is SUPPOSED to be an integer; I don't know why it was ever a string, but doing so was causing problems since comparing integers to strings in Trino is problematic. Rather than adding implicit type casting, it was simpler to just fix the test so it made more sense.

tests/test_sql_refsols/division_by_zero_database_postgres.sql
knassre-bodo: The actual test here was reworked so it orders it in a manner where the only row kept is the one where `l_discount` is 0.

tests/conftest.py
knassre-bodo: > Why is this needed?

Mongo/Cassandra setup time scales a lot with the number of tables. Only having it set up some of them (the ones that Trino will use from these connectors), instead of all of them makes it faster.

> When more defog db are added does this need to be updated?

Yes, unless we choose to make it so those defog db are only ever accessed from the MySQL/Postgres connectors.

.github/workflows/trino_testing.yml
knassre-bodo: This is the only/main part still under development

pydough/sqlglot/transform_bindings/base_transform_bindings.py
knassre-bodo: Discovered the hard way while doing some Trino UDF stuff that the UDF will not always be parsed by SQLGlot correctly unless the dialect is passed in. E.g. if I use `DATE_ADD` inside the SQL macro text, SQLGlot parses the arguments in the wrong order unless the dialect was provided.

pydough/sqlglot/transform_bindings/trino_transform_bindings.py
knassre-bodo: Going string->integer fails in Trino if the string has float components.

pydough/sqlglot/transform_bindings/trino_transform_bindings.py
knassre-bodo: Yes but I'm trying to future proof for when we do have a type system and use it properly. Keeping the types updated is a way to minimize the amount of insane bug catching down the line.

tests/test_pydough_functions/simple_pydough_functions.py
knassre-bodo: It was all combined into one in "test_pipeline_e2e_simple_week", which uses the TPC-H version.

pydough/sqlglot/override_simplify.py
knassre-bodo: Will need to verify that both terms are actually integers... e.g. I don't want to have `6.5 / 3.2` get folded into `3`

pydough/database_connectors/builtin_databases.py
knassre-bodo: They are required if the connection is not provided, not sure where it says they are optional? Do you mean for Trino itself they are optional?

john-sanchez31 (approved):
Great job Kian! Ready to be merged after CI is working

pydough/database_connectors/builtin_databases.py
hadia206: I meant ,are they optional for the user?
The docstring says they can have a default value. So should we used that as a default and not make it required for the user.
From the docstring
```
- host: Trino server host (str, default: "127.0.01"/"localhost")
- port: Trino server port (int, default: 8080)
```

pydough/sqlglot/transform_bindings/trino_transform_bindings.py
hadia206: Reminder^

pydough/sqlglot/transform_bindings/trino_transform_bindings.py
hadia206: Let's clarify that in the comment and be specific about what is safety

pydough/sqlglot/transform_bindings/trino_transform_bindings.py
hadia206: Reminder^

pydough/sqlglot/execute_relational.py
hadia206: Reminder^

pydough/sqlglot/override_simplify.py
hadia206: Did you verify? If yes, what's final decision?

hadia206 (commented):
Thanks Kian.
Some comments were not addressed and also  missing the user documentation.

pydough/sqlglot/transform_bindings/trino_transform_bindings.py
knassre-bodo: No, `.is_string` is allowed on non-Literals in SQLGlot since it is a property of all expressions.

pydough/sqlglot/override_simplify.py
knassre-bodo: Changed it to this:
```
    # Attempt to fold constants for cases like `4 / 2` -> `2`
    if (
        isinstance(numerator, exp.Literal)
        and numerator.is_int
        and isinstance(denominator, exp.Literal)
        and denominator.is_int
    ):
        num_term: int = numerator.this
        denom_term: int = denominator.this
        if denom_term != 0 and num_term % denom_term == 0:
            return exp.Literal.number(num_term // denom_term)
```

PR #507, opened 2026-03-27
Optimization of COUNT on top of SEMI join to NDISTINCT on the RHS

### LINKED ISSUE #491: Optimize COUNT on top of SEMI join to NDISTINCT on the right hand side
Consider the question below and two different implementations:

```py
# How many customers have made an urgent order in 1994?

# Method 1: count how many customers have an order with those properties
selected_orders = orders.WHERE((order_priority == '1-URGENT') & (YEAR(order_date) == 1994))
result = TPCH.CALCULATE(
  n=COUNT(customers.WHERE(HAS(selected_orders))
)

# Method 2: count the number of unique customer keys out of the orders with those properties
selected_orders = orders.WHERE((order_priority == '1-URGENT') & (YEAR(order_date) == 1994))
result = TPCH.CALCULATE(
  n=NDISTINCT(selected_orders.customer_key)
)
```

These two different ways of writing the query are both valid and produce identical answers, but the former does a SEMI join from customers to orders before aggregating, while the latter only scans and aggregates orders (even though it is a `COUNT(DISTINCT)` which is less performant than `COUNT(*)`). Almost always, the second method will be more performant since the cost of the extra scan/join will be worse than the less-performant aggregation.

The goal of this optimization is to rewrite the former case into the latter in relational structure. Here are examples of the two relational structures, the former which is to be optimized and the latter that it should be rewritten to:.

```
AGGREGATE(keys={}, aggregations={'n_rows': COUNT()})
 JOIN(condition=t0.LEFT_UNIQUE_KEY == t1.RIGHT_KEY, type=SEMI, columns={})
  LEFT_TREE
  RIGHT_TREE
```

```
AGGREGATE(keys={}, aggregations={'n_rows': NDISTINCT(RIGHT_KEY)})
 RIGHT_TREE
```

The pre-requisites for the transformation:
* Current node is a no-groupby aggregation where the only aggfunc is `COUNT(*)`
* Input to the node is a SEMI join where the condition is a single equality check where the column form the left hand side is a uniqueness key from the left hand side (e.g. every row of `LEFT_TREE` has a unique value of `LEFT_UNIQUE_KEY`)
* The join has a reverse cardinality that always matches (every row from `RIGHT_TREE` is matched by at least 1 row from `LEFT_TREE`, e.g. each row of `orders` has a match in `customers`).

If this is so, the aggregate should be transformed to get rid of the join & `LEFT_TREE`, replacing the aggregate input with `RIGHT_TREE` and replacing `COUNT()` with `NDISTINCT(RIGHT_KEY)`, where `RIGHT_KEY` is the column from the right hand side that was part of the join condition being checked for equality against.

The reverse cardinality check is important because of queries such as this one:
```py
# How many customers in the building market segment have made an urgent order in 1994?

selected_orders = orders.WHERE((order_priority == '1-URGENT') & (YEAR(order_date) == 1994))
result = TPCH.CALCULATE(
  n=COUNT(customers.WHERE((market_segment == 'BUILDING') & HAS(selected_orders))
)
```

The relational structure for this is as follows:
```
AGGREGATE(keys={}, aggregations={'n_rows': COUNT()})
 JOIN(condition=t0.key == t1.customer_key, type=SEMI, columns={})
  FILTER(market_segment == 'BUILDING')
    SCAN(CUSTOMER)
  FILTER(YEAR(order_date) == 1994)
    SCAN(ORDERS)
```

Here, the reverse cardinality is `SINGULAR_FILTERS`, meaning each row from orders in the right hand side matches onto 0 or 1 rows from the left hand side, so the rewrite as-is is not valid since we would be losing the market segment filter. HOWEVER, the following alternative rewrite is possible:

```
AGGREGATE(keys={}, aggregations={'n_rows': NDISTINCT(key)})
 JOIN(condition=t0.key == t1.customer_key, type=INNER, columns={"key": t0.customer_key})
  FILTER(YEAR(order_date) == 1994)
    SCAN(ORDERS)
  FILTER(market_segment == 'BUILDING')
    SCAN(CUSTOMER)
```

^ This alternative rewrite is possible and operates on the same principle, just that if the reverse cardinality is not always matching then instead of pruning the join and LHS, we just flip the join inputs/cardinalities, pass the former-rhs key (now lhs) for the `NDISTINCT` to use, and make the join type INNER. However, this case is **NOT** as clear-cut of an optimization (since it still does a join, even though `INNER` is better than `SEMI`, but has also downgraded from `COUNT(*)` to `NDISTINCT`, so the performance being an improvement vs downgrade will depend on the relative sizes of the tables and the number of unique rows. 

Ignoring this alternative case, the main optimization can be added as a special case in the redundant aggregation removal optimization in `agg_removal.py`. Inside this file, the function `aggregation_uniqueness_helper` is the main recursive protocol responsible for traversing the tree, figuring out which columns (or combinations of columns) are unique, and removing redundant joins if the grouping columns are already unique. 

Inside the `Aggregate` case of this function, if the removal optimization did _not_ trigger, then this new optimization can be inserted as a call to a helper function dealing with the new logic. See below:

```py
        case Aggregate():
            node._input, input_uniqueness = aggregation_uniqueness_helper(node.input)
            agg_keys: frozenset[str] = frozenset(node.keys)
            output_uniqueness: set[frozenset[str]] = {agg_keys}
            for unique_set in input_uniqueness:
                if agg_keys.issuperset(unique_set):
                    node = delete_aggregation(node)
                    # If deleting the aggregation, then the uniqueness of the
                    # input is propagated through the new projection.
                    output_uniqueness = bubble_uniqueness(
                        input_uniqueness, node.columns, None
                    )
                    break
            # NEW CODE (if the optimziation does not fire, just returns `node` and `output_uniqueness` as-is.
            if isinstance(node, Aggregate):
              node, output_uniqueness = rewrite_count_semi(node, input_uniqueness, output_uniqueness)
            return node, output_uniqueness
```

There are several existing tests that cover this behavior, but new ones will need to be written:
* cryptbank (masked column sqlite tests): `filter_count_15`, `filter_count_16`, `general_join_02`
* sf_masked (masked column snowflake tests): `patient_claims`
* misc TPC-H tests (all in test_pipeline_tpch_custom): `redundant_has_on_plural`, `redundant_has_on_plural_lineitems`, `has_cross_correlated_singular`

In particular the two example questions shown should be added as tests, as should more tests covering various edge cases:
* Left hand side has a more complex subtree (e.g. `customers.WHERE((nation.region == 'ASIA') & (nation.name != 'CHINA'))`)
* All different sorts of combinations of cardinality between left vs right subtree. A few examples:
   * LEFT->RIGHT=PLURAL_FILTER, RIGHT->LEFT=SINGULAR_FILTER (should not rewrite). Example: left = nations in europe, right = customers in those nation in the building market segment (every such nation can have 0+ such customers, every customer has 0 or 1 such nations), aggregation would be counting the number of nations with at least 1 such customer
   * LEFT->RIGHT=PLURAL_FILTER, RIGHT->LEFT=SINGULAR_ACCESS (should rewrite). Example: left = all nations, right = customers in those nation in the automible market segment with an account balance below $-975 (every such nation can have 0+ such customers, every customer has 0 or 1 such nations), aggregation would be counting the number of nations with at least 1 such customer (would be transformed to do NDISTINCT on the nation keys of the customers)
   * LEFT->RIGHT=SINGULAR_FILTER, RIGHT->LEFT=PLURAL_FILTER (should not rewrite)
   * LEFT->RIGHT=SINGULAR_FILTER, RIGHT->LEFT=PLURAL_ACCESS (should rewrite)

john-sanchez31:
Resolves #491


pydough/relational/rel_util.py
```diff
@@ -810,3 +811,95 @@ def add_input_name(
...
     return expr
+
+
+def rewrite_count_ndistinct(
+    node: Aggregate,
+    input_unique_sets: set[frozenset[str]],
+) -> RelationalNode:
+    """
+
+    This function optimize COUNT on top of SEMI join to NDISTINCT on the right
```
john-sanchez31: Fix docstring (include INNER join)

tests/test_plan_refsols/redundant_has_on_plural.txt
```diff
@@ -1,6 +1,4 @@
-ROOT(columns=[('n', n_rows)], orderings=[])
```
john-sanchez31: Good example of the optimization

pydough/relational/rel_util.py
```diff
@@ -810,3 +811,95 @@ def add_input_name(
...
+    Pre-requisites for the transformation:
+        - Current node is a no-groupby aggregation where the only aggfunc is
+        COUNT(*)
+        - Input to the node is a SEMI join where the condition is a single equality
+        check where the column form the left hand side is a uniqueness key from
+        the left hand side.
+        - The join has a reverse cardinality that always matches.
+
+    Args:
+        `node`: The node being transformed
```
hadia206: missing `input_unique_sets`

pydough/relational/rel_util.py
```diff
@@ -810,3 +811,95 @@ def add_input_name(
...
+        ):
+            rhs_key, lhs_key = cond.inputs
+        else:
+            return node
+    else:
+        return node
+
+    assert isinstance(lhs_key, ColumnReference)
+
+    if agg_input.join_type in (JoinType.SEMI, JoinType.INNER):
```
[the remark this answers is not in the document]
hadia206: Is this a reminder or a followup?

hadia206 (commented):
Overall looks good.

I don’t see change to these tests that are mentioned in the issue `filter_count_15`, `filter_count_16`, `general_join_02` , `patient_claims` and `has_cross_correlated_singular`


Let’s add the test mentioned in the issue as well
```
# How many customers in the building market segment have made an urgent order in 1994?

selected_orders = orders.WHERE((order_priority == '1-URGENT') & (YEAR(order_date) == 1994))
result = TPCH.CALCULATE(
  n=COUNT(customers.WHERE((market_segment == 'BUILDING') & HAS(selected_orders))
)
```

knassre-bodo (commented):
Well done John! A few things to potentially address, but this is close to being done :)

pydough/conversion/agg_removal.py
```diff
@@ -262,6 +263,12 @@ def aggregation_uniqueness_helper(
                         input_uniqueness, node.columns, None
                     )
                     break
+
+            if isinstance(node, Aggregate):
+                # If the aggregation was not deleted, attempt to rewrite COUNT to
+                # NDISTINCT to a non-aggregate expression if the input
+                # uniqueness allows for it
```
knassre-bodo: Since we generalized this, let's rephrase:
```suggestio
```suggestion
                # If the aggregation was not deleted, attempt to rewrite
                # aggregations atop joins into an aggregation atop one of the 
                # join's input if skipping the join and aggregating the input directly
                # would produce the same answer. For example, COUNT(*) on
                # the output of a SEMI join could become NDISTINCT on the
                # join key from the right hand side.
```

pydough/conversion/agg_split.py
```diff
@@ -396,8 +396,13 @@ def attempt_join_aggregate_transpose(
             ), False
 
     # Keep a dictionary for the projection columns that will be used to post-process
-    # the output of the aggregates, if needed.
+    # the output of the aggregates, if needed. Start it out as an identity map.
```
knassre-bodo:
```suggestion
    # the output of the aggregates, if needed. Initialize as an identity
    # mapping.
```

pydough/relational/rel_util.py
```diff
@@ -810,3 +811,149 @@ def add_input_name(
...
+
+    # Join must be SEMI or INNER
+    if agg_input.join_type not in (JoinType.SEMI, JoinType.INNER):
+        return node
+
+    cond: RelationalExpression = agg_input.condition
+
+    # The condition is a single equality check where the column form the left
+    # hand side is a uniqueness key from the left hand side
+    # (e.g. every row of LEFT_TREE has a unique value of LEFT_UNIQUE_KEY)
```
knassre-bodo: This comment explains what several steps here do. I'd break it down step by step for each of these chunks.

1. Ensure the join condition is in the form `column_ref_a == column_ref_b`.
2. Ensure that each column ref comes from a different side of the join (and extract which one is from the LHS vs the RHS). This way, when doing `COUNT(*)` -> `NDISTINCT` rewrite, we know which columns to use for the transformation.
3. Identify which columns come from the RHS of the join.

pydough/relational/rel_util.py
```diff
@@ -810,3 +811,149 @@ def add_input_name(
...
+            and join_col_val.input_name == agg_input.default_input_aliases[1]
+        ):
+            rhs_columns.add(join_col_key)
+
+    new_aggregations: dict[str, CallExpression] = {}
+    new_agg: CallExpression
+    for agg_key, agg_value in node.aggregations.items():
+        match agg_value.op:
+            case pydop.COUNT if not agg_value.inputs:
+                # LHS key must be unique
```
knassre-bodo, line 921: I'd add a comment explaining this case with an example. For instance:

```suggestion
            case pydop.COUNT if not agg_value.inputs:
                """
                Apply the following equivalent transformation:
               
                   SELECT COUNT(*)
                   FROM T1
                   SEMI JOIN T2
                   ON T1.x = T2.y
                
                Becomes:
               
                   SELECT COUNT(DISTINCT y)
                   FROM T2
                """
                
                # LHS key must be unique
```

pydough/relational/rel_util.py
```diff
@@ -810,3 +811,149 @@ def add_input_name(
...
+
+                # Rewrite COUNT(*) → NDISTINCT(rhs_key)
+                new_agg = CallExpression(
+                    pydop.NDISTINCT,
+                    agg_value.data_type,
+                    [add_input_name(rhs_key, None)],
+                )
+                new_aggregations[agg_key] = new_agg
+
+            case pydop.MIN | pydop.MAX | pydop.ANYTHING:
```
knassre-bodo, line 934: Same for both of these cases, show an example comment of the transformation (can just use MIN for both examples, but make it clear it applies to the other two.

pydough/relational/rel_util.py
```diff
@@ -810,3 +811,149 @@ def add_input_name(
...
+                agg_arg: RelationalExpression = agg_value.inputs[0]
+                assert isinstance(lhs_key, ColumnReference)
+                if (
+                    isinstance(agg_arg, ColumnReference)
+                    and agg_arg.name == lhs_key.name
+                ):
+                    if frozenset([lhs_key.name]) not in input_unique_sets:
+                        return node
+
+                    # Rewrite MIN(lhs_key) → MIN(rhs_key)
```
knassre-bodo, line 956: Make it clear this also applies to MAX/ANYTHING

pydough/relational/rel_util.py
```diff
@@ -810,3 +811,149 @@ def add_input_name(
...
+
+    Returns:
+        If all pre-requisites are met a new node with the transformation, the same
+        node otherwise.
+    """
+    agg_input: RelationalNode = node.input
+
+    # Aggregate must be over a join, with no GROUP BY
+    if not isinstance(agg_input, Join) or node.keys:
+        return node
```
knassre-bodo, line 845: IGNORE EARLIER SUGGESTION FROM THIS COMMENT, WAS NOT WELL THOUGHT OUT AND I DELETED IT

knassre-bodo (commented):
Mostly minor comments for documentation, plus one thing to try experimenting with before we wrap this up.

pydough/relational/rel_util.py
```diff
@@ -810,3 +811,149 @@ def add_input_name(
...
+
+    # Reverse cardinality that always matches
+    if not agg_input.reverse_cardinality.accesses:
+        return node
+
+    # Join must be SEMI or INNER
+    if agg_input.join_type not in (JoinType.SEMI, JoinType.INNER):
+        return node
```
knassre-bodo, line 853: I think I just realized a big bug: if the join type is inner, then the **forward** cardinality needs to be singular for the `COUNT` optimization to work.

E.g. if doing `COUNT(*)` on a join of regions -> nations on `r_regionkey == n_regionkey`, we cannot transform into `COUNT(DISTINCT n_regionkey)` on nations because that returns 5 instead of 25.

You may want to change how the conditions here work so different conditions are checked depending on the type of join and which aggregates are occurring. E.g. inside the MATCH cases,  switch on the join type, and check different conditions accordingly?

knassre-bodo (approved):
Ignore what I said about an experimentation idea; we'll leave that for later. This can be approved as-is, just work on the comments (and address the bugfix I mentioned) before merging.

pydough/relational/rel_util.py
```diff
@@ -813,93 +813,181 @@ def add_input_name(
...
         and cond.op == pydop.EQU
         and len(cond.inputs) == 2
+        and isinstance(cond.inputs[0], ColumnReference)
+        and isinstance(cond.inputs[1], ColumnReference)
     ):
         return node
 
+    # Ensure the each column reference is from a different side of the join,
+    # and assign the LHS and RHS keys accordingly. This way, when doing
+    # COUNT(*) -> NDISTINCT rewrite, we know which columns to use for the
```
hadia206: Is this still applicable only to COUNT? I see code mention  MIN, MAX, ANYTHING

hadia206 (approved):
Looks good to me. 
I checked that comments from previous reviews were addressed.

pydough/relational/rel_util.py
john-sanchez31: Yes I changed to be more general

PR #508, opened 2026-04-08
Make dealership_adv8 Test For SQL Files Check Deterministic 

hadia206:
- Fix flaky `dealership_adv8` SQL files test: `impl_defog_dealership_adv8` asks “what are the sales metrics for the last 6 months?” It computes those 6 month boundaries in Python using today’s date, then adds them as hardcoded timestamps into the generated SQL (e.g. `VALUES ('2025-10-01'), ('2025-11-01'), ...`). The SQL comparison test works by saving that generated SQL to a reference file and comparing future runs against it. Since the 6-month window shifts every month, the reference file goes stale on the 1st of every month and the SQL file test fails.

- Solution: Add a `fixed_today: pd.Timestamp | None` field to `PyDoughSQLComparisonTest`. When set, `pd.Timestamp.today` is patched during SQL generation so the output is deterministic. The reference files are pinned to the `Oct 2025–Mar 2026 window (today = 2026-04-01)`. Execution tests are unaffected, they use SQL-native date expressions `(DATE('now', ...))`.

AND

`Migrate [tool.uv] dev-dependencies` to `[dependency-groups] dev` in `pyproject.toml` to silence a uv deprecation warning.

### DIFF
--- a/.github/workflows/bodosql_testing.yml
+++ b/.github/workflows/bodosql_testing.yml
@@ -28,7 +28,7 @@ jobs:
       - name: Install uv
         uses: astral-sh/setup-uv@v3
         with:
-          version: "0.4.23"
+          version: "0.6.0"
 
       - name: Create virtual environment
         # uv requires an existing virtual environment to install packages.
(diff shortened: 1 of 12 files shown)


john-sanchez31 (approved):
If it doesn't affect the e2e tests we're good

juankx-bodo (approved):

PR #509, opened 2026-04-10
Refactor explain to a common help 

hadia206:
Extracts shared logic from `explain.py` into a new `pydough/exploration/_common.py` module, laying groundwork for the upcoming explain_llm API.

No behavior changes. The user-facing output of `pydough.explain()` is identical.

**Changes:**
- Replaces inline term-gathering loop with `extract_terms()`
- Replaces inline AND-condition splitting with `extract_conditions()`
- Replaces bare `try/except `around `qualify_node` with `qualify_safely()`, which returns `(result, error)` instead of raising
- Adds `find_source_collection()` for walking the `preceding_context` chain to the `root` `TableCollection`

### DIFF
--- a/pydough/exploration/_common.py
+++ b/pydough/exploration/_common.py
@@ -1,0 +1,117 @@
+"""
+Shared helpers used by both `explain` and `explain_llm`.
+"""
+
+__all__ = [
+    "extract_conditions",
+    "extract_terms",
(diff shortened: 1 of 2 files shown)


john-sanchez31 (approved):
Looks good, thanks Hadia!

PR #511, opened 2026-04-20
Explain LLM Version

### LINKED ISSUE #512: Create `explain_llm` API for LLM queries
Add a new `pydough.explain_llm()` function that returns a structured description of a PyDough collection expression, designed for use in LLM-assisted text-to-PyDough pipelines. 

The existing `explain()` API produces human-readable description which can be difficult for a model to parse programmatically. 
`explain_llm` returns a deterministic, machine-parseable output that an LLM can use to verify whether generated PyDough code correctly answers a user's question.

hadia206:
## Linked ticket

Closes #512 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Docs / config

## What changed and why?

Adds `pydough.explain_llm()`, a new exploration API that returns a structured description of a PyDough collection expression designed for LLM consumption.

Unlike `explain()` (human-readable prose for interactive debugging), `explain_llm` returns a stable, machine-parseable payload, either a JSON dict (`format="json"`) or markdown (`format="md"`), so LLM judges can validate and self-correct generated PyDough code.

**Key design decisions:**

- **`available_terms` lives under a `debug` sub-key**: scope information is preserved for human debugging but kept out of the main payload so judge prompts aren't distracted by fields irrelevant to correctness.

- **Implicit scoping is made explicit** : PyDough's relationship-navigation scoping (e.g. `COUNT(orders)` inside `customers.CALCULATE(...)`) is correct but looks like a missing filter to a naive judge. The structured
  `implicit_scope_note` field and per-step `notes` surface this so evaluators can distinguish "missing filter" from "implicitly scoped via relationship navigation."

- **Structured error taxonomy**: 13 error categories (`unrecognized_term`, `plural_in_calculate`, `bad_window_per`, etc.) with `error_type`, `details`, and `hint` fields, so the LLM gets actionable guidance without parsing the raw exception message.

- **`conditions` as structured dicts** — Where step conditions expose `operator`, `left`, and `right` as parsed sub-dicts (not just strings), so a judge can inspect operands directly.

**Output shape (success):** 
- `query_summary` (one deterministic sentence)
- `steps` (ordered operations)
- `schema` (source collection, output columns + types, ordering, limit).

**On error:** always `{"error": true, "message": ..., "error_type": ..., "details": ..., "hint": ..., "steps": [], "schema": null}`.

**Implementation:**
- New shared helpers in `_common.py`: `describe_expression`, `describe_subcollection_arg`, `generate_step_notes`, `generate_query_summary`, `_cond_texts`, `_collation_entry`
- New `pydough/exploration/explain_llm.py`: qualification, step-walking, schema building, error classification, markdown render.

## How I tested this?

 - 82 tests in `tests/test_explain_llm.py`
 - local testing and CI
 - Testing with LLM team


pydough/exploration/_common.py
```diff
@@ -115,3 +154,623 @@ def qualify_safely(
...
+    Args:
+        `node`: a ``TableCollection`` QDAG node.
+
+    Returns:
+        ``True`` when the node was produced by a CROSS join, ``False``
+        otherwise.
+    """
+    return (
+        isinstance(node.ancestor_context, GlobalContext)
+        and node.ancestor_context.ancestor_context is not None
```
hadia206, line 191: A plain root table access has `GlobalContext` as `ancestor_context`, but that `GlobalContext`'s own `ancestor_context` is `None`. `CROSS` qualification inserts an extra intermediate `GlobalContext` whose `ancestor_context` IS set. 
That nesting is the only reliable signal that distinguishes CROSS from a normal table access.

pydough/exploration/_common.py
```diff
@@ -115,3 +154,623 @@ def qualify_safely(
...
+          the child list.  Needed to resolve ``ChildReferenceCollection``
+          arguments; pass ``None`` when not available.
+
+    Returns:
+        A dict with at least ``{"kind": str, "text": str}``.  Additional
+        keys depend on the expression kind.
+    """
+    text: str = expr.to_string()
+
+    match expr:
```
hadia206, line 367: `ChildReferenceExpression` and `BackReferenceExpression` are both subclasses of `Reference`, they must be matched before `case Reference()` or they fall through to the wrong branch.

pydough/exploration/_common.py
```diff
@@ -115,3 +154,623 @@ def qualify_safely(
...
+                f"The partition key(s) {keys} are available inside child "
+                f"scope '{child_name}' but not outside it."
+            )
+
+        case Calculate():
+            for term_name, detail in step.get("term_details", {}).items():
+                if detail.get("kind") != "Aggregation":
+                    continue
+                for arg in detail.get("args", []):
+                    implicit_note = arg.get("implicit_scope_note")
```
hadia206: Two distinct cases: 

1. non-null `implicit_scope_note` means the collection IS correctly scoped via relationship navigation (correct PyDough pattern, so emit an informational note, not a warning). e.g. `customers.CALCULATE(n=COUNT(orders))`

2. null means the collection may be unscoped relative to a cross-product context, only then warn. e.g. `nations.CROSS(regions).CALCULATE(n=COUNT(orders))` this is potentially wrong since `orders` has no relationship to `nations` or `regions`. In this case `access_path = []` so `implicit_scope_note` is `null`. If the `CALCULATE` doesn't filter on any of the CROSS-introduced terms, the `COUNT` aggregates all `orders` for every row, which is likely a bug.

documentation/usage.md
```diff
@@ -1327,6 +1328,200 @@ For example, the following is valid:
...
+
+Each entry in `steps` represents one operation in execution order (earliest first). Every step has:
+* `order` — 1-based position in the chain
+* `type` — stable string tag: `"GlobalContext"`, `"TableCollection"`, `"Cross"`, `"SubCollection"`, `"Where"`, `"Calculate"`, `"OrderBy"`, `"TopK"`, `"PartitionBy"`, `"PartitionChild"`, `"Singular"`, `"UserGeneratedCollection"`
+* `description` — short human-readable phrase
+* `debug` — `{"available_terms": {"expressions": [...], "collections": [...]}}` — scope information, separated from the main payload so LLM judges are not distracted by fields irrelevant to correctness
+* `notes` — list of strings; always present, may be empty
+
+The `schema` section (when `"error"` is `False`) includes:
+* `source_collection` — the root table name, or `null` for graph-level expressions
```
john-sanchez31: NIT:
```suggestion
* `source_collection` — the root collection name, or `null` for graph-level expressions
```

documentation/usage.md
```diff
@@ -1327,6 +1328,200 @@ For example, the following is valid:
...
+pydough.explain_llm(result)
+```
+
+```json
+{
+  "error": false,
+  "query_summary": "Accesses 'nations', filtered to rows where region.name == 'ASIA', selecting key, name.",
+  "steps": [
+    {
+      "order": 1, "type": "GlobalContext",
```
john-sanchez31:
```suggestion
"order": 1, 
"type": "GlobalContext",
```
Same applies to all steps

pydough/exploration/_common.py
```diff
@@ -87,10 +113,25 @@ def find_source_collection(node: PyDoughCollectionQDAG) -> str | None:
...
-        current = getattr(current, "preceding_context", None)
+        elif isinstance(current, ChildOperatorChildAccess):
+            current = current.child_access
+        elif isinstance(current, PartitionBy) and hasattr(current, "child"):
+            # PartitionBy.ancestor_context jumps directly to GlobalContext,
+            # skipping the collection being partitioned (e.g.
+            # customers.WHERE(...).orders).  Get that chain via child instead.
+            current = current.child
+        else:
+            nxt = getattr(current, "preceding_context", None)
```
john-sanchez31: Type hint

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+    name: str = (
+        access_path[0]
+        if access_path
+        else (find_source_collection(collection) or "unknown")
+    )
+
+    # Non-empty access_path → row-level scoping via relationship navigation.
+    implicit_scope_note: str | None = None
+    if access_path:
+        path_str = " → ".join(f"'{p}'" for p in access_path)
```
john-sanchez31:
```suggestion
        path_str: str = " → ".join(f"'{p}'" for p in access_path)
```

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+
+    Inside a ``CALCULATE``, aggregation arguments are represented as
+    ``ChildReferenceCollection`` nodes that point to the parent's child list
+    by index.  This function follows that indirection so
+    ``describe_subcollection_arg`` receives the raw ``SubCollection`` (or
+    user-generated collection) rather than a reference wrapper.
+
+    Args:
+        `arg`: the collection arg from an ``ExpressionFunctionCall``.
+        `parent`: the parent ``Calculate`` (or other child operator) that owns
```
john-sanchez31:
```suggestion
        `parent`: the parent ``CALCULATE`` (or other child operator) that owns
```

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+        The resolved ``PyDoughCollectionQDAG``, or ``arg`` unchanged if the
+        resolution path is unavailable.
+    """
+    if (
+        isinstance(arg, ChildReferenceCollection)
+        and parent is not None
+        and hasattr(parent, "children")
+        and arg.child_idx < len(parent.children)
+    ):
+        child = parent.children[arg.child_idx]
```
john-sanchez31: Type hint

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+
+    Returns:
+        A single sentence ending with ``.``.
+    """
+    parts: list[str] = []
+
+    # ------------------------------------------------------------------ #
+    # 1. Subject                                                           #
+    # ------------------------------------------------------------------ #
+    cross_step = next((s for s in steps if s["type"] == "Cross"), None)
```
john-sanchez31: type hint

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+    else:
+        # Global-level CALCULATE: infer subject from aggregation args across
+        # ALL Calculate steps (chained CALCULATEs put aggregations in earlier
+        # steps, not necessarily the last one).
+        agg_colls: list[str] = []
+        for s in steps:
+            if s["type"] != "Calculate":
+                continue
+            for tname in s.get("terms", []):
+                detail = s.get("term_details", {}).get(tname, {})
```
john-sanchez31: type hint

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+            # Generic built-in function call
+            return {
+                "kind": "FunctionCall",
+                "text": text,
+                "function": op.function_name,
+                "is_aggregation": op.is_aggregation,
+            }
+
+        case WindowCall():
+            op = expr.window_operator
```
john-sanchez31, line 494: type hint

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+        case ExpressionFunctionCall():
+            op = expr.operator
+            # Aggregation/predicate operators whose single arg is a collection
+            # get the full subcollection description.
+            if (
+                op in (pydop.COUNT, pydop.NDISTINCT)
+                and len(expr.args) == 1
+                and isinstance(expr.args[0], PyDoughCollectionQDAG)
+            ):
+                resolved = _resolve_collection_arg(expr.args[0], parent)
```
john-sanchez31: type hint

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+        parts.append(
+            "then subcollection filtered to rows where " + " and ".join(sub_conds)
+        )
+
+    # ------------------------------------------------------------------ #
+    # 3. Partition                                                         #
+    # ------------------------------------------------------------------ #
+    partition_step = next((s for s in steps if s["type"] == "PartitionBy"), None)
+    if partition_step:
+        keys_str = ", ".join(partition_step.get("keys", []))
```
john-sanchez31: type hint

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+        parts.append("filtered to rows where " + " and ".join(top_conds))
+    if sub_conds:
+        parts.append(
+            "then subcollection filtered to rows where " + " and ".join(sub_conds)
+        )
+
+    # ------------------------------------------------------------------ #
+    # 3. Partition                                                         #
+    # ------------------------------------------------------------------ #
+    partition_step = next((s for s in steps if s["type"] == "PartitionBy"), None)
```
john-sanchez31: type hint

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+                    ref_terms.append(f"{_n} ({_text})")
+                else:
+                    ref_terms.append(f"{_n} (computed)")
+        agg_terms: list[str] = []
+
+        first_calc_idx = next(
+            (i for i, s in enumerate(steps) if s["type"] == "Calculate"),
+            len(steps),
+        )
+        has_partition = any(s["type"] == "PartitionBy" for s in steps[:first_calc_idx])
```
john-sanchez31: type hint

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+        parts.append(f"partitioned by {keys_str}")
+
+    # ------------------------------------------------------------------ #
+    # 4. Compute                                                           #
+    # ------------------------------------------------------------------ #
+    # Use the last Calculate for output column names, but collect aggregation
+    # descriptions from ALL Calculate steps — chained patterns like
+    # .CALCULATE(n=COUNT(...)).TOP_K(...).CALCULATE(col) put the aggregation
+    # in an earlier step, not the final one.
+    calc_step = next((s for s in reversed(steps) if s["type"] == "Calculate"), None)
```
john-sanchez31: type hint

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+                ref_terms.append(_n)
+            elif _kind != "Aggregation":
+                _text = _d.get("text", "")
+                if _text and len(_text) <= _MAX_EXPR_LEN:
+                    ref_terms.append(f"{_n} ({_text})")
+                else:
+                    ref_terms.append(f"{_n} (computed)")
+        agg_terms: list[str] = []
+
+        first_calc_idx = next(
```
john-sanchez31: type hint

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+        )
+        has_partition = any(s["type"] == "PartitionBy" for s in steps[:first_calc_idx])
+        for s in steps:
+            if s["type"] != "Calculate":
+                continue
+            for name in s.get("terms", []):
+                detail = s.get("term_details", {}).get(name, {})
+                if detail.get("kind") != "Aggregation":
+                    continue
+                fn = detail.get("function", "AGG").lower()
```
john-sanchez31: Type hint

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+        first_calc_idx = next(
+            (i for i, s in enumerate(steps) if s["type"] == "Calculate"),
+            len(steps),
+        )
+        has_partition = any(s["type"] == "PartitionBy" for s in steps[:first_calc_idx])
+        for s in steps:
+            if s["type"] != "Calculate":
+                continue
+            for name in s.get("terms", []):
+                detail = s.get("term_details", {}).get(name, {})
```
john-sanchez31, line 923: type hint

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+            compute_parts.append("computing " + ", ".join(agg_terms))
+        if compute_parts:
+            parts.append(" and ".join(compute_parts))
+
+    # ------------------------------------------------------------------ #
+    # 5. Limit / Order                                                     #
+    # ------------------------------------------------------------------ #
+    topk_step = next((s for s in steps if s["type"] == "TopK"), None)
+    order_step = next((s for s in steps if s["type"] == "OrderBy"), None)
+    sort_step = topk_step or order_step
```
john-sanchez31: type hint

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+    ),
+    "not_callable": (
+        "The PyDough object is not callable as a function. "
+        "It may be a field or collection name — access it as an attribute, "
+        "not a function call."
+    ),
+    "sql_error": "The query produced invalid SQL. Check the query structure.",
+    "type_error": "Check that the argument types match what the function expects.",
+    "generic": None,
+}
```
john-sanchez31, line 173: Can we add some comments about this variable? What contains and how/where is used

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+        PyDoughTypeException,
+        PyDoughUnqualifiedException,
+    )
+
+    msg = str(e)
+
+    # ── Not callable — must precede the "Did you mean" check because the
+    #    undefined_function_call error appends "Did you mean to access an
+    #    attribute or method?" which would otherwise be misclassified.
+    if isinstance(e, PyDoughUnqualifiedException) and "not callable" in msg.lower():
```
john-sanchez31, line 214: What would happen if the error messages change? Even if it's a small change would break this.

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+def _build_table_collection_step(node: TableCollection, order: int) -> dict:
+    """
+    Accesses a named collection (table) from the graph.
+
+    CROSS joins also produce a TableCollection but with a nested intermediate
+    GlobalContext as their ancestor (see ``_is_cross``).  When detected, the
+    step type becomes ``"Cross"`` and both collection names are recorded.
+    """
+    expr_names, coll_names = extract_terms(node)
+    debug = {"available_terms": {"expressions": expr_names, "collections": coll_names}}
```
john-sanchez31: type hint

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+
+    if _is_cross(node):
+        assert node.ancestor_context is not None
+        assert node.ancestor_context.ancestor_context is not None
+        left_name = node.ancestor_context.ancestor_context.name
```
john-sanchez31: type hint

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+            "description": (
+                f"CROSS join: every row of '{left_name}' paired with every "
+                f"row of '{right_name}'."
+            ),
+            "left": left_name,
+            "right": right_name,
+            "debug": debug,
+        }
+
+    collection_name = node.collection.name
```
john-sanchez31: type_hint

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+
+
+def _build_sub_collection_step(node: SubCollection, order: int) -> dict:
+    """
+    Traverses a subcollection relationship (implicit JOIN in SQL).
+
+    The property name (e.g. ``"orders"``) and the collections it connects are
+    recorded.  The caller can use these to understand relationship navigation.
+    """
+    prop = node.subcollection_property
```
john-sanchez31: type hint

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+
+def _build_calculate_step(node: Calculate, order: int) -> dict:
+    """
+    Adds computed expressions to the collection.
+
+    ``term_details`` maps each new term name to its ``describe_expression``
+    dict.  Terms are sorted by their declared position so the order is stable
+    and matches the user's CALCULATE(...) call.
+    """
+    sorted_terms = sorted(
```
john-sanchez31: type hint

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+
+def _build_order_by_step(node: OrderBy, order: int) -> dict:
+    """
+    Sorts the collection by one or more collation keys.
+
+    Each key records direction (``"ASC"``/``"DESC"``) and null placement
+    (``"FIRST"``/``"LAST"``).  TopK is a subclass of OrderBy so the same
+    builder handles both; ``limit`` is included only for TopK.
+    """
+    collation = [_collation_entry(col) for col in node.collation]
```
john-sanchez31: type hint

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+
+def _build_partition_by_step(node: PartitionBy, order: int) -> dict:
+    """
+    Partitions the collection on one or more key expressions.
+
+    ``keys`` are the partitioning field names; ``child_name`` is the name of
+    the subcollection holding the per-partition rows (accessible inside
+    aggregations).
+    """
+    keys = [
```
john-sanchez31: type hint

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+    e: BaseException,
+) -> tuple[str, dict, str | None]:
+    """
+    Returns ``(error_type, details, hint)`` for a PyDough exception.
+
+    ``error_type`` is a stable machine-readable tag.  ``details`` carries
+    structured fields extracted from the message (e.g. the wrong term name
+    and the list of suggestions).  ``hint`` is the actionable guidance string
+    for the LLM — ``None`` when no specific hint is available.
+    """
```
john-sanchez31, line 201: Can we add Args/Returns to all functions on this file?

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+    # but its ancestor_context points to the Where/TableCollection that filters
+    # the parent collection before the subcollection hop.  We walk the full
+    # preceding_context chain of each ancestor so we don't miss nodes like a
+    # TableCollection that sits beneath a Where in the parent chain.
+    #
+    # Special case — PartitionBy: its ancestor_context jumps directly to
+    # GlobalContext, bypassing the collection being partitioned.  The real
+    # pre-partition chain (e.g. customers.WHERE(...).orders) is accessible via
+    # child.ancestor_context instead.
+    top = nodes[-1]
```
john-sanchez31: type hint

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+    Returns:
+        The schema dict.
+    """
+    expr_names, _ = extract_terms(root)
+    # Walk up to the nearest Calculate for output columns — TopK/OrderBy sit
+    # above a Calculate and shouldn't hide its terms.
+    calc_for_cols: PyDoughCollectionQDAG | None = root
+    while calc_for_cols is not None and not isinstance(calc_for_cols, Calculate):
+        calc_for_cols = getattr(calc_for_cols, "preceding_context", None)
+    output_columns = (
```
john-sanchez31: type hint

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+            break
+        if isinstance(current, OrderBy):
+            ordering = [_collation_entry(col) for col in current.collation]
+            break
+        current = getattr(current, "preceding_context", None)
+
+    # find_source_collection returns None for global-level CALCULATEs; fall
+    # back to inspecting aggregation args across the QDAG chain so the schema
+    # still names the collection(s) being accessed.
+    source = find_source_collection(root)
```
john-sanchez31: type hint

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+        "limit": limit,
+    }
+
+
+# ---------------------------------------------------------------------------
+# Markdown renderer
+# ---------------------------------------------------------------------------
+
+
+def _expr_display(detail: dict) -> str:
```
john-sanchez31, line 1060: Remember Args/Returns on the doctstring

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+    """
+    if detail.get("kind") == "Reference":
+        return detail.get("term_name", "?")
+    return detail.get("text", detail.get("kind", "?"))
+
+
+def _render_step_body(step: dict) -> list[str]:
+    """Returns the type-specific body lines for a single step dict."""
+    lines: list[str] = []
+    stype = step["type"]
```
john-sanchez31: type hint

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+        # Structured hint — rendered from the classified field, not string-matched
+        hint = result.get("hint")
+        if hint:
+            lines.append("")
+            lines.append(f"> **Constraint for next attempt:** {hint}")
+
+        return "\n".join(lines)
+
+    # ------------------------------------------------------------------ #
+    # Key Facts — quick-reference block at the top so the judge sees the  #
```
john-sanchez31: NIT: I think we shouldn't mention the judge here. Maybe change this comments for one more general?

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+    # most checkable facts before reading any steps.                      #
+    # ------------------------------------------------------------------ #
+    schema = result["schema"]
```
john-sanchez31: type hints

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+    for step in result["steps"]:
+        lines.append(f"### Step {step['order']} — {step['type']}")
+        lines.append("")
+        lines.append(step["description"])
+        lines.append("")
+
+        body = _render_step_body(step)
+        lines.extend(body)
+
+        notes = step.get("notes", [])
```
john-sanchez31: type hint

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+
+    # ------------------------------------------------------------------ #
+    # Schema                                                               #
+    # ------------------------------------------------------------------ #
+    lines.append("## Schema")
+    lines.append("")
+
+    lines.append(f"- **Source collection:** {f'`{src}`' if src else '_(none)_'}")
+
+    output_cols = schema.get("output_columns", [])
```
john-sanchez31: type hints

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+    if error is not None:
+        result = _error_payload(error)
+        return _render_md(result) if format == "md" else result
+
+    if isinstance(qualified, PyDoughExpressionQDAG):
+        msg = (
+            f"Expected a collection, but received an expression: "
+            f"{qualified.to_string()}. Did you mean to use explain_term?"
+        )
+        result = _error_payload(msg)
```
john-sanchez31, line 1388: type hint

john-sanchez31 (commented):
Great job Hadia! Please check my comments below before merging. Most are related to docstrings and type hints.

pydough/exploration/_common.py
juankx-bodo: same for `table_step` & `user_step `

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+        # steps, not necessarily the last one).
+        agg_colls: list[str] = []
+        for s in steps:
+            if s["type"] != "Calculate":
+                continue
+            for tname in s.get("terms", []):
+                detail = s.get("term_details", {}).get(tname, {})
+                if detail.get("kind") == "Aggregation":
+                    for arg_d in detail.get("args", []):
+                        cname = arg_d.get("name")
```
juankx-bodo: type hint

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+    # ------------------------------------------------------------------ #
+    # 2. Filter                                                            #
+    # ------------------------------------------------------------------ #
+    # Split WHERE conditions by hierarchy level: conditions before the first
+    # SubCollection belong to the root collection; conditions after belong to
+    # a deeper subcollection and must be described separately so the judge
+    # understands they filter a different level of the data.
+    top_conds: list[str] = []
+    sub_conds: list[str] = []
+    past_first_sub = False
```
juankx-bodo:
```suggestion
    past_first_sub: bool = False
```

pydough/exploration/_common.py
juankx-bodo: same for `arg_name `

pydough/exploration/_common.py
juankx-bodo: for `topk_step`, `order_step`, `sort_step`, `collation`, `by_str`, `suffix ` & `summary `

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+    """
+    from pydough.errors import (
+        PyDoughQDAGException,
+        PyDoughSessionException,
+        PyDoughSQLException,
+        PyDoughTypeException,
+        PyDoughUnqualifiedException,
+    )
+
+    msg = str(e)
```
juankx-bodo:
```suggestion
    msg: str = str(e)
```

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+    #    undefined_function_call error appends "Did you mean to access an
+    #    attribute or method?" which would otherwise be misclassified.
+    if isinstance(e, PyDoughUnqualifiedException) and "not callable" in msg.lower():
+        return "not_callable", {}, _HINT_MAP["not_callable"]
+
+    # ── Unrecognized term with suggestions ──────────────────────────────────
+    if "Did you mean" in msg:
+        details: dict = {}
+        # Extract the wrong term: "Unrecognized term of ...: 'TERM'."
+        term_match = _re.search(r":\s*'([^']+)'", msg)
```
juankx-bodo:
```suggestion
        term_match: Match[str] | None = _re.search(r":\s*'([^']+)'", msg)
```

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+        if term_match:
+            details["term"] = term_match.group(1)
+        # Extract suggestions: "Did you mean: a, b, c?"
+        sugg_match = _re.search(r"Did you mean:\s*([^?]+)\?", msg)
```
juankx-bodo:
```suggestion
        sugg_match: Match[str] | None = _re.search(r"Did you mean:\s*([^?]+)\?", msg)
```

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+) -> dict:
+    """
+    Builds a structured error payload from an exception or a raw message string.
+
+    The payload always has a stable shape so the markdown renderer and any
+    programmatic consumer can handle every error the same way.
+    """
+    if isinstance(e, str):
+        message = e
+        details: dict[str, object]
```
juankx-bodo, line 315: What would be the value of `details` in this case?

juankx-bodo (approved):

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
+"""
+Implementation of ``pydough.explain_llm``, which returns a structured JSON-
+serialisable dict describing a PyDough collection expression.
+
+The output is designed for LLM consumption: deterministic shape, stable
+``kind`` tags, and explicit scoping notes so a model can self-correct without
+parsing prose.
+
+Output schema (success)::
```
knassre-bodo, line 9: Should there be two colons here? I don't know the intended format.

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+            "source_collection": "<name or null>",
+            "available_expressions": ["…"],
+            "output_columns": ["…"],
+            "column_types": {"name": "string", …},
+            "ordering": [{"text": "name", "direction": "ASC", "nulls": "LAST"}],
+            "limit": null
+        }
+    }
+
+Output schema (error)::
```
knassre-bodo, line 40: Same here:

tests/test_explain_llm.py
```diff
@@ -0,0 +1,1340 @@
...
+    """A PyDoughSession loaded with the TPCH graph (no DB connection needed)."""
+    graph: GraphMetadata = get_sample_graph("TPCH")
+    session = PyDoughSession()
+    session.metadata = graph
+    return session
+
+
+@pytest.fixture
+def tpch_graph(get_sample_graph: graph_fetcher) -> GraphMetadata:
+    return get_sample_graph("TPCH")
```
knassre-bodo: These can be in `conftest`, and be session-level

tests/test_explain_llm.py
```diff
@@ -0,0 +1,1340 @@
...
+# ---------------------------------------------------------------------------
+# Top-level output shape
+# ---------------------------------------------------------------------------
+
+
+def test_success_shape(tpch_graph: GraphMetadata, tpch_session: PyDoughSession):
+    """A valid expression returns the full success shape."""
+
+    def impl():
+        return nations.CALCULATE(key, name)
```
knassre-bodo: To make these tests easier to create/run, and potentially even parameterize, you could perhaps turn these into strings and have `_run` use `pydough.from_string`.

tests/test_explain_llm.py
```diff
@@ -0,0 +1,1340 @@
...
+    session: PyDoughSession,
+) -> dict:
+    """Qualify ``impl`` under ``graph`` and call ``explain_llm``."""
+    node: UnqualifiedNode = pydough.init_pydough_context(graph)(impl)()
+    return cast(dict, pydough.explain_llm(node, session=session))
+
+
+def _step(result: dict, order: int) -> dict:
+    """Return the step with the given 1-based order."""
+    return next(s for s in result["steps"] if s["order"] == order)
```
knassre-bodo: Let's name these functions a bit more descriptively, also add arguments/returns to the docstrings.

tests/test_explain_llm.py
```diff
@@ -0,0 +1,1340 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Structured error payload — error_type, details, hint
+# ---------------------------------------------------------------------------
+
+
+def test_error_payload_has_all_structured_fields(
+    tpch_graph: GraphMetadata, tpch_session: PyDoughSession
+):
```
knassre-bodo: For small payloads like this, what if we just had a parameterized test where the input is the PyDough code, output is the json payload itself, that way we can just compare against the entire output. We could cover most of these tests in this manner far more compactly, and test even more behavior that way than the manual asserts are handling.

knassre-bodo: A more extreme option, if we wish to go that way, is to have the refsols be stored in a json file that we append to. The keys of this refsol file would be the name of the tests which are parameterized alongside the PyDough code text.

The downside is that we are moreso testing the _exact_ output format, but the upside is that it becomes easier to mass-update tests just by pasting the new values into the output file.

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+    and the list of suggestions).  ``hint`` is the actionable guidance string
+    for the LLM — ``None`` when no specific hint is available.
+    """
+    from pydough.errors import (
+        PyDoughQDAGException,
+        PyDoughSessionException,
+        PyDoughSQLException,
+        PyDoughTypeException,
+        PyDoughUnqualifiedException,
+    )
```
knassre-bodo: Why are these imported dow here instead of at the top of the file?

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+        "schema": None,
+    }
+
+
+# ---------------------------------------------------------------------------
+# Step builders — one per QDAG node type
+# ---------------------------------------------------------------------------
+
+
+def _build_global_context_step(node: GlobalContext, order: int) -> dict:
```
knassre-bodo, line 337: This and several other docstrings should have descriptions formatted in the standard way we have been (`args`, `returns`)

pydough/exploration/explain_llm.py
knassre-bodo: ^ Let's add a pre-declaration near the top: `result: dict`

pydough/exploration/explain_llm.py
```diff
@@ -0,0 +1,1177 @@
...
+        # Override the generic classification with the specific type
+        result["error_type"] = "expression_not_collection"
+        result["hint"] = _HINT_MAP["expression_not_collection"]
+        return _render_md(result) if format == "md" else result
+
+    assert isinstance(qualified, PyDoughCollectionQDAG)
+
+    # --- build output --------------------------------------------------------
+    steps = _collect_steps(qualified)
+    schema = _build_schema(qualified)
```
knassre-bodo, line 1398: Type hints

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+                f"terms are available downstream."
+            )
+
+        case Where():
+            # Detect window functions (RANKING, PERCENTILE, etc.) in the
+            # condition.  Their `per=` partition argument is resolved to SQL
+            # PARTITION BY during compilation and is NOT stored on the
+            # WindowCall QDAG node, so it cannot be shown in the condition
+            # text above.  Alert the judge so it doesn't mis-read a
+            # per-partition rank as a global rank.
```
knassre-bodo: Why only for `WHERE`? Technically, calculate/orderby/topk can also contain window functions inside their expression arguments.

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+
+def _resolve_collection_arg(
+    arg: PyDoughCollectionQDAG,
+    parent: PyDoughCollectionQDAG | None,
+) -> PyDoughCollectionQDAG:
+    """
+    Resolves a collection argument to the underlying collection node.
+
+    Inside a ``CALCULATE``, aggregation arguments are represented as
+    ``ChildReferenceCollection`` nodes that point to the parent's child list
```
knassre-bodo, line 312: It's not just aggregations: it can also be singular sub-collections that are referenced to pull data into the current context (e.g. `nations.CALCULATE(nation_name=name, region_name=region.name)`

tests/test_explain_llm.py
knassre-bodo: An even more extreme version of this idea is making the JSON file dynamically updatable when doing `pdunit_update`, which I am also like.

The _most_ extreme version is to do the same idea, but to also put the PyDough code in the same JSON file, e.g.:

```json
{
  "test_name_1": {"pydough": "...", "explain_output": {...}},
  "test_name_2": {"pydough": "...", "explain_output": {...}},
   ...
}
```

Then the parameterized tests are just going through all the keys

pydough/exploration/_common.py
```diff
@@ -115,3 +156,630 @@ def qualify_safely(
...
+    Generated purely from the already-built ``steps`` dict —
+    no QDAG re-walking, no external calls.  Each clause is omitted when the
+    corresponding step type is absent.
+
+    Clause order:
+    1. Subject  — ``TableCollection`` / ``Cross`` / ``UserGeneratedCollection``
+    2. Filter   — all ``Where`` step conditions joined with ``" and "``
+    3. Partition — ``PartitionBy`` keys
+    4. Compute  — final ``Calculate`` step (refs + aggregations)
+    5. Limit/Order — ``TopK`` or ``OrderBy``
```
knassre-bodo, line 771: How does this order interact with far more complex queries that have multiple layers of partitioning / stepping back into the children? I'm struggling to visualize this (may help to do so with my testing suggestions).

knassre-bodo (commented):
This entire logic is quite thorough and interesting! I didn't go through every single detail of some of the middle functions, but I have left some comments on places where I think we can potentially iterate a bit further at the macro level.

My biggest wish after reading everything is something that I think would be tricky to conceptually figure out, but if you can do it would be amazing for extensibility in future: could we find a way to move some aspects of this, particularly stuff that is extremely specific to each type of QDAG node, into the QDAG APIs? Those classes use a LOT of ABC logic, with extensive class hierarchies, so perhaps there is a way to make this work by folding in different methods/templates to some of the abstract base classes, then having the explain_llm logic case on the object ancestry (e.g. calculate vs where vs topk vs orderby vs singular all inherit from `AugmentingChildOperator`, so we could have explain_llm case on whether it is an instance of that in order to resolve a lot of common logic, and a lot of other things inherit from `ChildAccess`).

If you look into this and it seems horrifically impractical, we can disregard for now, but if it is even somewhat viable I would encourage doing it. After all, we'll need to extend this all over again for `EXPLODE`, and I'd prefer if it was _literally_ impossible to miss adding any implementations because if we did, the ABC would fail due to un-implemented methods.

Some areas that I think are particularly ripe for moving into the ABCs:
- `describe_expression`
- most/all of the `_build_xxx_step` can just be made into a single abstract method that the classes implement
- Possibly `_render_step_body`?

Besides that, I think the overhaul to the testing approach is probably what I would consider the most. I think actually being able to see the output will help us tell if we are missing anything serious, or any glaring bugs jump out.

pydough/exploration/explain_llm.py
hadia206: Agree. Added a comment explaining the tradeoff. The fix would require structured attributes on the error classes (out of scope here),

pydough/exploration/explain_llm.py
hadia206: `details` is `{}` (an empty dict). When `_error_payload` receives a plain string (not an exception), there's no exception to classify, so there are no structured fields to extract (no wrong term name, no suggestions list, etc.). The "generic" error type is used and details is left empty.

pydough/exploration/explain_llm.py
hadia206: This is not part of the actual output. It's just to show what would the output in case of success/failure will look like.

pydough/exploration/explain_llm.py
hadia206: No reason, updated that.

pydough/exploration/_common.py
hadia206: Probably didn't encounter them while testing so didn't notice that. updated it.

pydough/exploration/_common.py
hadia206: updated

pydough/exploration/_common.py
hadia206: Added comment to clarify that.

hadia206:
@knassre-bodo  regarding your overall feedback.
Agreed, the ABC approach is the right long-term design, especially with EXPLODE on the horizon. 
I think `_build_xxx_step` to abstract method and `describe_expression` to `describe()` on expression nodes are both clearly viable. 

The `_render_step_body` case is trickier since I'd rather not put markdown rendering knowledge into QDAG, but we could at least make the step dict richer so rendering is purely structural.

Given the scope, I'd like to keep this PR as-is and do the ABC refactor as a follow-up since it would touch ~10+ QDAG files and I want give it time to work on. 

Agreed on the testing overhaul too; file snapshots for the MD output would make regressions obvious. Updated PR to do that.

tests/test_explain_llm_refsols/nations_iff_partition.md
```diff
@@ -0,0 +1,73 @@
...
+### Step 4 — Calculate
+
+Adds computed expressions to the collection.
+
+- Terms:
+  - `tier` → IFF(key > 15, 'high', 'low')
+
+### Step 5 — PartitionBy
+
+Partitions the collection by ['tier'].
```
knassre-bodo, line 39: We should include a mention of what the new virtual collection created by the partitioning is called (in this case, `g`)

tests/test_explain_llm.py
```diff
@@ -0,0 +1,557 @@
...
+            (
+                "nations_count_customers",
+                "result = nations.CALCULATE(n=COUNT(customers))",
+            ),
+            id="nations_count_customers",
+        ),
+        pytest.param(
+            ("nations_cross_regions", "result = nations.CROSS(regions)"),
+            id="nations_cross_regions",
+        ),
```
knassre-bodo, line 351: Test suggestions:
- both sides of the CROSS have multiple terms (`customers.WHERE(...).CROSS(suppliers.WHERE()).WHERE(...)`)
- CROSS is inside a CALCULATE (e.g. for each number from 1 to 10 in a range collection, count how many parts have that size)

tests/test_explain_llm.py
```diff
@@ -0,0 +1,557 @@
...
+                "result = customers.WHERE(market_segment == 'BUILDING').CALCULATE(name=name, n=COUNT(orders)).WHERE(RANKING(by=n.ASC()) == 1).CALCULATE(name)",
+            ),
+            id="customers_pre_post_calculate_where",
+        ),
+        pytest.param(
+            (
+                "customers_orders_ranking_where",
+                "result = customers.orders.WHERE(RANKING(by=total_price.DESC(), per='customers') == 1).CALCULATE(key=key)",
+            ),
+            id="customers_orders_ranking_where",
```
knassre-bodo, line 417: Test suggestions:
- What about `BEST`? Both with / without a `per=` anscestor

tests/test_explain_llm.py
```diff
@@ -0,0 +1,557 @@
...
+            ),
+            id="customers_orders_lines_partition",
+        ),
+        pytest.param(
+            (
+                "global_count_nations_where",
+                "result = CALCULATE(n=COUNT(nations.WHERE(key > 5)))",
+            ),
+            id="global_count_nations_where",
+        ),
```
knassre-bodo, line 479: Test Suggestions:
- Multiple aggregations on the same child collection
- Multiple aggregations on different children
- More complex children that have more steps to them (e.g. partitioning a partition, or the data being aggregated is a more complex PyDough logic that has its own children, etcetera)
- More stuff with user generated collections
- Having a partition then stepping back into the unpartitioned data (e.g. for window functions)

tests/test_explain_llm.py
```diff
@@ -0,0 +1,557 @@
...
+        `session`: the PyDough session used for graph context and qualification.
+
+    Returns:
+        The JSON-serialisable dict produced by ``pydough.explain_llm``.
+    """
+    node: UnqualifiedNode = pydough.from_string(code, session=session)
+    return cast(dict, pydough.explain_llm(node, session=session))
+
+
+def _run_explain_md_str(code: str, session: PyDoughSession) -> str:
```
knassre-bodo: These two functions are so tiny, and have so much in common, we might as well just inline them in `test_explain_llm_refsol_json`

tests/test_explain_llm.py
```diff
@@ -0,0 +1,557 @@
...
+        `session`: the PyDough session used for graph context and qualification.
+
+    Returns:
+        The markdown string produced by ``pydough.explain_llm(format="md")``.
+    """
+    node: UnqualifiedNode = pydough.from_string(code, session=session)
+    return cast(str, pydough.explain_llm(node, session=session, format="md"))
+
+
+def test_explain_llm_refsol_json(
```
knassre-bodo: We can probably just combine this function with the next one, having each test cover both the JSON file and the MD file

tests/test_explain_llm.py
knassre-bodo: A good example of what I mean about more complex children:
```
selected_customers = (
 customers
 .WHERE((nation.region.name == 'EUROPE') & HAS(orders.WHERE(YEAR(order_date) == 1994)))
)
result = TPCH.CALCULATE(n=COUNT(selected_customers))
```

This is a very basic sort of query that I'm unsure of how it renders under this model since 80% of the logic is just inside the argument to `COUNT(...)`

tests/test_explain_llm.py
```diff
@@ -0,0 +1,557 @@
...
+            ),
+            id="customers_where_orders_where",
+        ),
+        pytest.param(
+            (
+                "customers_singular_nation_name",
+                "result = customers.CALCULATE(nation_name=nation.SINGULAR().name)",
+            ),
+            id="customers_singular_nation_name",
+        ),
```
knassre-bodo, line 390: Test Suggestions:
- Accessing a child that is normally plural, but becomes singular (e.g. doing a filter with `RANKING`)

tests/test_explain_llm.py
hadia206: Good call about testing user generated collections. 
Done and done.

tests/test_explain_llm_refsols/cross_both_sides_filtered.md
```diff
@@ -0,0 +1,44 @@
+## Key Facts
+
+- **Source collection:** `suppliers`
+- **Limit:** none
+- **Data filters:** account_balance > 0 AND account_balance > 1000
+
+## Query Summary
+
+Pairs every 'customers' row with every 'suppliers' row, then subcollection filtered to rows where account_balance > 0 and account_balance > 1000.
```
knassre-bodo: This isn't fully accurate since the first filter is done before the pairing. It is fine in this context, but I wonder if there are other contexts where the summarization would be more drastically incorrect because of this.

tests/test_explain_llm_refsols/cross_both_sides_filtered.md
```diff
@@ -0,0 +1,44 @@
...
+- Left: `customers`
+- Right: `suppliers`
+
+> Each row now represents a unique combination of 'customers' × 'suppliers'. After CROSS, only 'suppliers' terms are directly accessible as expressions; 'customers' terms were available before the CROSS.
+
+### Step 3 — Where
+
+Filters rows to those matching the given conditions.
+
+- Condition: `account_balance > 0`
```
knassre-bodo: Yeah, here is where things start going wrong: it thinks the where is AFTER the cross. What other operations will it mess up like this?

tests/test_explain_llm_refsols/customers_orders_lines_partition.md
```diff
@@ -0,0 +1,65 @@
+## Key Facts
+
+- **Source collection:** `customers`
+- **Output collection:** `lines`
+- **Limit:** none
+- **Data filters:** market_segment == 'BUILDING'
+
+## Query Summary
+
+Accesses 'customers', filtered to rows where market_segment == 'BUILDING', partitioned by return_flag, selecting return_flag.
```
knassre-bodo: No mention of the fact that we are partitioning lineitems made by those customers.

tests/test_explain_llm_refsols/customers_orders_ranking_where.md
```diff
@@ -0,0 +1,51 @@
+## Key Facts
+
+- **Source collection:** `customers`
+- **Output collection:** `orders`
+- **Limit:** none
+- **Data filters:** RANKING(by=(total_price.DESC(na_pos='last')), levels=1) == 1
+
+## Query Summary
+
+Accesses 'customers', then subcollection filtered to rows where RANKING(by=(total_price.DESC(na_pos='last')), levels=1) == 1, selecting key.
```
knassre-bodo: Displaying it as `levels=1` seems wrong. Could we find a way to display it where it shows that it is getting the ranking per-customer? (the number 1 can still find its way in there as explaining it is the first ancestor of the current context, orders).

tests/test_explain_llm_refsols/customers_where_orders_partition.md
```diff
@@ -0,0 +1,76 @@
+## Key Facts
+
+- **Source collection:** `customers`
+- **Output collection:** `orders`
+- **Limit:** 1
+- **Data filters:** market_segment == 'BUILDING'
+
+## Query Summary
+
+Accesses 'customers', filtered to rows where market_segment == 'BUILDING', partitioned by order_status, selecting order_status and computing counting orders records per group as 'n', keeping the top 1 rows by n desc.
```
knassre-bodo: Again it feels like the step of customers->orders should be mentioned in the summary.

knassre-bodo: Also, should it perhaps say `n descending` instead of `n desc`?

tests/test_explain_llm_refsols/nations_count_range.md
```diff
@@ -0,0 +1,37 @@
+## Key Facts
+
+- **Source collection:** `nations`
+- **Limit:** none
+- **Data filters:** none
+
+## Query Summary
+
+Accesses 'nations', selecting name and computing counting r records as 'n'.
```
knassre-bodo, line 9: This does not seem to explain what `r` is or the range collection?

tests/test_explain_llm_refsols/nations_cross_regions.json
```diff
@@ -0,0 +1,65 @@
+{
+  "pydough": "result = nations.CROSS(regions)",
```
knassre-bodo, line 2: Wtb examples where there are multiple collections on each side? E.g. `a.b.CROSS(c.d)`? Or where each side has a `PARTITION` on it?

knassre-bodo (approved):
LGTM, my remaining comments can be adressed in followups if need be. My main lasting concern is how we handle explaining child contexts, since in some queries that could be where most of the stuff to explain will actually be.

tests/test_explain_llm_refsols/cross_both_sides_filtered.md
hadia206: Fixed. The issue was that `_collect_steps` was silently dropping the whole left-hand chain (including this filter) whenever `CROSS` was involved.

tests/test_explain_llm_refsols/cross_both_sides_filtered.md
hadia206: Same root cause as above

tests/test_explain_llm_refsols/customers_orders_lines_partition.md
hadia206: Added a `then navigates to 'orders' then 'lines'` clause to the summary generator so subcollection hops are not silently skipped.

tests/test_explain_llm_refsols/customers_orders_ranking_where.md
hadia206: Fixed, now it resolves `levels` back to the actual ancestor name by walking the QDAG

tests/test_explain_llm_refsols/customers_where_orders_partition.md
hadia206: Updated

tests/test_explain_llm_refsols/nations_count_range.md
hadia206: Added note about what is the range collection

tests/test_explain_llm_refsols/nations_cross_regions.json
hadia206: Added suggested tests and left notes about the current gaps. to fix in a followup as needed.

PR #513, opened 2026-04-30
PyDough DSL Benchmarker

john-sanchez31:
## Linked ticket

Closes #

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Docs / config

## What changed and why?
Adding the new benchmarker. This new class will run a list of questions and get metrics related to the performance of pydough. All this data can be used for detecting regressions and opportunities of optimizations. This generates a new file with all collected information and adds a new row on the `benchmark_metrics.csv` file.

## How I tested this?
This new class has been tested locally using the command `python executer.py`. There are not tests on our test suite for this new feature that could be discussed later.

## Notes for reviewers

The workflow spins up the Postgres Docker container as a service using the image `bodoai1/pydough-benchmarker:latest`, installs the project dependencies using uv, and runs executer.py to execute the benchmark. Once the benchmark finishes and the metrics files are generated, it commits those files to a new branch and automatically opens a PR against main for review.
It triggers automatically on every new GitHub release published, and can also be run manually at any time from the Actions tab.


.github/workflows/pr_testing.yml
```diff
@@ -70,6 +70,11 @@ on:
         type: boolean
         required: false
         default: false
+      run-benchmark:
```
john-sanchez31, line 79: This is a temporary flag so I can trigger the workflow without having to merge to main. The idea would be that the 2 main ways to execute the benchmark are through github actions tab and on published release

.github/workflows/pr_testing.yml
```diff
@@ -307,3 +312,14 @@ jobs:
       python-versions: ${{ github.event_name == 'workflow_dispatch'
                       && needs.get-py-ver-matrix.outputs.matrix
                       || '["3.10", "3.11", "3.12", "3.13"]' }}
+
```
john-sanchez31, line 339: This can be deleted before merging to main

benchmark/questions.csv
```diff
@@ -0,0 +1,975 @@
+benchmark,question_id,pydough,sql
```
john-sanchez31: I've been thinking about adding the questions itself here and add it to the generated files. Do you think is something that would help us or not really?

benchmark/questions.csv
```diff
@@ -0,0 +1,975 @@
...
+                ('13', '31', '23', '29', '30', '18', '17')
+      )
+      AND NOT EXISTS (
+          SELECT 1
+          FROM tpch.orders o
+          WHERE o.o_custkey = c.c_custkey
+      )
+) AS custsale
+GROUP BY cntrycode
+ORDER BY cntrycode;"
```
john-sanchez31: Reminder to fix this

benchmark/README.md
```diff
@@ -0,0 +1,124 @@
+# PyDough Performance Benchmarker
```
john-sanchez31, line 1: Pretty much is the same we have in the design document. Let me know if I need to add something else

john-sanchez31 (commented):
Also take a look of the generated PR for the last run of the benchmarker: https://github.com/bodo-ai/PyDough/pull/517

.github/workflows/pr_testing.yml
hadia206: Why? I think it's good to have an option to run benchmark on demand rather than rely on releases only.

benchmark/benchmark_metadata.json
```diff
@@ -0,0 +1,870 @@
+[
```
hadia206, line 1: I'm assuming this is the same as sample one in tests for now, right?

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+        exec_time (float): Execution duration of the ground truth SQL in seconds.
+        pydough_exec_time (float): Execution duration of the PyDough SQL in seconds.
+        exec_plan (str): The database execution plan for the ground truth query.
+        pydough_exec_plan (str): The database execution plan for the PyDough query.
+        status (str): The outcome of the comparison (e.g., 'SUCCESS', 'FAILURE', 'TIMEOUT').
+    """
+
+    @property
+    def connection(self) -> Connection:
+        """Independent connection to the database"""
```
hadia206: What does independent connection mean in this context?

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+        run_id = os.getenv("RUN_ID", "manual")
+
+        today = date.today().strftime("%Y%m%d")
+
+        if run_type == "R":
+            # e.g. 20260508_R_v1.2.0
+            return f"{today}_R_{run_id}"
+        else:
+            # e.g. 20260508_M_25526200301
+            return f"{today}_M_{run_id}"
```
hadia206, line 183: Let's add a comment clarifying what does `R` and `M` refer to.

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+        """
+        full_path_file: str = f"{self.questions_path}/{self.filename}.csv"
+        if os.path.isfile(full_path_file):
+            return pd.read_csv(full_path_file)
+
+        raise ValueError(f"File {full_path_file} doesn't exists")
+
+    def execute_pydough(self, pydough_str: str) -> tuple[DataFrame, str]:
+        """
+        Executes the given pyodugh code using `from_string` api
```
hadia206: docstring args/return

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+        if run_type == "R":
+            # e.g. 20260508_R_v1.2.0
+            return f"{today}_R_{run_id}"
+        else:
+            # e.g. 20260508_M_25526200301
+            return f"{today}_M_{run_id}"
+
+    def get_questions_df(self) -> DataFrame:
+        """
+        Returns the questions full path
```
hadia206: This is no returning the full path.
It returns the questions as dataframe

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+    @property
+    def run_label(self) -> str:
+        """
+        Label for the benchmark run
+        """
+        return self._run_label
+
+    def get_run_label(self) -> str:
+        """
+        Generate the label for this run
```
hadia206: docstring return

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+            result: DataFrame = pydough.to_df(pydough_code)
+            return result, pydough.to_sql(pydough_code)
+        except Exception as e:
+            raise ValueError(
+                f"An error occurred while generating the PyDough code: {e}"
+            )
+
+    def execute_sql(self, query: str) -> DataFrame:
+        """
+        Executes the given sql in the established connection
```
hadia206, line 225: docstring args/return/exception

hadia206: Please do that for all methods

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+                return "\n".join([row[0] for row in plan])
+
+        except (Exception, psycopg2.DatabaseError) as error:
+            raise ValueError(
+                f"An error occurred while generating the execution plan: {error}"
+            )
+
+    def generate_metrics_file(self) -> None:
+        """
+        Generates an csv file with all the results using
```
hadia206:
```suggestion
        Generates a csv file with all the results using
```

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+
+            execution_plan: str = self.generate_execute_plan(postgres_sql)
+
+            # PyDough metrics
+
+            start_pyd = time.perf_counter()
+            pydough_result, pydough_sql = self.execute_pydough(row["pydough"])
+            elapsed_pydough_time = time.perf_counter() - start_pyd
+
+            pyd_execution_plan: str = self.generate_execute_plan(pydough_sql)
```
hadia206: A wrong question will crash the entire benchmark run. The exceptions raised by `execute_sql`, `generate_execute_plan`, and `execute_pydough` should be captures and recorded
Wrap the entire code with try/except and record the error status. This allows moving to following question without crashing

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+
+            # Check for correctness
+            row_metrics: dict[str, str | int | float]
+            try:
+                for col_name in pydough_result.columns:
+                    pydough_result[col_name], ground_truth_result[col_name] = (
```
hadia206: `ground_truth_result` is indexed using PyDough's column name, is it guaranteed that it will have that name too?
If not, we need to check before indexing and record error (column not found or something)

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+                        harmonize_types(
+                            pydough_result[col_name], ground_truth_result[col_name]
+                        )
+                    )
+
+                pd.testing.assert_frame_equal(
+                    pydough_result, ground_truth_result, rtol=1e-6, atol=1e-9
```
hadia206: What if columns returned are in different orders?
Let's add `check_like=True` See [here](https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_frame_equal.html)
```
check_likebool, default False
If True, ignore the order of index & columns.
```

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+class Connection:
+    def __init__(
+        self,
+        db_name: str,
+        user: str,
+        password: str,
+        host: str = "localhost",
+        port: int = 5432,
+    ):
+        self._conn: psycopg2.extensions.connection = psycopg2.connect(
```
hadia206, line 33: connection is opened here but never closes.
Probably should close in end of `measure`

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+                    "benchmark": row["benchmark"],
+                    "ground_truth": postgres_sql,
+                    "pydough_code": row["pydough"],
+                    "pydough_sql": pydough_sql,
+                    "exec_time": elapsed_exec_time,
+                    "pydough_exec_time": elapsed_pydough_time,
+                    "exec_plan": execution_plan,
+                    "pydough_exec_plan": pyd_execution_plan,
+                    "status": f"FAILURE: Wrong answer - {e}",
+                }
```
hadia206, line 515: Add finally block to close the connection

benchmark/executer.py
```diff
@@ -0,0 +1,65 @@
...
+        cur = conn.cursor()
+        cur.execute("SELECT COUNT(*) FROM tpch.lineitem;")
+        row = cur.fetchone()
+        if row and row[0] == 59986052:
+            conn.close()
+            break
+        else:
+            print(f"Waiting {i + 1}/600 seconds for data to be load...")
+            time.sleep(1)
+        break
```
hadia206: why this `break`?
It's always executed. So the loop exits the first time regardless of `tpch.lineitem` has been generated.

benchmark/executer.py
```diff
@@ -0,0 +1,65 @@
...
+database: str = "benchmarker"
+
+# Check for the necessary environment variables before running the benchmark
+if username is None or password is None:
+    raise ValueError(
+        "Environment variables POSTGRES_USER and POSTGRES_PASSWORD must be set."
+    )
+
+# Wait for Postgres to be ready for 10 minutes max
+for i in range(600):
```
hadia206: This whole block should be in `if __name__ == "__main__"`

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+    the metrics generating the metrics file with the results.
+    """
+
+    def __init__(
+        self,
+        postgres_conn: Connection,
+        questions_path: str = "./",
+        filename: str = "questions",
+        export_metrics: bool = False,
+        metadata_path: str = "./benchmark_metadata",
```
hadia206, line 54: missing `retries` from the README

benchmark/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+```python
+from benchmarker import Benchmarker
+
+benchmarker = Benchmarker(
+    postgres_conn=conn,           # psycopg2 (or compatible) connection
+    questions_path="./",          # directory containing the questions file
+    filename="questions",         # questions filename (without .csv extension)
+    export_metrics=False,         # write metrics CSV on completion
+    retries=3                     # retry attempts per query on failure
```
hadia206: This is not in the code. Either update the code or delete from here.

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+        for index, row in questions.iterrows():
+            print(f"Executing question {index + 1}/{len(questions)}")
+
+            if len(current_graph) == 0 or current_graph != row["benchmark"]:
+                current_graph = row["benchmark"]
+                pydough.active_session.load_metadata_graph(
+                    self.metadata_path, current_graph
+                )
+
+            postgres_sql: str = row["sql"]
```
hadia206: In README, it's named `postgres_sql` so change either here or README for consistency.
I prefer change here to be more descriptive

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+            start = time.perf_counter()
+            ground_truth_result: DataFrame = self.execute_sql(postgres_sql)
+            elapsed_exec_time = time.perf_counter() - start
+
+            execution_plan: str = self.generate_execute_plan(postgres_sql)
+
+            # PyDough metrics
+
+            start_pyd = time.perf_counter()
+            pydough_result, pydough_sql = self.execute_pydough(row["pydough"])
```
hadia206: Same.
In README, it's named `pydough_code` so change either here or README for consistency.
I prefer change here to be more descriptive

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+        self._run_label = self.get_run_label()
+
+    METRICS_COLUMNS: dict[str, str] = {
+        "question_id": "int",
+        "benchmark": "str",
+        "ground_truth": "str",
+        "pydough_code": "str",
+        "pydough_sql": "str",
+        "exec_time": "float",
+        "pydough_exec_time": "float",
```
hadia206, line 75: Same. In README these are named `exec_time_ms`, `pydough_exec_time_ms`
and I believe these are seconds not milliseconds, right?
Update either here or README to be consistent

hadia206: Also, missing `pydough_time_ms` Not sure what's the difference between that and execution time.

benchmark/benchmarker.py
```diff
@@ -0,0 +1,351 @@
...
+        "exec_plan": "str",
+        "pydough_exec_plan": "str",
+        "status": "str",
```
hadia206, line 78: Also, where is `pydough_time_ms` (Time to generate SQL via PyDough)?

benchmark/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+Each question produces one row in the metrics DataFrame, written to a CSV named `{date}_{pydough_version}.csv`.
+
+| Column | Type | Description |
+|---|---|---|
+| `question_id` | `int` | Unique identifier for the question |
+| `benchmark` | `str` | Name of the benchmark suite |
+| `ground_truth` | `str` | Reference SQL query |
+| `pydough_sql` | `str` | SQL generated by PyDough |
+| `pydough_time_ms` | `float` | Time to generate SQL via PyDough (ms) |
```
hadia206: Missing from the code.

benchmark/README.md
```diff
@@ -0,0 +1,124 @@
...
+Execution times are measured using `time.perf_counter()` for high-resolution timing.
+
+---
+
+## API Reference
+
+### `measure() -> None`
+Main entry point. Orchestrates the full benchmark run — loads questions, executes both SQL paths, validates results, and collects all metrics.
+
+### `execute_pydough(pydough_code: str) -> DataFrame`
```
hadia206:
```suggestion
### `execute_pydough(pydough_code: str) -> tuple[DataFrame, str]`
```

benchmark/README.md
```diff
@@ -0,0 +1,124 @@
...
+Executes the given PyDough DSL code using the `from_string` API and returns the result as a DataFrame.
```
hadia206: Update description here to reflect code changes.

benchmark/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+### `execute_sql(query: str) -> DataFrame`
+Executes a raw SQL query against the configured PostgreSQL connection and returns the result as a DataFrame.
+
+### `generate_execute_plan(query: str) -> str`
+Runs `EXPLAIN` (or equivalent) on the given query and returns the execution plan as a string.
+
+### `query_validation(pydough_result: DataFrame, ground_truth_result: DataFrame) -> bool`
+Compares the PyDough result against the ground truth. Supports complex comparison logic to account for ordering and floating-point tolerances.
```
hadia206: This one is not in the code as its own function. 
It's better to add it in the code so that if we want to change how we do validation later own, it's easier to spot.

hadia206 (commented):
Good job, John!
Please see my comments below. Mostly about handling what happens if things go wrong and matching README with code.

benchmark/benchmarker.py
john-sanchez31: deleted from the README file

benchmark/benchmarker.py
john-sanchez31: The idea was to add how much time takes to generate the pydough code, but that doesn't really provide any valuable insights, so updating the readme

benchmark/benchmarker.py
john-sanchez31: Just a single database connection for one specific engine

benchmark/benchmarker.py
john-sanchez31: I added some assertions before this loop

benchmark/executer.py
john-sanchez31: You're right, I forgot to delete it

benchmark/README.md
john-sanchez31: Good catch!

benchmark/executor.py
```diff
@@ -0,0 +1,65 @@
+"""
```
knassre-bodo, line 1: I think it should be spelled `executor`

benchmark/questions.csv
```diff
@@ -0,0 +1,1135 @@
...
+            SUM_DISC_PRICE=SUM(lines.extended_price * (1 - lines.discount)),
+            SUM_CHARGE=SUM(
+                lines.extended_price * (1 - lines.discount) * (1 + lines.tax)
+            ),
+            AVG_QTY=AVG(lines.quantity),
+            AVG_PRICE=AVG(lines.extended_price),
+            AVG_DISC=AVG(lines.discount),
+            COUNT_ORDER=COUNT(lines),
+        )
+        .ORDER_BY(L_RETURNFLAG.ASC(), L_LINESTATUS.ASC())
```
knassre-bodo, line 24: Probably not something to be addressed right now, but I believe in our original discussions about the benchmarker we discussed having the LLM generate the PyDough code? I forget, did we decide against that at some point?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,101 @@
...
+          gh pr create \
+            --base main \
+            --head "$BRANCH" \
+            --title "Benchmark metrics update — $(date +'%Y-%m-%d')" \
+            --label "benchmark" \
+            --label "automated" \
+            --body "Automated benchmark run completed successfully.
+
+            **Triggered by:** \`${{ github.event_name }}\`
+            **Run:** [${{ github.run_id }}](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})
```
knassre-bodo: Let's ensure the commits/PR generated indicate:
- Whether it was generated from a release or manually
- If on a release, which release
- If manually, on which commit of which branch

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,101 @@
...
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          BRANCH="benchmark/metrics-$(date +'%Y-%m-%d-%H%M%S')"
+
+          git config user.name  "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+
+          git checkout -b "$BRANCH"
+          git add benchmark/metrics/*.csv benchmark/benchmark_metrics.csv
```
knassre-bodo: The PR that was opened included all of these changes AND the metric file. Will this always happen? If so that is potentially problematic if people are doing active development on a branch while they run the workflow, since the PR will contain all of their active development changes.

benchmark/README.md
```diff
@@ -0,0 +1,123 @@
...
+Exports the collected metrics to a CSV file named `{date.today()}_{pydough_version}.csv`.
+
+---
+
+## Supported Benchmarks
+
+| Benchmark | Queries | Status |
+|---|---|---|
+| TPC-H (Scale Factor 10) | 22 | Available |
+| TPC-DS (Scale Factor 10) | 99 | Added incrementally |
```
knassre-bodo: Are we sure we want to do SF-10 for DS? That database can get pretty big. You may want to experiment to see which one is more viable.

benchmark/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+A benchmarking framework for measuring and tracking the performance of the PyDough DSL across standard industry benchmarks (TPC-H, TPC-DS).
+
+## Overview
+
+The Benchmarker collects performance data by executing PyDough DSL code against benchmark queries, validating results against ground truth SQL, and recording execution metrics. This enables regression tracking across releases and surfaces opportunities for new optimizations.
+
+## Automated Runs
+
+The benchmarker runs automatically on every PyDough DSL release via GitHub Actions. Results are committed as a new PR containing a timestamped CSV file. The workflow can also be triggered manually.
```
knassre-bodo: Let's have a more in-depth explanation of the following:
- The CSV file is in the format described in `## Metrics`
- The file is generated from the workflow run on release, or the manual one
- The exact naming scheme of the generated file in those two versions

Also, perhaps we should have a comparator script that takes in 2x of these benchmark run file names and outputs a comparison file:
- How many tests were right/wrong before/after (list all tests that went wrong->right or right->wrong)
- The total time performance only looking at queries that were correct in both versions
- For all queries that were correct in both versions, a breakdown of the absolute & relative % time improvement/regression (can sort this from best-to-worst)

This can be dumped into stdout or into a file for analysis.

knassre-bodo: For instance, suppose file R1 had the following data:
- q1: correct, pydough_time=10
- q2: correct, pydough_time=15
- q3: correct, pydough_time=20
- q4: incorrect, pydough_time=25
- q5: incorrect, pydough_time=30
- q6: correct, pydough_time=42

And suppose file R2 had the following data:
- q1: incorrect, pydough_time=30
- q2: correct, pydough_time=10
- q3: correct, pydough_time=23
- q4: correct, pydough_time=5
- q5: incorrect, pydough_time=25
- q6: correct, pydough_time=30

What the comparison file would tell you (formatted a bit nicer, and probably going to decimals):
- R1: 4/6 correct
- R2: 4/6 correct
- Incorrect->Correct: q6
- Correct->Incorrect: q1
- Total Correct Time: 77->63 (-18%)
- Time breakdown:
  - q2: 15->10 (-33%)
  - q6: 42->30 (-29%)
  - q3: 20->23 (+15%)


Could also do the same analysis, but instead of comparing the absolute times, we compare the differences relative to the refsol times (e.g. did R2 get faster/slower relative to the refsol than R1). Perhaps both of those time analyses could be in the same file, in different sections?

knassre-bodo (commented):
@john-sanchez31 FANTASTIC job accomplishing this! There are a few things to tinker with, but this is definitely a great tool.

benchmark/questions.csv
john-sanchez31: Yes, at some point the original discussed task was divided into 2 different task. For the LLM team I created a script for data generation and for the DSL I'm working on the benchmarker. We could discuss about adding a LLM generation for the benchmarker for a followup task.

benchmark/README.md
john-sanchez31: Well I already have both TPCH and TPCDS SF-10 on the docker image. Is it worth the experimentation at this point?

benchmark/README.md
john-sanchez31: Is this still applicable taking in count that for instance every query is written manually and will be correct?

benchmark/benchmark_metadata.json
hadia206: ^ is that assumption true?

benchmark/benchmarker.py
hadia206: Let's clarify that in the docstring too

benchmark/benchmarker.py
hadia206: missing the finally block^

benchmark/benchmarker.py
```diff
@@ -0,0 +1,543 @@
...
+
+            self.question_metrics = pd.concat(
+                [self.question_metrics, pd.DataFrame([row_metrics])], ignore_index=True
+            )
+
+        if self._export_metrics:
+            self.generate_metrics_file()
+            self.generate_benchmark_metrics((time.perf_counter() - start_measure) / 60)
+
+        # clossing the connection
```
hadia206:
```suggestion
        # closing the connection
```

benchmark/questions.csv
hadia206: Might be useful to add a ticket and put its number here so we don't forget

benchmark/README.md
hadia206: Do we need to update the status here? like add 22/99 or something

hadia206 (commented):
Overall looks good to me. Please see comments below

.github/workflows/pr_testing.yml
knassre-bodo: Then it needn't be temporary, we can just keep this here.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,129 @@
...
+            --label "automated" \
+            --body "Automated benchmark run completed successfully.
+
+          **Generated from:** $SOURCE_TYPE  
+          **Source:** $SOURCE_DESC  
+
+          **Triggered by:** \`${EVENT}\`  
+          **Workflow run:** [${{ github.run_id }}](${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }})
+
+          Please review the updated metrics files before merging."
```
knassre-bodo: Newline at bototm

benchmark/benchmarker.py
knassre-bodo: I thought that these connection objects now auto-close when garbage collected?

benchmark/benchmarker.py
```diff
@@ -0,0 +1,543 @@
...
+        """
+
+        if not Path("benchmark/metrics/").exists():
+            Path("benchmark/metrics").mkdir()
+
+        self.question_metrics.to_csv(f"benchmark/metrics/{self.run_label}.csv")
+
+    def generate_benchmark_metrics(self, total_time_min: float) -> None:
+        """
+        Generates general metrics for the benchmark (AVG, Desviation, MIN, MAX)
```
knassre-bodo:
```suggestion
        Generates general metrics for the benchmark (average, standard deviation, min, and max)
```

PS: not sure why std is here? What information to we get from measuring the standard deviation from different questions in the same run, considering that these questions will take different amounts of time to run anyways based on what they are asking?

knassre-bodo (approved):
A few minor final comments, but I'm eager to see this in action once the PR + automatic branches are published, and we later have a release generate an auto-comparison :)

benchmark/benchmark_metadata.json
john-sanchez31: Yes, nothing changed but since I wanted to add also TPCDS I had to create a new file

benchmark/benchmarker.py
john-sanchez31: You're right. Deleting std from general metrics

benchmark/benchmarker.py
john-sanchez31: Not sure if I can add the finally block. The try-exception is inside the for loop which means that after the first query the connection will be close. Also if something fails the benchmark continues until the end of the questions adding the error on the metrics.

benchmark/questions.csv
john-sanchez31: Done: #534

PR #523, opened 2026-05-26
Add support for LISTOF function, array return, and EXPLODE operator in PyDough

### LINKED ISSUE #521: Add LISTOF function to aggregate data into arrays
Goal: add a new PyDough function aggregates `LISTOF` which aggregates the grouped data into an array, i.e. the `ARRAY_AGG` function from SQL. Not every dialect can support this feature (e.g. ANSI, SQLITE), so testing will need to include ways to exclude specific dialects from sql+e2e testing.

### LINKED ISSUE #522: Add EXPLODE operator to PyDough
Add a new operator EXPLODE in the PyDough DSL syntax emulating the `LATERAL FLATTEN` and `SPLIT_TO_ARRAY` functionalities from dialects such as Snowflake. Like #521, this functionality will only be allowed for certain dialects.

Suppose we have a collection `tbl` that looks like this:
```
    A  B  C
0 | 1  A  [1] 
1 | 2  B  []
2 | 3  A  [2, 3]
3 | 4  A  [4, 5]
4 | 5  B  [6, 7, 8]
5 | 6  A  [9]
```

The PyDough code would look something like:
```py
tbl.EXPLODE(C, index_name="C1", value_name="C2", version="array", keep=True, filtering=True, is_distinct=True)
```

What the various (sometimes optional) arguments mean:
- Explode `tbl` using column `C`
- `index_name="C1"`: create a new column C1 in the output, indicating the index within each list that the elements belonged to
- `value_name="C2"`: create a new column C2 in the output, which has the elements from C
- `version="array"`: Assume column C is an array (alternative is a string that gets split on a delimiteter)
- `keep=True`: Keep column C in the output answer
- `filtering=True`: Not every row in tbl will be kept in the final answer
- `is_distinct=True`: every element inside each list is unique

The resulting output collection:
```
    A  B  C          C1  C2
0 | 1  A  [1]         0   1
2 | 3  A  [2, 3]      0   2
2 | 3  A  [2, 3]      1   3  
3 | 4  A  [4, 5]      0   4  
3 | 4  A  [4, 5]      1   5 
4 | 5  B  [6, 7, 8]   0   6 
4 | 5  B  [6, 7, 8]   1   7  
4 | 5  B  [6, 7, 8]   2   8  
5 | 6  A  [9]         0   9  
```

If the original uniqueness information about `tbl` was `["A"]`, then the new uniqueness would be: `[["A", "C1"], ["A", "C2"]]`.

For each unique value of `A`, it will get duplicated but the duplicates are uniquely identifiable by the index within the list (`C1`)
Alternatively, can uniquely identify them by the elements IN the list since we know `is_distinct=True`

What the SQL for this looks like (only valid in certain dialects, this example is Snowflake; postgres/trino should also work):
```sql
SELECT T.A, T.B, T.C, L.index AS C1, L.value AS C2
FROM T
LATERAL FLATTEN(T.C) AS L
```

The main variant is the string-based version (version="string", and a new argument `delimeter=...`), where each row of `C` is something like `"I love you"` getting split on `" "` to form 3 rows `"I"`, `"love"`, and `"you"`.

The exact argument rules:
- Positional argument: required, this is the column that will be exploded
- `index_name`: required, unless `is_distinct=True`. If not provided, the enumeration column will not be included in the answer
- `value_name`: required
- `version`: optional, must either be `"array"` or `"string"` to indicate which version is used. Default is `"array"`.
- `keep`: optional, default is `True`, keeps the positional argument in the output answer.
- `filtering`: optional, default is `False`, indicates whether every original row of `tbl` is preserved at least once in the final answer
- `is_distinct`: optional, default is `False`, indicates whether every element within the lists/strings is unique, meaning that the enumeration column is not required to uniquely identify every row in the final output.
- `delimiter`: required if `version="string"`, prohibited otherwise

Note: relies on #521 for ease of testing.

knassre-bodo:
## Linked ticket

Closes #521
Closes #522

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Docs / config

## What changed and why?

Adds the `LISTOF` function to PyDough to aggregate data into arrays, and the `EXPLODE` operator to flatten such arrays into multiple rows of table data. Also added support for DataFrame collections containing arrays, which become array literals in the SQL.
- All of these features are not supported for ANSI or SQLITE because those dialects do not have those capabilities.
- Array literals are not supported for Snowflake because the dialect does not have that capability.
- EXPLODE is not yet supported currently for MySQL or Oracle due to additional complications with those dialects.

## How I tested this?

- Multiple new tests on TPCH data with the name prefix `array_data_` to test the LISTOF and array literal features.
- Multiple new tests on TPCH data with the name prefix `explode_` to test different variations of the EXPLODE operator, both on arrays and strings. Variations include:
  - Regular `collection.EXPLODE(data)`
  - Inside of a child context `collection.CALCULATE(x=COUNT(EXPLODE(data)))`
  - Exploding array data vs string data
  - Different cardinality/distinctness information about the exploded data
  - Exploded data being used for grouping and for window operations
  - Nested explosions (e.g. splitting an already split string with `collection.EXPLODE(...).EXPLODE(...)`)
  - Splitting strings on an empty delimiter
  - DataFrame tables with varying data-types, empty arrays, nulls
- Error tests for incorrect uses of the EXPLODE operator (e.g. bad arguments)

## Notes for reviewers

- A few changes to how the database connections are handled to account for difference in how the connectors return arrays.
- Several of the tests have dialect-specific skip marks. Some of these are truly unsupported features, and some are just _for now_ unsupported.


pydough/unqualified/unqualified_node.py
```diff
@@ -792,10 +878,30 @@ class UnqualifiedGeneratedCollection(UnqualifiedNode):
     def __init__(self, user_collection: PyDoughUserGeneratedCollection):
         self._parcel: tuple[PyDoughUserGeneratedCollection] = (user_collection,)
 
-    @property
-    def user_collection(self) -> PyDoughUserGeneratedCollection:
-        """The wrapped user-generated collection."""
-        return self._parcel[0]
```
knassre-bodo, line 798: This is not a thing that should be allowed with unqualified nodes, since any method or property name you add becomes something that can no longer be used as a PyDough term name.

documentation/dsl.md
```diff
@@ -1536,6 +1537,119 @@ People.CALCULATE(Packages=COUNT(People.packages)).CROSS(Packages)
...
+- `data` (required): the expression from the current context being exploded (either an array or string). This expression cannot reference any sub-collections of the current context.
+- `name` (required): the name of the collection created from the explosion operation (similar to `PARTITION`).
+- `value_name` (required): a string literal declaring the name of the new column that will be used to store the exploded data.
+- `index_name` (optional): a string literal declaring the name of the new column that will be used to store the indices of the exploded data.  If not provided, this column is not generated. The `index_name` is required if `is_distinct` is False. The indices are 0-indexed.
+- `version` (optional, default=`"array"`): either `"array"` or `"string"`, stating whether the data to explode is an array being flattened or a string being split on a delimiter.
```
john-sanchez31, line 1548: Is this argument really needed? Can't we just check the type of data?

documentation/dsl.md
```diff
@@ -1536,6 +1537,119 @@ People.CALCULATE(Packages=COUNT(People.packages)).CROSS(Packages)
...
+- `delimiter` (optional): a string literal indicating the delimiter that should be used to split up the string if `version="string"`. If `delimiter` is an empty string, the string will be split into individual characters.
+- `filtering` (optional, default=`True`): `True` if it is possible for not every row in the original collection to be preserved in the exploded sub-collection (i.e. if one of the arrays is empty), and `False` otherwise.
```
john-sanchez31, line 1550: I find this one kind of confusing. Will this filter on certain conditions or will be more like filtering Empty/None values? Assuming the second one, the description could be something as follow:
```suggestion
`filtering_empty` (optional, default=`True`): `True` will filter all empty or None values from the original collection to the exploded sub-collection, and `False` otherwise.
```

documentation/dsl.md
```diff
@@ -1536,6 +1537,119 @@ People.CALCULATE(Packages=COUNT(People.packages)).CROSS(Packages)
...
+- `is_distinct` (optional, default=`False`): `True` if each row of the exploded data is unique within the set of all other values from that same original row of unexploded data, and `False` otherwise. An `index_name` can only be omitted if `is_distinct` is `True`.
+
+> [!IMPORTANT]
+> This feature is only supported in certain dialects. It is currently supported for Snowflake, DataBricks, DuckDB, Postgres and Trino.
+
+**Good Example #1**: List out every phone number had by every person (assuming `phone_numbers` is an array of strings).
```
john-sanchez31, line 1556: For the good examples can we add how to result would look like? I will make it easier for the user to understand the operator.

documentation/dsl.md
```diff
@@ -1605,12 +1719,16 @@ The supported PyDough types for `dataframe_collection` are:
...
-- `Datetype`: date and datetime.
+- `DateType`: date and datetime.
+- `ArrayType`: arrays of data.
 - `UnknownType`: used for all `None` columns.
 
 Note: MySQL by default does not support infinity values. When PyDough detects 
 infinity value with `DatabaseDiatect.MYSQL` an error will be raised.
 
+> [!IMPORTANT]
+> `ArrayType` is only supported for certain dialects: Trino, Postgres, DuckDB, Databricks.
```
john-sanchez31, line 1758: What about Snowflake? I saw it among the supported dialects for the `EXPLODE` operator

documentation/functions.md
```diff
@@ -1089,6 +1090,23 @@ Parts.CALCULATE(std = STD(supply_records.supply_cost))
...
+### LISTOF
+
+The `LISTOF` function collects a set of values into an array.
+
+> [!IMPORTANT]
+> This function is only supported in certain dialects. It is currently supported for Snowflake, DataBricks, DuckDB, Postgres and Trino.
+
+```py
+# For each region, list the names of all nations inside that region
+Regions.CALCULATE(region_name=name, nation_names=LISTOF(nations.name))
```
john-sanchez31, line 1104: Can we add how the result for each of the good example would look like?

pydough/conversion/agg_removal.py
```diff
@@ -218,6 +219,29 @@ def deduce_join_uniqueness(
     return result
 
 
+def deduce_explode_uniqueness(
+    unique_terms: set[frozenset[str]], explode: Explode
+) -> set[frozenset[str]]:
+    """
+    Helper function to transforms the uniqueness sets after an Explode
+    operation duplicates rows.
```
john-sanchez31, line 227: Lets also add the Args/Returns on the docstring

pydough/conversion/relational_converter.py
```diff
@@ -1402,6 +1404,39 @@ def translate_child_pullup(self, node: HybridChildPullUp) -> TranslationOutput:
         # expressions mapping
         return TranslationOutput(child_result.relational_node, new_expressions)
 
+    def translate_explode(
+        self, operation: HybridExplode, context: TranslationOutput
+    ) -> TranslationOutput:
+        """
+        TODO
```
john-sanchez31: Reminder

pydough/database_connectors/builtin_databases.py
```diff
@@ -528,11 +528,11 @@ def load_duckdb_connection(**kwargs) -> DatabaseConnection:
         )
 
     if connection := kwargs.pop("connection", None):
-        return DatabaseConnection(connection)
+        return DatabaseConnection(connection, DatabaseDialect.DUCKDB)
 
     database = kwargs.pop("database", ":memory:")
     connection = duckdb.connect(database=database, **kwargs)
-    return DatabaseConnection(connection)
+    return DatabaseConnection(connection, DatabaseDialect.DUCKDB)
```
john-sanchez31, line 535: Why is this change?

pydough/database_connectors/database_connector.py
```diff
@@ -57,17 +60,45 @@ def execute_query_df(self, sql: str) -> pd.DataFrame:
...
+                    # into Python types.
+                    semi_structured_cols.update(
+                        {
+                            idx
+                            for idx, desc in enumerate(self.cursor.description)
+                            if desc[1] in (5, 9, 10)
+                        }
+                    )
+                case DatabaseDialect.MYSQL:
+                    # Snowflake returns JSON data (type 245) as strings, so we
```
john-sanchez31:
```suggestion
                    # MySQL returns JSON data (type 245) as strings, so we
```

pydough/sqlglot/transform_bindings/duckdb_transform_bindings.py
```diff
@@ -120,6 +128,105 @@ def convert_rpad(
...
+        self,
+        input_expr: SQLGlotExpression,
+        explode_expr: SQLGlotExpression,
+        explode_spec: ExplodeSpec,
+        exprs: list[SQLGlotExpression],
+        val_index: int | None,
+        idx_index: int | None,
+        lateral_alias: str,
+        subquery_alias: str,
+    ) -> SQLGlotExpression:
```
john-sanchez31, line 151: Can we add a docstring describing this implementation? What would be the final SQL for every dialect ?

pydough/sqlglot/transform_bindings/oracle_transform_bindings.py
```diff
@@ -97,6 +97,28 @@ def convert_call_to_sqlglot(
 
         return super().convert_call_to_sqlglot(operator, args, types)
 
+    def convert_listof(
+        self, args: SQLGlotExpression, types: list[PyDoughType]
+    ) -> SQLGlotExpression:
+        return sqlglot_expressions.Anonymous(this="JSON_ARRAYAGG", expressions=args)
+
+    def generate_dataframe_array_expression(
```
john-sanchez31: Can we add something about Oracle support in the documentation?

pydough/sqlglot/transform_bindings/postgres_transform_bindings.py
```diff
@@ -65,6 +76,123 @@ def convert_call_to_sqlglot(
...
+                    inner_term = "true"
+                case StringType():
+                    inner_term = "''"
+                case NumericType():
+                    inner_term = "0"
+                case DatetimeType():
+                    inner_term = "CAST('1970-01-01' AS TIMESTAMP)"
+                case _:
+                    raise ValueError(
+                        f"Cannot support empty array of type {inner_type} in Postgres."
```
john-sanchez31, line 216: Also can we add in the documentation what exactly is supported for Postgres as well?

pydough/unqualified/qualification.py
```diff
@@ -1272,6 +1278,50 @@ def qualify_cross(
...
+    ) -> PyDoughCollectionQDAG:
+        """
+        Transforms an `UnqualifiedExplode` into a PyDoughCollectionQDAG node.
+
+        Args:
+            `unqualified`: the UnqualifiedExplode instance to be transformed.
+            `context`: the collection QDAG whose context the collection is being
+            evaluated within.
+            `is_child`: whether the collection is being qualified as a child
+            of a child operator context, such as CALCULATE or PARTITION.
```
john-sanchez31, line 1295: Add Returns section

tests/testing_utilities.py
```diff
@@ -1386,6 +1399,14 @@ def run_sql_test(
         if self.skip_sql:
             pytest.skip(f"Skipping SQL text test for {self.test_name}")
 
+        if (
```
john-sanchez31, line 1402: I like this new functionality. Can you check if the skip of MySQL for infinity values, BodoSQL with to_table and DuckDb/Databricks with keywords can be skipped with this new property?

john-sanchez31 (commented):
Great job Kian, almost there. Left some comments below :)

pydough/sqlglot/transform_bindings/oracle_transform_bindings.py
knassre-bodo: Ooop, I meant to delete this one (and it is in the docs)

pydough/sqlglot/transform_bindings/postgres_transform_bindings.py
knassre-bodo: Will do. To be clear, this is a weird edge case where Postgres struggles to handle empty array literals, and this is a workaround that only works for certain types.

pydough/database_connectors/builtin_databases.py
knassre-bodo: After my changes, `DatabaseConnection` now needs to take in the dialect enum because how it reads ARRAY data will need to change depending on the dialect.

documentation/dsl.md
knassre-bodo: Array literals aren't a thing in Snowflake; you cannot put arrays inside `VALUES`

documentation/dsl.md
knassre-bodo: That's not what this does. It is just a boolean telling PyDough whether the operation will potentially remove rows  (e.g., **CAN** any of the rows from the original be an empty array.

documentation/dsl.md
knassre-bodo: No because our typing system is not fully robust/reliable

documentation/dsl.md
knassre-bodo: Added

documentation/functions.md
knassre-bodo: For the regions, sure. The other one is too big.

documentation/dsl.md
```diff
@@ -1536,6 +1537,147 @@ People.CALCULATE(Packages=COUNT(People.packages)).CROSS(Packages)
...
+Now suppose the following PyDough code is used to transform `thesaurus` using the `EXPLODE` operator (note: `filtering=True` because one of the rows is an empty array):
+```py
+%%pydough
+thesaurus.CALCULATE(word)
+         .EXPLODE(synonyms, "words", value_name='synonym', index_name='syn_idx', filtering=True)
+         .CALCULATE(word, syn_idx, synonym)
+```
+
+The result would be the following table:
+| name    | syn_idx | synonym      |
```
hadia206: shouldn't this be `word` like what is in the `CALCULATE`?

documentation/dsl.md
```diff
@@ -1536,6 +1537,147 @@ People.CALCULATE(Packages=COUNT(People.packages)).CROSS(Packages)
...
+|---------|---------|--------------|
+| 'wise'  | 0       | 'sage'       |
+| 'wise'  | 1       | 'insightful' |
+| 'wise'  | 2       | 'keen'       |
+| 'old'   | 0       | 'elderly     |
+| 'old'   | 1       | 'ancient     |
```
```suggestion
| 'old'   | 0       | 'elderly'     |
| 'old'   | 1       | 'ancient'     |
```

documentation/functions.md
```diff
@@ -1089,6 +1090,34 @@ Parts.CALCULATE(std = STD(supply_records.supply_cost))
 Parts.CALCULATE(std = STD(supply_records.supply_cost, type="sample"))
 ```
 
+<!-- TOC --><a name="std"></a>
```
hadia206, line 1093:
```suggestion
<!-- TOC --><a name="listof"></a>
```

pydough/database_connectors/database_connector.py
```diff
@@ -76,7 +107,10 @@ def execute_query_df(self, sql: str) -> pd.DataFrame:
                 # TODO: (gh #174) Cache the cursor?
                 # TODO: (gh #175) enable typed DataFrames.
                 data = self.cursor.fetchall()
-                return pd.DataFrame(data, columns=column_names)
+                pd_table = pd.DataFrame(data, columns=column_names)
+            for idx in semi_structured_cols:
+                pd_table.iloc[:, idx] = pd_table.iloc[:, idx].apply(json.loads)
```
hadia206: Do we guarantee that `pd_table.iloc[:, idx]` won't be `NULL`?

pydough/pydough_operators/expression_operators/__init__.py
```diff
@@ -48,6 +48,7 @@
     "LEQ",
     "LET",
     "LIKE",
+    "LISTOF",
```
hadia206, line 51: Should we also add it to the README file?

pydough/pydough_operators/expression_operators/registered_expression_operators.py
```diff
@@ -168,6 +169,9 @@
 QUANTILE = ExpressionFunctionOperator(
     "QUANTILE", True, RequireNumArgs(2), ConstantType(NumericType())
 )
+LISTOF = ExpressionFunctionOperator(
+    "LISTOF", True, RequireNumArgs(1), SelectArgumentType(0)
```
hadia206: Is it correct to have argument type be `SelectArgumentType(0)` not arraytype?

pydough/qdag/node_builder.py
```diff
@@ -401,6 +403,32 @@ def build_singular(
         """
         return Singular(preceding_context)
 
+    def build_explode(
+        self,
+        preceding_context: PyDoughCollectionQDAG,
+        data: PyDoughExpressionQDAG,
+        name: str,
+        explode_spec: ExplodeSpec,
+    ):
```
hadia206: return type hint

pydough/relational/relational_nodes/explode.py
```diff
@@ -0,0 +1,84 @@
...
+        input: RelationalNode,
+        explode_data: RelationalExpression,
+        explode_spec: ExplodeSpec,
+        columns: dict[str, RelationalExpression],
+    ) -> None:
+        self._explode_data: RelationalExpression = explode_data
+        self._explode_spec: ExplodeSpec = explode_spec
+        total_columns: dict[str, RelationalExpression] = {**columns}
+        total_columns[explode_spec.value_name] = ColumnReference(
+            explode_spec.value_name, explode_data.data_type
```
hadia206, line 39: why use explode_data.data_type and not `NumericType()` like it was used in HybridExplode

pydough/relational/relational_nodes/relational_expression_dispatcher.py
```diff
@@ -80,3 +81,7 @@ def visit_root(self, root: RelationalRoot) -> None:
 
     def visit_generated_table(self, generated_table) -> None:
         self.visit_common(generated_table)
+
+    def visit_explode(self, explode: Explode):
```
hadia206, line 85: return type hint

pydough/sqlglot/transform_bindings/base_transform_bindings.py
```diff
@@ -2586,6 +2648,22 @@ def generate_dataframe_item_expression(
...
+        Generate the sqlglot expression for an array of items with given pydough type.
+
+        Args:
+            `items` : The list of SQLGlotExpressions representing the items in the array.
+            `inner_type` : The mapped PydDough type for the items in the array.
+        Returns:
+            A SQLGlotExpression representing the array of items.
+        """
+        raise NotImplementedError(
+            f"Array types are not currently supported in dialect {self._visitor._expr_visitor._dialect.name}."
```
hadia206: why raise `NotImplementedError` and not PyDough exception like convert_listof

pydough/sqlglot/transform_bindings/databricks_transform_bindings.py
```diff
@@ -81,6 +91,123 @@ def convert_integer(
...
+            )
+        explode_op: SQLGlotExpression
+        lateral_columns: list[SQLGlotExpression] = [
+            sqlglot_expressions.Identifier(this="val")
+        ]
+        if val_index is None:
+            explode_op = Explode(this=explode_expr)
+        else:
+            explode_op = Posexplode(this=explode_expr)
+            lateral_columns.insert(0, sqlglot_expressions.Identifier(this="idx"))
```
hadia206, line 177: Is it possible for a query to only use exploded index and not the value? What would happen in this case here?

Same question apply to other dialects

pydough/unqualified/unqualified_node.py
```diff
@@ -435,6 +436,91 @@ def BEST(
...
+            `data` is unique within that row of the original data. If so, then
+            the index column is no longer mandatory since the values of the
+            exploded data can be used to co-identify unique rows. Default is
+            False.
+        """
+        match version:
+            case "array":
+                if delimiter is not None:
+                    raise PyDoughUnqualifiedException(
+                        "Cannot provide a `delimiter` to EXPLODE when `version` is 'array'"
```
hadia206, line 487: Why not use `assert` like the other checks below?
or make that other checks raise and exception

pydough/unqualified/unqualified_node.py
```diff
@@ -435,6 +436,91 @@ def BEST(
...
+        self,
+        data: "UnqualifiedNode",
+        name: str,
+        value_name: str,
+        index_name: str | None = None,
+        version: str = "array",
+        delimiter: str | None = None,
+        filtering: bool = True,
+        is_distinct: bool = False,
+    ):
```
hadia206: return type hint

pydough/utilities/explode_spec.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+@dataclass
+class ExplodeSpec:
+    """
+    Dataclass storing information about an EXPLODE operation being performed.
+    """
+
+    value_name: str
+    index_name: str | None
```
hadia206, line 15: Is it okay for these to be the same name and can they be the same name as an existing column? If no, I didn't see any check on that to prevent it from happening.

pydough/utilities/explode_spec.py
```diff
@@ -0,0 +1,45 @@
...
+    version: str
+    delimiter: str | None
+    filtering: bool
+    is_distinct: bool
+
+    @property
+    def arg_list_string(self):
```
hadia206: return hint type

pydough/utilities/explode_spec.py
```diff
@@ -0,0 +1,45 @@
...
+            args.append(self.index_name)
+        args.append(self.version)
+        if self.delimiter is not None:
+            args.append(self.delimiter)
+        args.append(repr(self.filtering))
+        args.append(repr(self.is_distinct))
+        return ", ".join(args)
+
+    @property
+    def keyword_arg_string(self):
```
hadia206: return hint type

tests/test_pydough_functions/simple_pydough_functions.py
```diff
@@ -59,6 +59,124 @@ def dumb_aggregation():
...
+        "exploded_nation",
+        value_name="v",
+        index_name="i",
+        version="array",
+        filtering=False,
+        is_distinct=False,
+    )
+
+
+def good_explode_04():
```
hadia206, line 98: This looks the same like `good_explode_01`, is there a reason to have it?

tests/test_pydough_functions/simple_pydough_functions.py
```diff
@@ -59,6 +59,124 @@ def dumb_aggregation():
...
+        value_name="v",
+        index_name="i",
+        version="string",
+        delimiter=" ",
+        filtering=False,
+        is_distinct=False,
+    )
+
+
+def good_explode_08():
```
hadia206, line 149: Also, this looks the same like `good_explode_05`.

hadia206: These duplicates are probably copy/paste and should be test with combination `filtering=False, is_distinct=True` for both array and string explode

tests/testing_utilities.py
```diff
@@ -1386,6 +1399,14 @@ def run_sql_test(
...
+            self.skipped_dialects is not None
+            and database.dialect.name in self.skipped_dialects
+        ):
+            pytest.skip(
+                f"Skipping SQL text test for {self.test_name} on {database.dialect.name} dialect"
```
hadia206, line 1407: nit: add reason and link to ticket (if there's one)

hadia206 (commented):
Thanks Kian for the big effort and work done here.
Overall looks great to me.
I have some minor questions that might be bugs so would like to get your answers before approving.

pydough/pydough_operators/expression_operators/registered_expression_operators.py
knassre-bodo: This was a filler, I'll add something like that. So far, our type system isn't entirely robust in that way.

pydough/relational/relational_nodes/explode.py
knassre-bodo: NumericType is for the index column; this is for the value column.

pydough/sqlglot/transform_bindings/base_transform_bindings.py
knassre-bodo: Good point

pydough/sqlglot/transform_bindings/databricks_transform_bindings.py
knassre-bodo: The index name is sometimes not generated because the index name is an optional argument to explode, but the value name is not. However, if either the index or value are not used, they will just be pruned out of the final result.

pydough/unqualified/unqualified_node.py
knassre-bodo: Generally, `assert` is to make mypy cooperate since the logic below those lines of code requires the arguments to be certain types. When doing things besides basic type checking, I use full errors.

pydough/utilities/explode_spec.py
knassre-bodo: There are checks for existing column (and tests for it) handled in qualification steps. As for each other, good point that is an error I should account for!

tests/test_pydough_functions/simple_pydough_functions.py
knassre-bodo: Good catch, supposed to be False/True

tests/test_pydough_functions/simple_pydough_functions.py
knassre-bodo: Indeed

tests/testing_utilities.py
knassre-bodo: A bit tricky to add since the reason could vary per-dialect, so it would need to become a dictionary, and also the reason could often just be "I only want to test this certain thing for this dialect." This may need to be a followup addition since it would require quite a bit of rewriting of tests.

pydough/database_connectors/database_connector.py
knassre-bodo: We do have some NULL examples in the tests, but that is a good point; I'll make it a none-handling lambda

hadia206 (approved):
Looks good to me. 
Thanks Kian

PR #547, opened 2026-08-05
Matching GETPART() behavior between dialects 

### LINKED ISSUE #403: Fix GETPART function implementation in base transform binding
The behavior of `GETPART` when there is nothing in the original string after the delimiter is currently inconsistent between the ansi/sqlite implementation versus what other dialects to akin to the Snowflake `SPLIT_PART` function (which `GETPART` is based on)

The current implementation of `GETPART` in the ansi/sqlite translation compared to `GETPART` in MySQL and Snowflake is as follow:

| GETPART |ANSI/SQLITE | MySQL/SnowFlake |
|---|---|---|
|GETPART("string", "string", 1) -> | "" | "" |
|GETPART("string", "string", 2) ->  | None| "" |
|GETPART("string", "string", 3) -> | None| None |
|GETPART("hello.world.", ".", 1) -> | "hello"|  "hello"|
|GETPART("hello.world.", ".", 2) -> | "world"|  "world" | 
|GETPART("hello.world.", ".", 3) -> | None| "" |
|GETPART("hello.world.", ".", -1) -> | None| "" |
|GETPART("hello.world.", ".", -2) -> | "world"| "world" |
|GETPART("hello.world.", ".", -3) -> | "hello"|  "hello" |
|GETPART("abab", "ab", 1) ->  | ""|  "" |
|GETPART("abab", "ab", 2) ->  | ""|  "" |
|GETPART("abab", "ab", 3) ->  | None|  "" |
|GETPART("abab", "ab", -1) -> | None|  "" |
|GETPART("abab", "ab", -2) -> | ""| "" |
|GETPART("abab", "ab", -3) -> | "" | "" |

The desire behavior for the ansi/sqlite `GETPART` to maintain consistency through the dialects is the behavior of the MySQL/Snowflake implementation.

john-sanchez31:
## Linked ticket

Closes #403 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Docs / config

## What changed and why?

- Fixes the difference between dialects for `GETPART()`. Makes index out of range and when the delimiter is at the beginning/end of the string `GETPART(".string.", ".", idx)` return empty string `''` instead of NULL. This matches the Snowflake function `SPLIT_PART` on which `GET_PART` is based.
- No more harmonization for `''` and `None`
- Since Oracle treats the empty string as NULL, override some test to match this behavior is required (`get_part_test`, `replace`, `strip`, `step_slicing`, `padding_functions`, `get_part_test` and `get_part_multiple`)

## How I tested this?

 - Through the current tests for this specific functionality for all dialects.


tests/testing_utilities.py
```diff
@@ -1749,16 +1749,8 @@ def sanitize_string(val):
             lambda x: pd.NA if pd.isna(x) else float(x)
         )
 
-    # String vs None. Convert to empty string.
```
john-sanchez31, line 1752: I had to do this for oracle but definitely no ideal for our testing harmonize empty string with NULL.

tests/testing_utilities.py
```diff
@@ -1749,16 +1749,8 @@ def sanitize_string(val):
...
-    if any(isinstance(elem, (str, NoneType)) for elem in column_a) and any(
-        isinstance(elem, (str, NoneType)) for elem in column_b
-    ):
-        return column_a.apply(lambda x: "" if pd.isna(x) else str(x)), column_b.apply(
-            lambda x: "" if pd.isna(x) else str(x)
-        )
-
     # float vs None. Convert to nullable floats
-    if any(isinstance(elem, (float, NoneType)) for elem in column_a) and any(
+    if all(isinstance(elem, (float, NoneType)) for elem in column_a) and all(
```
john-sanchez31, line 1753: Prevents a column with None type and for example string entering the if.

tests/testing_utilities.py
hadia206: Can you explain why this was removed? I see you already updated the Oracle test reference solution to be `None`.

pydough/sqlglot/transform_bindings/base_transform_bindings.py
```diff
@@ -2058,7 +2058,8 @@ def convert_get_part(
                     '' AS part,
                     FIRST_ARGUMENT AS rest,
                     SECOND_ARGUMENT AS delim,
-                    THIRD_ARGUMENT AS idx
+                    THIRD_ARGUMENT AS idx,
+                    0 AS is_last
```
hadia206: Add comment clarifying what's the purpose of this.

tests/test_pipeline_oracle.py
```diff
@@ -305,20 +302,236 @@ def defog_custom_oracle_test_data(
...
+    custom_functions_test_data: PyDoughPandasTest,  # noqa: F811
+) -> PyDoughPandasTest:
+    """
+    Modify reference solution data for some custom queries.
+    Return an instance of PyDoughPandasTest containing the modified data.
+    """
+    if custom_functions_test_data.test_name == "get_part_test":
+        # In Oracle None and empty string are treated the same, so we need to
+        # modify the reference solution data to match this behavior.
+        return PyDoughPandasTest(
```
hadia206, line 464: Isn't `get_part_test` already updated above in the same file. Why change here again?

hadia206 (commented):
Looks good to me. I have a couple of questions regarding Oracle tests.

tests/testing_utilities.py
john-sanchez31: Yes, this was added because Oracle return None instead of '', but not ideal to treat `'' == NULL` for our testing since for Python and other dialects `''` is different than `NULL`

tests/test_pipeline_oracle.py
john-sanchez31: Yes, this test is from `custom_functions` not `defog` so I deleted the first one good catch!

tests/testing_utilities.py
hadia206: Still can't get why this is removed. This just checks if we have `None`, replace it with `''` to be consistent. So if your refsol is `''` and Oracle returns `None`, it should enter here and replaces it to `''` and now it would match the refsol result.

john-sanchez31: You're right, but this also hides the differences in the results for other dialects too. For example with Snowflake  `GETPART(".string.", ".", 4) -> ""` but with MySQL `GETPART(".string.", ".", 4) -> None`. Basically with this we are saying that `"" == NULL` in our testing which I'd say is not `True`, because expect for Oracle the other dialects (also Python) treat `"" == NULL` as `False`.

hadia206: Oh I get it now, thanks for clarification

knassre-bodo (approved):
Make sure to document the new behavior in `functions.md`, otherwise LGTM :)

PR #551, opened 2026-08-18
Templates support

### LINKED ISSUE #173: Support parameters in executed SQL queries
In some situations rather than passing literals for building a queries a User may actually want to provide parameters for building a parameterized query. An example reason for this could be to allow caching the same underlying query and then asking "WHAT-IF" solely by modifying the underlying values.

Currently we don't have a way for users to specify a set of literals as parameters and we don't have any way to provide parameters to the underlying execution. This is more difficult because we may need to precompile the underlying query.

john-sanchez31:
## Linked ticket
Closes #173

## Type of change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Docs / config

## What changed and why?
Adds support for **PyDough templates** — reusable, parametrized PyDough/Python snippets registered in metadata, so common query patterns can be defined once and reused. This PR adds:
- A new `templates` metadata object, with `attributes` (labeled values usable as template arguments) and `definitions` (the templates themselves).
- Support for calling a loaded template directly, as an ordinary PyDough function, or via the new `pydough.call_template(name, labels=...)` API using human-facing labels.
- Support for recursive templates, templates returning plain literals, and templates that build a user-generated collection from an input `pd.DataFrame`.

## How I tested this?
- New test file `test_templates.py`, covering end-to-end template calls (direct and via `call_template`), metadata parsing/validation error cases, and `call_template` runtime error cases (unknown template, unknown/missing labels, type mismatches, disallowed usage).

## Notes for reviewers
Minor `from_string` refactor as part of this: `from_string` and `call_template` now share a common `_execute_source` helper, with a `return_unqualified` flag (`True` requires the source to produce an `UnqualifiedNode`; `False` allows any return type, needed for literal-returning templates).


documentation/dsl.md
```diff
@@ -1923,6 +1924,258 @@ Result:
 2          12    1
 ``` 
 
+<!-- TOC --><a name="templates"></a>
+## PyDough Templates
+
+Once a template is registered in metadata (see [Templates](metadata#templates) for how to define one), it can be used in two ways:
+
+- **Directly**, as an ordinary PyDough function call, passing raw literal/PyDough values as arguments.
+- **Through the API**, via [`pydough.call_template`](usage#call_template_api), passing human-facing labels instead of raw values.
```
hadia206: A question, can we mix and match in the same query between both usage types? I'm assuming yes, but didn't see an example so want to confirm. If we can't , we need to make that explicit in the docs.

hadia206: nit: I suggest using `user-facing` instead of `human-facing`

documentation/dsl.md
```diff
@@ -1923,6 +1924,258 @@ Result:
...
+selected_customers = customers.WHERE(account_balance >= min_account_balance)
+final_result = TPCH.CALCULATE(n_custs=COUNT(selected_customers))
+
+pydough.to_df(final_result)
+```
+
+<!-- TOC --><a name="using-templates-dataframe"></a>
+### Templates Producing a User-Generated Collection
+
+A template's `source` can also produce a **user-generated collection** — e.g. one built from an in-memory `pd.DataFrame` via `pydough.dataframe_collection` — rather than deriving from the graph itself. This works like any other template, but the `pd.DataFrame` and `list` typed parameters let external, ad hoc data be wrapped into a proper collection and then combined with the rest of the graph (e.g. via `CROSS`, `WHERE`, `CALCULATE`) just as if it were a regular collection.
```
```suggestion
A template's `source` can also produce a **user-generated collection** — e.g. one built from an in-memory `pd.DataFrame` via `pydough.dataframe_collection` — rather than deriving from the graph itself. This works like any other template, but the `pd.DataFrame` and `list` typed parameters let you wrap external, ad hoc data into a a proper collection and combine it with the rest of the graph (e.g. via `CROSS`, `WHERE`, `CALCULATE`) just as if it were a regular collection.
```

documentation/dsl.md
```diff
@@ -1923,6 +1924,258 @@ Result:
...
+    name, n_customers=COUNT(customers.WHERE(selected_orders > 0))
+).TOP_K(3, by=(n_customers.DESC(), name.ASC()))
+
+pydough.to_df(result)
+```
+
+<!-- TOC --><a name="using-templates-follow-up"></a>
+### Chaining Templates Together
+
+The result of one template call can be fed directly into another template as an argument, letting templates build on each other without needing to re-express the earlier result in raw PyDough. Works the same way whether the first call was made directly or through the API.
```
hadia206: nit:
```suggestion
The result of one template call can be fed directly into another template as an argument, letting templates build on each other without needing to re-express the earlier result in raw PyDough. It works the same way whether the first call was made directly or through the API.
```

documentation/metadata.md
```diff
@@ -505,6 +506,278 @@ Below are several examples the JSON for such deducers:
...
+
+<!-- TOC --><a name="template-attributes"></a>
+### Attributes
+
+An attribute describes a reusable, labeled set of values that one or more templates can use as an argument. Attributes are exclusively used with `call_template()` — they map a user-facing **label** (e.g. `"Year 1996"`) to the actual PyDough/Python **value** (e.g. `1996`) that gets substituted into the template.
+
+Every JSON object describing an attribute has the following fields:
+- `name` (required): unique identifier for the attribute. Referenced by templates that use it.
+- `usage` (required): which templates (and optionally which parameters of those templates) may use this attribute. An empty list `[]` means "usable by all templates." [See here](#attribute-usage-restriction) for the specification of this field.
+- `type` (required): the value's type. This is used to verify if the value match the template's parameter. To know more about the supported types [See here](#template-supported-types)
```
hadia206:
```suggestion
- `type` (required): the value's type. This is used to verify if the value matches the template's parameter. [See supported types](#template-supported-types) for more.
```

documentation/metadata.md
```diff
@@ -505,6 +506,278 @@ Below are several examples the JSON for such deducers:
...
+*Note*: `pydough` is a special type that represents a raw PyDough expression/snipet.
```
hadia206:
```suggestion
*Note*: `pydough` is a special type that represents a raw PyDough expression/snippet.
```

documentation/metadata.md
```diff
@@ -505,6 +506,278 @@ Below are several examples the JSON for such deducers:
...
+{ "label": "DICT VALUE", "value": "{'key': pydough.expression}" }
+```
+
+<!-- TOC --><a name="template-definitions"></a>
+### Template Definitions
+
+A definition describes one callable template: its parameters, its PyDough source code, and what it returns.
+
+Every JSON object describing a template definition has the following fields:
+- `name` (required): unique template name, which must be a valid PyDough identifier and must not overlap with the name of other collections/properties/relationships/functions within the graph. Becomes both the generated PyDough function name and the string passed to `call_template()`. This name will become reserved like other function names in PyDough (`COUNT`, `LOWER`, etc.)
```
hadia206:
```suggestion
- `name` (required): unique template name, which must be a valid PyDough identifier and must not overlap with the names of other collections/properties/relationships/functions within the graph. It becomes both the generated PyDough function name and the string passed to `call_template()`. This name will become reserved like other function names in PyDough (`COUNT`, `LOWER`, etc.)
```

documentation/metadata.md
```diff
@@ -505,6 +506,278 @@ Below are several examples the JSON for such deducers:
...
+- `description` (required): a semantic description of what the template computes.
+- `parameters` (required, `{}` if none): the template's arguments, keyed by parameter name. [See here](#template-parameters) for the specification of these objects.
+- `source` (required): the PyDough code body. Must assign to the variable named in `answer_variable`. Positional placeholders (`{1}`, `{2}`, …) are substituted with parameter values/expressions before evaluation. All details about `source` for template definitions can be found [here](#template-source).
+- `answer_variable` (required): name of the variable inside `source` whose value the generated function returns.
+
+<!-- TOC --><a name="template-parameters"></a>
+#### Definition Field: `parameters`
+
+`parameters` is a JSON object (map) keyed by parameter name. Each value is a JSON object with the following fields:
+- `type` (required): `"int"`, `"pydough"`, etc. `"pydough"` signals the value is a raw expression rather than a literal. To know more about the supported types [See here](#template-supported-types)
```
hadia206:
```suggestion
- `type` (required): `"int"`, `"pydough"`, etc. `"pydough"` signals the value is a raw expression rather than a literal. [See supported types](#template-supported-types) for more.
```

documentation/metadata.md
```diff
@@ -505,6 +506,278 @@ Below are several examples the JSON for such deducers:
...
+}
+```
+
+<!-- TOC --><a name="template-naming-collisions"></a>
+### Naming Collisions
+
+`name` (for both attributes and template definitions) must be validated against:
+1. Other attribute/template names (no duplicates within their own graph).
+2. Existing PyDough built-in functions (e.g. `GETPART`, `STRCOUNT`) — a template cannot shadow one of these.
+3. Python builtins, to the extent the generated function is exposed in a namespace where that matters.
```
hadia206:
```suggestion
3. Python reserved keywords and built-in names (e.g `list`), to avoid the name colliding with either.
```

documentation/metadata.md
```diff
@@ -505,6 +506,278 @@ Below are several examples the JSON for such deducers:
 - Returns an array of strings: `{"type": "array of", "element type": {"type": "constant", "value": "string"}}`
 - Returns an array where the elements have the same type as the first argument: `{"type": "array of", "element type": {"type": "select argument", "value": 0}}`
 
+
+<!-- TOC --><a name="templates"></a>
+## Templates
+
+A template registers a reusable, **parametrized function** in metadata, which becomes available for later use either as a normal PyDough function call or through the `pydough.call_template()` API using human-readable **labels** instead of raw values. More about templates' usage [here](dsl#pydough-templates)
```
hadia206:
```suggestion
A template registers a reusable, **parametrized function** in metadata, which becomes available for later use either as a normal PyDough function call or through the `pydough.call_template()` API using human-readable **labels** instead of raw values. Know more about templates' usage [here](dsl#pydough-templates).
```

documentation/metadata.md
```diff
@@ -505,6 +506,278 @@ Below are several examples the JSON for such deducers:
...
+        result = COUNT(orders.WHERE(YEAR(order_date) == base_year))
+    else:
+        result = (
+            COUNT(orders.WHERE(YEAR(order_date) == base_year))
+            + cumulative_orders_counter(base_year + 1, last_year)
+        )
+    return result
+```
+
+You can find how to use templates and more examples [here](dsl#pydough-templates)
```
hadia206:
```suggestion
You can find more on how to use templates, along with more examples, [here](dsl#pydough-templates).
```

documentation/metadata.md
```diff
@@ -505,6 +506,278 @@ Below are several examples the JSON for such deducers:
...
+}
+```
+
+An empty object `{}` means the attribute is usable by any parameter of any template. Also an empty list `[]` for a template indicates that the attribute can be used in any parameter of the template.
+
+<!-- TOC --><a name="attribute-options"></a>
+#### Attribute Field: `options`
+
+Each entry in `options` is a JSON object with the following fields:
+- `label` (required): the human-facing name shown to / provided by the user (e.g. in a question or UI). This is what `call_template()` accepts. This label must be unique among all labels in the graph.
```
hadia206:
```suggestion
- `label` (required): the human-facing name shown to or provided by the user (e.g. in a question or UI). This is what `call_template()` accepts. This label must be unique among all labels in the graph.
```

documentation/metadata.md
```diff
@@ -505,6 +506,278 @@ Below are several examples the JSON for such deducers:
...
+The `usage` field must unambiguously express **which templates**, and **which specific parameter(s) of those templates**, can use the attribute. This field is a JSON object (map) keyed by template name, whose value is a list of parameter names within that template allowed to use the attribute:
+
+```json
+"usage": {
+  "orders_revenue_by": ["arg_dimension"],
+  "top_bottom_comparison": ["arg_calculation"]
+}
+```
+
+An empty object `{}` means the attribute is usable by any parameter of any template. Also an empty list `[]` for a template indicates that the attribute can be used in any parameter of the template.
```
hadia206:
```suggestion
An empty object `{}` means the attribute is usable by any parameter of any template. Similarly, an empty list `[]` for a template indicates that the attribute can be used in any parameter of the template.
```

documentation/metadata.md
```diff
@@ -505,6 +506,278 @@ Below are several examples the JSON for such deducers:
...
+#### Attribute Field: `options`
+
+Each entry in `options` is a JSON object with the following fields:
+- `label` (required): the human-facing name shown to / provided by the user (e.g. in a question or UI). This is what `call_template()` accepts. This label must be unique among all labels in the graph.
+- `value` (required): the literal or PyDough snippet substituted into the template source when this label is selected.
+
+<!-- TOC --><a name="attribute-options-list-dict"></a>
+##### Options with `list` or `dict` Values
+
+For attributes whose `type` is `list` or `dict`, `value` cannot be written as a native JSON array/object, because a raw JSON array or object cannot hold an unquoted PyDough expression (e.g. `pydough.expression`) — JSON only allows quoted strings, numbers, booleans, `null`, or nested arrays/objects as elements. Instead, the entire list or dict literal is written as a single string, using Python/PyDough syntax. This string is substituted verbatim into the source (it is not JSON-decoded).
```
hadia206:
```suggestion
For attributes whose `type` is `list` or `dict`, `value` cannot be written as a native JSON array/object, — JSON only allows quoted strings, numbers, booleans, `null`, or nested arrays/objects as elements, so it has no way to represent an unquoted PyDough expression (e.g. `pydough.expression`). Instead, the entire list or dict literal is written as a single string, using Python/PyDough syntax. This string is substituted verbatim into the source (it is not JSON-decoded).
```

documentation/usage.md
```diff
@@ -966,6 +967,72 @@ ORDER BY
...
+| Condition | Message behavior |
+|---|---|
+| No metadata is loaded in the active session | `"No metadata loaded in the current active session"` |
+| `name` doesn't match any registered template | Reports the template isn't found; if similarly-named templates exist, suggests them. |
+| The graph has no attributes registered at all | Reports that no attributes are available for the template. |
+| `arg_name` in `labels` isn't a parameter of the template | Reports the parameter doesn't exist; suggests the template's actual parameter names. |
+| `label` isn't restricted to this template/parameter by its attribute's `usage` | Reports the label isn't available for that parameter on that template. |
+| The matching attribute's `type` differs from the parameter's expected `type` | Reports the type mismatch between the attribute and the parameter. |
+| `label` doesn't appear in any attribute's `options` | Reports the label wasn't found in any attribute's options. |
+| A required template parameter has no corresponding entry in `labels` | Reports the template is missing labels for those parameter(s), by name. |
```
hadia206:
```suggestion
| A required template parameter has no corresponding entry in `labels` | Reports which parameter(s) it's missing labels for, by name. |
```

documentation/metadata.md
```diff
@@ -505,6 +506,278 @@ Below are several examples the JSON for such deducers:
...
+
+Example of the structure of the metadata for an attribute named `years`:
```
hadia206: Felt repetitive to me so a suggestion to just do it like this. Repeat for all other places
```suggestion
Example: an attribute named `years`:
```

documentation/metadata.md
```diff
@@ -505,6 +506,278 @@ Below are several examples the JSON for such deducers:
...
+  "type": "int",
+  "description": "The year for which to calculate the revenue of orders.",
+  "options": [
+    { "label": "Year 1992", "value": 1992},
+    { "label": "Year 1993", "value": 1993}
+  ]
+}
+```
+
+Example of the structure of the metadata for an attribute of type `pydough`, whose options are filter conditions:
```
hadia206:
```suggestion
Example: a `pydough`-typed attribute whose options are filter conditions:
```

documentation/metadata.md
```diff
@@ -505,6 +506,278 @@ Below are several examples the JSON for such deducers:
...
+  "options": [
+    { "label": "Status O", "value": "(order_status == 'O')" },
+    { "label": "Price above 3000", "value": "(total_price > 3000)" },
+    { "label": "March", "value": "(MONTH(order_date) == 3)" },
+    { "label": "High priority", "value": "ISIN(order_priority, ('1-URGENT', '2-HIGH'))" }
+  ]
+}
+```
+
+Example of the structure of the metadata for an attribute of type `str`:
```
hadia206:
```suggestion
Example: `str`-typed attribute:
```

pydough/metadata/graphs/graph_metadata.py
```diff
@@ -188,8 +224,103 @@ def add_function(self, name: str, function: "ExpressionFunctionOperator") -> Non
             raise PyDoughMetadataException(
                 f"Function name {name!r} cannot be the same as a collection name in {self.error_name}"
             )
+        if name in builtin_registered_operators():
+            raise PyDoughMetadataException(
+                f"Function name {name!r} already in use for a PyDough operador"
```
hadia206:
```suggestion
                f"Function name {name!r} already in use for a PyDough operator"
```

pydough/metadata/graphs/graph_metadata.py
```diff
@@ -188,8 +224,103 @@ def add_function(self, name: str, function: "ExpressionFunctionOperator") -> Non
...
+            if self.templates_definitions[name] == new_template:
+                raise PyDoughMetadataException(
+                    f"Already added {name!r} to {self.error_name}"
+                )
+            raise PyDoughMetadataException(
+                f"Duplicate templates: {name} versus {self.templates_definitions[name]}"
+            )
+        if name in builtin_registered_operators():
+            raise PyDoughMetadataException(
+                f"The template '{name}' already exists as a PyDough operador"
```
hadia206:
```suggestion
                f"The template '{name}' already exists as a PyDough operator"
```

pydough/metadata/graphs/graph_metadata.py
```diff
@@ -34,6 +37,20 @@ class GraphMetadata(AbstractMetadata):
...
 
+    ALLOWED_TYPES: set[str] = {
+        "str",
+        "int",
+        "float",
+        "list",
+        "dict",
+        "pydough",
+        "pd.DataFrame",  # Type for pandas dataframe for Dataframe Collections
+        "datetime",  # Type for datatimes
```
hadia206: Do we need the comment here, the name seems obvious :)?

pydough/metadata/graphs/graph_metadata.py
```diff
@@ -188,8 +224,103 @@ def add_function(self, name: str, function: "ExpressionFunctionOperator") -> Non
...
+        Validate that type_str is exactly one of the allowed type names.
+
+        Args:
+            `type_str`: the type name to validate.
+
+        Returns:
+            Bool: True if type_str is a valid type name, False otherwise.
+        """
+
+        return type_str.strip() == type_str and type_str in self.ALLOWED_TYPES
```
hadia206: Why the `strip` check here?

pydough/metadata/templates/attribute_metadata.py
```diff
@@ -0,0 +1,202 @@
...
+                    graph_labels[label] if label in graph_labels else attribute_name
+                )
+                raise PyDoughMetadataException(
+                    f"Duplicate option label: {label!r} for attribute {attribute_name!r}. "
+                    f"The label is already in use by attribute {by_attribute!r} "
+                    f"in graph {graph.name!r}."
+                )
+            new_attribute.add_attribute_option(label, value)
+
+        graph.add_template_attribute(new_attribute)
```
hadia206, line 222: What happens for extra fields in the JSON? Is it acceptable or should it have ` NoExtraKeys` check?
See `SimpleTableMetadata.parse_from_json` for an example

pydough/metadata/templates/attribute_metadata.py
```diff
@@ -0,0 +1,202 @@
...
+        return self._options
+
+    @property
+    def error_name(self):
+        return self.create_error_name(self.name, self.graph.error_name)
+
+    @property
+    def components(self):
+        comp: list = [self.name, self.description, self.type]
+        comp.extend(self.usage)
```
hadia206: Add options too, and why doing `comp.extend` not adding directly to the list? with `extend` it only adds the keys not the whole dict.

comp: list = [self.name, self.description, self.type, self.options, self.usage]

pydough/metadata/templates/template_metadata.py
```diff
@@ -0,0 +1,370 @@
...
+        """
+        return self._template_callable
+
+    @property
+    def error_name(self):
+        return self.create_error_name(self.name, self.graph.error_name)
+
+    @property
+    def components(self):
+        comp: list = [self.name, self.description, self.source]
```
hadia206: Same. Should it have `parameters` and `answer_variable`?

pydough/metadata/templates/template_metadata.py
```diff
@@ -0,0 +1,370 @@
...
+        new_template: TemplateMetadata = TemplateMetadata(
+            name,
+            graph,
+            description,
+            kwargs,
+            source,
+            answer_variable,
+        )
+
+        graph.add_template_definition(name, new_template)
```
hadia206, line 200: Same, should it have `NoExtraKeys(...)` check?

pydough/metadata/templates/template_metadata.py
```diff
@@ -0,0 +1,370 @@
...
+            part = f"{key}: {spec.type}" if spec.type != "pydough" else f"{key}"
+            args_parts.append(part)
+        template_args: str = ", ".join(args_parts)
+
+        # --- Replace {1}, {2}, ... with the corresponding argument name ---
+        # NOTE: Add a detail documentation about this replacement
+        def _replace_placeholder(match: re.Match) -> str:
+            index = int(match.group(1))
+            if not (1 <= index <= len(arg_names)):
+                raise ValueError(
```
hadia206: Why raise ValueError not `PyDoughMetadataException` like everywhere else?

pydough/metadata/templates/template_metadata.py
```diff
@@ -0,0 +1,370 @@
...
+        substituted_code = re.sub(r"\{(\d+)\}", _replace_placeholder, self.source)
+
+        # --- Indent and assemble the final template ---
+        indented_code: str = textwrap.indent(substituted_code.strip(), "    ")
+
+        template_str: str = f"def {self.name}({template_args}):\n{indented_code}\n    return {self.answer_variable}\n"
+
+        return template_str
+
+    def create_template_call(self, kwargs: dict[str, dict[str, str | int]] = {}) -> str:
```
hadia206: `{}` is a mutable default argument, it's not recommended to do so in Python.
```suggestion
    def create_template_call(self, kwargs: dict[str, dict[str, str | int]] = None) -> str:
```

Then initialize it inside the function `kwargs = kwargs or {}`

pydough/metadata/templates/template_metadata.py
```diff
@@ -0,0 +1,370 @@
...
+        intended to be embedded in a larger source snippet (e.g. via
+        `from_string`) rather than executed on its own.
+        """
+
+        args_parts: list = []
+        part: str = ""
+        for key, spec in kwargs.items():
+            display_quotes: bool = "type" in spec and spec["type"] == "str"
+
+            arg_value = f"'{spec['value']}'" if display_quotes else f"{spec['value']}"
```
hadia206: What would happen if the value contains a single quote? `It's`

pydough/unqualified/unqualified_node.py
```diff
@@ -135,8 +136,21 @@ def __getitem__(self, key):
...
+        if pydough.active_session.metadata:
+            metadata_templates: dict[str, TemplateMetadata] = (
+                pydough.active_session.metadata.templates_definitions
+            )
+
+            name = str(self)
+            if name in metadata_templates:
+                return metadata_templates[name].template_callable(*args, **kwargs)
+
+            available_templates.extend(metadata_templates.keys())
```
hadia206: This doesn't look right to me. Can we resolve templates the same way UDFs are resolved?
Right now `__call__` looks up templates through `pydough.active_session.metadata` (a global), instead of using the graph that was actually passed in for this query. That can break if more than one graph is in use at the same time, since the global might not match the graph this code is actually running against.

Let's do what UDFs are doing it. Check `UnqualifiedRoot.__init__` in `unqualified_node.py`

It builds func_map once, directly from the graph that was passed in. So we should do the same when `UnqualifiedRoot` is created, using `graph.templates_definitions` instead of `pydough.active_session.metadata.templates_definitions`.

That way you would also be able to remove the `pydough.active_session` swapping that got added to `tests/testing_utilities.py`

pydough/unqualified/unqualified_transform.py
```diff
@@ -534,14 +590,161 @@ def from_string(
...
+        raise ValueError(
+            f"Template {name!r} is missing a label for the following "
+            f"parameter(s): {missing_params_str}"
+        )
+
+    available_attributes: dict[str, AttributeMetadata] = (
+        pydough.active_session.metadata.templates_attributes
+    )
+
+    if len(available_attributes) == 0:
```
hadia206: This check looks to me like "does this graph have any attributes at all," not "does this template call need any." 
It requires the graph to have at least one attribute defined anywhere, even for a template like `order_revenue` (which takes zero parameters and needs no attributes at all). It happens to work in the TPC-H sample graph only because other templates there define attributes, but on a graph with no `templates.attributes` section at all, calling any template via `call_template()`, even a zero-param one, would fail with `"No attributes available"`, which seems like the wrong behavior. Shouldn't this check be scoped to whether this specific call needs any attributes, rather than whether the graph has any at all?

documentation/metadata.md
```diff
@@ -505,6 +506,278 @@ Below are several examples the JSON for such deducers:
...
+  "answer_variable": "result"
+}
+```
+
+Example of the structure of the metadata template named `order_revenue`:
+
+```json
+{
+  "name": "order_revenue",
+  "dependencies": [],
```
hadia206: Is this used anywhere?
I don't see it in `Template Definitions` field spec and and the actual parser `TemplateMetadata.parse_from_json` doesn't reads a "dependencies" key.
Also, not in any of the test cases

tests/test_pydough_functions/tpch_templates.py
```diff
@@ -0,0 +1,122 @@
+"""
```
hadia206, line 1: Since this file deals with user collection specifically, I suggest rename filename to reflect that.

tests/test_templates.py
```diff
@@ -0,0 +1,644 @@
+"""
+TODO
```
hadia206: Reminder

tests/test_templates.py
```diff
@@ -0,0 +1,644 @@
...
+"""
+
```
hadia206, line 4: Could you also add tests for these cases
- Test for `float`, `dict`, `datetime`, and `pd.DataFrame` as attribute types
- Test for the two documented usage shorthand forms: an empty object {} (attribute usable by any template) and an empty list [] 
- Test for a graph with zero attributes calling a zero-parameter template via `call_template()`
- Test for a string-typed attribute option value containing a quote character
- Test for out of range N in the placeholder in the template source

hadia206 (commented):
Great work John.
Please see my comments below.

documentation/dsl.md
john-sanchez31: The example `Chaining Templates Together` uses both calls on the same pydough code. Is this the type of example you're referring to?

documentation/metadata.md
john-sanchez31: Yes, at the end it was not necessary. I'll delete it

pydough/metadata/graphs/graph_metadata.py
john-sanchez31: Now that I'm looking closer seems unnecessary. Fixed

pydough/metadata/templates/attribute_metadata.py
john-sanchez31: Added check

pydough/metadata/templates/template_metadata.py
john-sanchez31: Added

pydough/metadata/templates/template_metadata.py
john-sanchez31: I didn't notice, fixed

pydough/metadata/templates/template_metadata.py
john-sanchez31: It would brake it, I added the fix plus 2 new test for this case. Good catch!

pydough/unqualified/unqualified_node.py
john-sanchez31: I changed how the template is called directly, but I still need the session swapping in `transform_and_exec_pydough` for tests that are build with a function instead of a string

tests/test_pydough_functions/tpch_templates.py
john-sanchez31: I changed the docstring. This files contains function for testing templates on tpch

pydough/unqualified/unqualified_node.py
```diff
@@ -135,10 +136,30 @@ def __getitem__(self, key):
...
+        graph = self._locate_root_graph()
+        available_templates: list[str] = (
+            list(graph.templates_definitions.keys()) if graph is not None else []
+        )
         raise pydough.active_session.error_builder.undefined_function_call(
-            self, *args, **kwargs
+            self, available_templates, *args, **kwargs
         )
 
+    def _locate_root_graph(self) -> "GraphMetadata | None":
```
knassre-bodo, line 149: Adding methods like this to `UnqualifiedNode` isn't a thing we should do because then it means you are adding a new banned name for PyDough fields. Granted, `_locate_root_graph` isn't a likely name to have, but still we should not do it this way. Instead, make it a function defined in this file that takes in an UnqualifiedNode and finds the root graph.

pydough/unqualified/unqualified_node.py
```diff
@@ -135,10 +136,30 @@ def __getitem__(self, key):
...
+        and returns its bound graph. A template call is always a name resolved
+        directly off the root, so this should only return None in genuinely
+        unreachable cases (e.g. no graph loaded at all).
+        """
+        node: UnqualifiedNode = self
+        while isinstance(node, UnqualifiedAccess):
+            node = node._parcel[0]
+        if isinstance(node, UnqualifiedRoot):
+            return node._parcel[0]
+        return None
```
knassre-bodo, line 161: There are other ways to get to the root from an unqualified node. E.g. where/calculate/partition should go to the predecessor, `CROSS` should search one of the sides for the root, etc.

knassre-bodo (commented):
LGTM @john-sanchez31! Just some tiny notes on the UnqualifiedNode section

pydough/unqualified/unqualified_node.py
john-sanchez31: Moved outside the class

documentation/dsl.md
hadia206: I meant one uses the call_template and the other uses template directly.

documentation/usage.md
```diff
@@ -985,6 +986,72 @@ ORDER BY
   3 DESC
 ```
 
+<!-- TOC --><a name="call_template_api"></a>
+### `pydough.call_template`
+
+The `call_template` API invokes a previously registered [template](metadata#templates) using human-facing **labels** instead of raw PyDough/Python values. Each label is resolved against the `options` of the graph's [attributes](metadata#template-attributes) to find the concrete value (and its type) to pass into the template, after which the template is called like a normal PyDough function and the result is returned.
```
hadia206:
```suggestion
The `call_template` API invokes a previously registered [template](metadata#templates) using user-facing **labels** instead of raw PyDough/Python values. Each label is resolved against the `options` of the graph's [attributes](metadata#template-attributes) to find the concrete value (and its type) to pass into the template, after which the template is called like a normal PyDough function and the result is returned.
```

documentation/dsl.md
```diff
@@ -1923,6 +1924,258 @@ Result:
...
+    n_orders=orders_filter_count(
+        (total_price > 3000) & (YEAR(order_date) == 1996)
+    ),
+).TOP_K(5, by=(n_orders.DESC(), key.ASC()))
+```
+
+<!-- TOC --><a name="using-templates-api"></a>
+### Calling a Template via `pydough.call_template`
+
+Instead of passing raw values, `pydough.call_template` looks up each argument by a human-facing **label**, resolving it to the underlying value through the graph's [attributes](metadata#template-attributes). See [`pydough.call_template`](usage#call_template_api) for the full parameter/error reference.
```
```suggestion
Instead of passing raw values, `pydough.call_template` looks up each argument by a user-facing **label**, resolving it to the underlying value through the graph's [attributes](metadata#template-attributes). See [`pydough.call_template`](usage#call_template_api) for the full parameter/error reference.
```

documentation/metadata.md
```diff
@@ -505,6 +506,277 @@ Below are several examples the JSON for such deducers:
...
+}
+```
+
+An empty object `{}` means the attribute is usable by any parameter of any template. Similarly, an empty list `[]` for a template indicates that the attribute can be used in any parameter of the template.
+
+<!-- TOC --><a name="attribute-options"></a>
+#### Attribute Field: `options`
+
+Each entry in `options` is a JSON object with the following fields:
+- `label` (required): the human-facing name shown to or provided by the user (e.g. in a question or UI). This is what `call_template()` accepts. This label must be unique among all labels in the graph.
```
hadia206:
```suggestion
- `label` (required): the user-facing name shown to or provided by the user (e.g. in a question or UI). This is what `call_template()` accepts. This label must be unique among all labels in the graph.
```

documentation/metadata.md
```diff
@@ -505,6 +506,277 @@ Below are several examples the JSON for such deducers:
 - Returns an array of strings: `{"type": "array of", "element type": {"type": "constant", "value": "string"}}`
 - Returns an array where the elements have the same type as the first argument: `{"type": "array of", "element type": {"type": "select argument", "value": 0}}`
 
+
+<!-- TOC --><a name="templates"></a>
+## Templates
+
+A template registers a reusable, **parametrized function** in metadata, which becomes available for later use either as a normal PyDough function call or through the `pydough.call_template()` API using human-readable **labels** instead of raw values. Know more about templates' usage [here](dsl#pydough-templates)
```
hadia206:
```suggestion
A template registers a reusable, **parametrized function** in metadata, which becomes available for later use either as a normal PyDough function call or through the `pydough.call_template()` API using user-readable **labels** instead of raw values. Know more about templates' usage [here](dsl#pydough-templates)
```

documentation/metadata.md
```diff
@@ -505,6 +506,277 @@ Below are several examples the JSON for such deducers:
...
+All template-related metadata lives under a single top-level `"templates"` key, with two sub-parts:
+
+```json
+"templates": {
+  "attributes": [ ... ],
+  "definitions": [ ... ]
+}
+```
+
+- **`attributes`** — (optional) the shared pool of *values* (with human-readable labels) that templates can use as an argument. Only needed if a template will be invoked via `call_template()`; a template that is only ever called directly with literal arguments does not require any attributes.
```
hadia206:
```suggestion
- **`attributes`** — (optional) the shared pool of *values* (with user-readable labels) that templates can use as an argument. Only needed if a template will be invoked via `call_template()`; a template that is only ever called directly with literal arguments does not require any attributes.
```

documentation/metadata.md
```diff
@@ -505,6 +506,277 @@ Below are several examples the JSON for such deducers:
...
+### Definition Field: `source`
+
+`source` is not limited to a single PyDough expression — it is the **body of a Python function**, and can contain anything valid inside one: `if`/`else` branching, `assert` statements, intermediate variables, loops, and even recursive calls back to the template itself (referenced by its own `name`). The only requirements are:
+
+- The code must assign to the variable named in `answer_variable` along every code path.
+- Positional placeholders (`{1}`, `{2}`, …) are substituted with the corresponding parameter names — following the [parameter numeration](#template-parameter-numeration) — before the code is parsed.
+
+Because of this, a template's `answer_variable` does not need to hold a PyDough collection/expression — it can just as easily be a plain Python/PyDough literal (`int`, `str`, `dict`, etc.), the result of ordinary Python control flow, or the result of the template recursively calling itself.
+
+Example of the structure of the metadata for a recursive template, which sums a per-year order counter from `base_year` through `last_year`:
```
hadia206: Do same changes done before [See here](https://github.com/bodo-ai/PyDough/pull/551#discussion_r4020511256)

documentation/metadata.md
```diff
@@ -505,6 +506,277 @@ Below are several examples the JSON for such deducers:
...
+- `description` (required): a semantic description of the parameter's purpose.
+
+<!-- TOC --><a name="template-parameter-numeration"></a>
+##### Parameter Numeration
+
+Parameters are numbered implicitly by their position within the `parameters` object: the first key defined is parameter `1`, the second is parameter `2`, and so on. This numeration determines which placeholder each parameter fills in `source` — parameter `1` replaces every `{1}` in the source code, parameter `2` replaces every `{2}`, etc.
+
+> **Note:** this relies on the `parameters` object preserving key insertion order. All standard JSON parsers used by this project (e.g. Python's `json` module) preserve insertion order, but this ordering is not part of the JSON specification itself — reordering, alphabetizing, or otherwise regenerating the keys of a `parameters` object will change which placeholder each parameter maps to.
+
+Example of the structure of the metadata for a parameter map with two parameters — `arg_year` is `{1}` and `arg_dimension` is `{2}` in `source`:
```
hadia206: Same

documentation/metadata.md
```diff
@@ -505,6 +506,277 @@ Below are several examples the JSON for such deducers:
...
+    "description": "The year for which to calculate the revenue of orders."
+  },
+  "arg_dimension": {
+    "type": "pydough",
+    "description": "The dimension by which to partition the revenue calculation."
+  }
+}
+```
+
+Example of the structure of the metadata for a template named `orders_revenue_by` (depends on another template, `order_revenue`):
```
hadia206: Same

pydough/metadata/graphs/graph_metadata.py
```diff
@@ -188,8 +224,103 @@ def add_function(self, name: str, function: "ExpressionFunctionOperator") -> Non
...
+                raise PyDoughMetadataException(
+                    f"Already added {name!r} to {self.error_name}"
+                )
+            raise PyDoughMetadataException(
+                f"Duplicate templates: {name} versus {self.templates_definitions[name]}"
+            )
+        if name in builtin_registered_operators():
+            raise PyDoughMetadataException(
+                f"The template '{name}' already exists as a PyDough operator"
+            )
```
hadia206, line 328: shouldn't it also check against UDF names and existing collections too?

documentation/metadata.md
```diff
@@ -505,6 +506,277 @@ Below are several examples the JSON for such deducers:
...
+
+<!-- TOC --><a name="template-attributes"></a>
+### Attributes
+
+An attribute describes a reusable, labeled set of values that one or more templates can use as an argument. Attributes are exclusively used with `call_template()` — they map a user-facing **label** (e.g. `"Year 1996"`) to the actual PyDough/Python **value** (e.g. `1996`) that gets substituted into the template.
+
+Every JSON object describing an attribute has the following fields:
+- `name` (required): unique identifier for the attribute. Referenced by templates that use it.
+- `usage` (required): which templates (and optionally which parameters of those templates) may use this attribute. An empty list `[]` means "usable by all templates." [See here](#attribute-usage-restriction) for the specification of this field.
+- `type` (required): the value's type. This is used to verify if the value matches the template's parameter. [See supported here](#template-supported-types) for more.
```
hadia206:
```suggestion
- `type` (required): the value's type. This is used to verify if the value matches the template's parameter. [See supported types here](#template-supported-types) for more.
```

pydough/metadata/templates/template_metadata.py
```diff
@@ -0,0 +1,396 @@
...
+        args_parts: list[str] = []
+
+        part: str
+        for key, spec in self.parameters.items():
+            part = f"{key}: {spec.type}" if spec.type != "pydough" else f"{key}"
+            args_parts.append(part)
+        template_args: str = ", ".join(args_parts)
+
+        # --- Replace {1}, {2}, ... with the corresponding argument name ---
+        # NOTE: Add a detail documentation about this replacement
```
hadia206: Reminder

pydough/errors/error_utils.py
```diff
@@ -706,6 +711,31 @@ def extract_integer(json_obj: dict, key_name: str, obj_name: str) -> int:
...
+        contains the string.
+        `obj_name`: the name of the object being extracted from, to be used
+        in error messages.
+
+    Returns:
+        The float value of the field.
+
+    Raises:
+        `PyDoughMetadataException` if the JSON object does not contain a key
+        with the name `key_name`, or if the value of the key is not an integer.
```
hadia206:
```suggestion
        with the name `key_name`, or if the value of the key is not a float.
```

tests/testing_utilities.py
```diff
@@ -1054,7 +1054,23 @@ def transform_and_exec_pydough(
         return pydough.from_string(pydough_impl, metadata=graph, environment=kwargs)
     else:
         # Otherwise, transform the function with the decorator and call it.
-        return init_pydough_context(graph)(pydough_impl)(**kwargs)
+
+        # Temporarily set the active session (or just metadata) so that
+        # UnqualifiedNode.__call__ can resolve templates via
+        # pydough.active_session.metadata, and so functions like
+        # pydough.to_table can access session info during execution.
```
hadia206: This comment is outdated name after the changes that were made.
`__call__` now resolves templates through the node's own `_parcel`, not the global session.
So update the comment to say the swap is needed for `pydough.call_template()` (and things like `pydough.to_table`)

tests/test_templates.py
hadia206: Thanks for adding these, looks like 4 of the 5 are covered now (float/dict/datetime/DataFrame attribute types, both usage shorthands, the attribute option value containing a quote, and the out of range placeholder case). 

One's still missing: a test calling a zero-parameter template like via `call_template()` on a graph with no attributes at all.

hadia206 (commented):
Almost there :)

hadia206 (approved):
Great job John! Thanks a lot!

PR #566, opened 2026-09-18
Fix bug with wrong alias generation for quoted name

### LINKED ISSUE #567: Wrong renaming refs when original column name requires quoting (""name""_idx instead of "name_idx")
**Summary:**

CALCULATE fails with a PyDoughSQLException when a calculated field references a column whose original name requires SQL quoting (e.g. "Percent (%) Eligible FRPM (K-12)") and that reference gets deduplicated/renamed with an index suffix. The generated SQL doubles the quotes and places the _idx suffix outside the closing quote instead of inside it, producing an invalid identifier like ""Percent (%) Eligible FRPM (K-12)""_1, which sqlglot cannot parse as part of the IIF(...) expression.

**Reproducible example:**

```python
result = california_schools.schools.CALCULATE( 
    frpm_level_name=IFF( frpm.percent_percentage_eligible_frpm_k_12 >= 0.75, 'High', 'Medium-Low' )
    ).PARTITION( 
        name="county_frpm_groups", 
        by=(county, frpm_level_name) 
    ).schools.CALCULATE( 
        percent_eligible_frpm_k_12=frpm.percent_percentage_eligible_frpm_k_12, 
        frpm_category=frpm_level_name 
    )
```

**Error:**

```
pydough.errors.error_types.PyDoughSQLException: SQL query optimization failed. Please check the query syntax:
SELECT
  "Percent (%) Eligible FRPM (K-12)" AS "percent_eligible_frpm_k_12",
  IIF(""Percent (%) Eligible FRPM (K-12)""_1 >= 0.75, 'High', 'Medium-Low') AS frpm_category
FROM (
  SELECT
    _s2."Percent (%) Eligible FRPM (K-12)" AS ""Percent (%) Eligible FRPM (K-12)""_1,
    _s3."Percent (%) Eligible FRPM (K-12)" AS "Percent (%) Eligible FRPM (K-12)"
  FROM (
    SELECT
      _s0."CDSCode" AS "CDSCode",
      _s1."Percent (%) Eligible FRPM (K-12)" AS "Percent (%) Eligible FRPM (K-12)"
    FROM (
      SELECT
        "CDSCode" AS "CDSCode"
      FROM main.schools
    ) AS _s0
    LEFT JOIN (
      SELECT
        "CDSCode" AS "CDSCode",
        "Percent (%) Eligible FRPM (K-12)" AS "Percent (%) Eligible FRPM (K-12)"
      FROM main.frpm
    ) AS _s1
      ON _s0."CDSCode" = _s1."CDSCode"
  ) AS _s2
  LEFT JOIN (
    SELECT
      "CDSCode" AS "CDSCode",
      "Percent (%) Eligible FRPM (K-12)" AS "Percent (%) Eligible FRPM (K-12)"
    FROM main.frpm
  ) AS _s3
    ON _s2."CDSCode" = _s3."CDSCode"
) AS _t0
ORDER BY
  "Percent (%) Eligible FRPM (K-12)" DESC
LIMIT 5
Error: Required keyword: 'true' missing for <class 'sqlglot.expressions.If'>. Line 1, Col: 88.
  SELECT "Percent (%) Eligible FRPM (K-12)" AS "percent_eligible_frpm_k_12", IIF(""Percent (%) Eligible FRPM (K-12)""_1 >= 0.75, 'High', 'Medium-Low') AS frpm_category FROM (SELECT _s2."Perc
```

**Expected behavior:**

The query compiles to valid SQL, with the renamed reference correctly quoted, e.g.:
```
IIF("Percent (%) Eligible FRPM (K-12)_1" >= 0.75, 'High', 'Medium-Low') AS frpm_category
```

john-sanchez31:
## Linked ticket

Closes #567 

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Docs / config

## What changed and why?

For the `widen_columns` function on `projection_pullup.py` file the new name generation removes all non alphanumeric characters using a regular expression. This avoids the generation of wrong names for quoted names like `""Percent (%) Eligible FRPM (K-12)""_1`.

## How I tested this?

 - New tests added for s3 called `california_schools_quoted_alias`


pydough/conversion/projection_pullup.py
```diff
@@ -83,7 +84,10 @@ def widen_columns(
                 idx: int = 0
                 while new_name in node.columns:
                     idx += 1
-                    new_name = f"{name}_{idx}"
+                    # Removes all characters except letters, numbers, and underscores.
+                    input_name = re.sub(r"[^a-zA-Z0-9_]", "", name)
```
hadia206: It doesn't rely on the loop (`new_name` or `idx`) so it can be moved up outside/before the loop.

hadia206: nit:
rename variable so we don't confuse it as a real input (updated_name, sanitized_name, or something else)

hadia206: Also, what happens if `input_name` ended up being empty, are we okay with `_1`, `_2`, ...etc? Does all dialects allow names that begins with underscore.

hadia206 (commented):
Shouldn't the fix instead insert the suffix before the closing quote rather than discarding the quotes and every other special character? That would round-trip correctly through `normalize_column_name`

Also, it looks like this suffix-appending pattern shows up in a few places such as `join_outputs` and `get_column_name` in `relational_converter.py` at least. Worth checking those too, or opening a follow-up ticket to check if we need the same fix in multiple spots.

john-sanchez31:
> Shouldn't the fix instead insert the suffix before the closing quote rather than discarding the quotes and every other special character? That would round-trip correctly through `normalize_column_name`
> 
> Also, it looks like this suffix-appending pattern shows up in a few places such as `join_outputs` and `get_column_name` in `relational_converter.py` at least. Worth checking those too, or opening a follow-up ticket to check if we need the same fix in multiple spots.

I created a followup issue (#569), since those possible bugs will require investigation/experimentation to actually have a test that triggers the same issue in other parts of PyDough.

pydough/conversion/projection_pullup.py
```diff
@@ -35,6 +35,20 @@
...
+    """
+    Build a new reference name by appending an index suffix.
+
+    If `name` is wrapped in double quotes (e.g. '"name"'), the index is
+    inserted before the closing quote, so the result stays quoted
+    (e.g. '"name_idx"'). Otherwise, the index is simply appended
+    (e.g. 'name_idx').
+    """
+    if len(name) >= 2 and name.startswith('"') and name.endswith('"'):
+        return f'"{name[1:-1]}_{idx}"'
```
hadia206, line 48: While we're at it, it should check for backtick-quotes (`) too

pydough/conversion/projection_pullup.py
```diff
@@ -39,13 +39,30 @@ def build_new_ref_name(name: str, idx: int) -> str:
...
+        `column` -> `column_1`
+
     """
-    if len(name) >= 2 and name.startswith('"') and name.endswith('"'):
-        return f'"{name[1:-1]}_{idx}"'
+    if len(name) >= 2:
+        quote = name[0]
+
+        if quote == name[-1] and quote in {'"', "`"}:
+            return f"{name[:-1]}_{idx}{name[-1]}"
```
hadia206: nit:
use `quote` instead of `name[-1]`

hadia206 (approved):
Thanks John.

pydough/conversion/projection_pullup.py
```diff
@@ -35,6 +35,37 @@
...
+        name       -> name_1
+        "column name" -> "column name_1"
+        "column ""name""\" -> "column ""name""_1"
+        `column` -> `column_1`
+
+    """
+    if len(name) >= 2:
+        quote = name[0]
+
+        if quote == name[-1] and quote in {'"', "`"}:
```
juankx-bodo: This is OK for the dialects we have right now. But we need to keep a watch on this because other dialetcs may use different characters for quoting. i.e. in SQLServer you traditionally use `[` and `]` for "quoting" a name.

Is there a property or something that say if `name` is already quoted?

juankx-bodo (approved):
LGTM

